Issue #63, opened 2019-11-01
Unify and customizable Conll-like readers


hunterhector:
The current ConllU and ontonotes reader are both in the same format, the conll-like readers. There are many variations to the CoNLL format, such as different fields for different columns. The current readers are hard-coded on the fields, and the file extensions. However, one single reader should be able to handle these, with some configurable file extensions and field names.

PR #66, opened 2019-11-02
Resolve #64 defects in OntonotesReader

### LINKED ISSUE #64: Defects in `OntonotesReader`
`OntonotesReader` has the following issues:

1. Breaks when encountering `V` spans with length greater than 1. `V` spans correspond to "verb" and are ignored, but the current logic for doing so is as follows:
https://github.com/asyml/forte/blob/beae4e923c9a6873b582588972e6ec9919079271/forte/data/readers/ontonotes_reader.py#L236-L241
   If a `V` span with length greater than 1 is encountered, only the opening bracket is ignored, and an exception would be raised when encountering the closing bracket.

2. Actual time complexity is O(n^2) due to appending each word to the string `text`:
https://github.com/asyml/forte/blob/beae4e923c9a6873b582588972e6ec9919079271/forte/data/readers/ontonotes_reader.py#L147
   Since Python strings are immutable, a new string is created for each word, yielding O(n^2) complexity.

huzecong:
Also add support for partial CoNLL format files -- files where several (unrelated) columns are missing. Users can specify the partial format via the `missing_fields` constructor argument.


hunterhector:
> Also add support for partial CoNLL format files -- files where several (unrelated) columns are missing. Users can specify the partial format via the `missing_fields` constructor argument.

The second objective may be the same as in https://github.com/asyml/forte/issues/63 . Would it be a better design if we can specify the column number for the targeted columns? For example, 2 for predicate, 3-4 for argument, etc. This way we might be able to expand the ontonotes reader to be a universal CoNLL reader.

huzecong:
Good point. I did that. Also fixed mypy and unit test errors related to the reader.

forte/data/readers/ontonotes_reader.py
```diff
@@ -25,7 +26,80 @@ class OntonotesReader(PackReader):
...
+            - ``"entity_label"``
+            - ``"coreference"``
+            - ``"*predicate_labels"``
+
+            Field types marked with ``*`` indicate a variable-column field: it
+            could span multiple fields. Only one such field is allowed in the
+            format specification.
+
+            If a column should be ignored, fill in `None` at the corresponding
+            position.
```
hunterhector: I think this function can handle different numbers of fields by providing different lengths of ``column_format``, this might not be immediately clear when reading the docstring.

PR #70, opened 2019-11-04
Automatic code generation

mgupta1410:
- Simplified code generation JSON format 
- Added code generation utils
- Added user script to generate ontology in the base folder


forte/data/ontology/configs/example_import_ontology_config.json
```diff
@@ -1,27 +1,31 @@
 {
-    "name": "ft.onto.example_import_ontology",
-    "definitions": {
-        "Token": {
+    "name": "example_import_ontology",
+    "definitions": [
+        {
             "namespace": "ft.onto.example_import_ontology.Token",
```
hunterhector: - This design can not achieve what we planned. Our decision is to add a prefix to all user-generated types (ft.onto) by default. But if we are asking the user to write ``ft.onto``every time, this will not be enforced.  The ``ft.onto`` should be inserted automatically by default. If the user wants to change it, he can use a top-level non-empty field ``prefix``, to change that (so the default behavior is "prefix": "ft.onto")

- This is not a namespace, this is simply a name. Please name it ``entry_name``. See https://github.com/hunterhector/forte/blob/master/forte/data/ontology/configs/stanfordnlp_ontology.json

forte/data/ontology/code_generation_util.py
```diff
@@ -0,0 +1,190 @@
+import os
+from collections import OrderedDict
+from typing import List, Optional, Any
+
+
+class Config:
+    indent: int = 4
+    line_break: str = "\n"
```
hunterhector: Try ``os.linesep``

forte/data/ontology/code_generation_util.py
```diff
@@ -0,0 +1,190 @@
...
+                 name: str,
+                 type_str: str,
+                 items: List[str],
+                 description: Optional[str] = None,
+                 default: Any = None):
+        super().__init__(name, type_str, description, default)
+        self.items = items
+
+    def type_name(self):
+        return f"typing.Optional[{self.type_str}[{', '.join(self.items)}]]"
```
hunterhector: Can a List type contain a list of comma-separated types?

requirements.txt
```diff
@@ -5,4 +5,8 @@ jsonpickle
 torch
 pyyaml
 smart-open
-allennlp
\ No newline at end of file
+allennlp
```
hunterhector: Is allennlp needed here?

requirements.txt
```diff
@@ -5,4 +5,8 @@ jsonpickle
...
+typed_astunparse
+funcsigs
+mypy_extensions
+typed_ast
```
hunterhector: Could you specify the versions of these newly added requirements.

forte/models/ner/model_factory.py
```diff
@@ -174,7 +174,7 @@ def prepare_rnn_seq(rnn_input, lengths, hx=None, masks=None, batch_first=False):
 
     Args:
         rnn_input: [seq_len, batch, input_size]:
-            tensor containing the features of the input sequence.
+            tensor containing the ud_features of the input sequence.
```
hunterhector: I think this is changed by accident.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -0,0 +1,491 @@
...
+                 ontology_base_module: ModuleType = top):
+        """
+        Args:
+            ontology_base_module: The entries of the `base_ontology_module`
+            serve as ancestors of the user-defined entries.
+            json_dir_paths: Additional user provided paths to search the
+            imported json configs from. By default paths provided in the json
+            configs and the current working directory would be searched.
+            indent: Number of indent required in the indentation.
+        """
```
hunterhector: The docstring does not match the function signature, and probably the indent of the docstring is not correct.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -0,0 +1,491 @@
...
+        # (default is `top.py`), to their `__init__` arguments.
+        self.top_init_args: Dict[str, str] = {}
+        self.initialize_top_entries()
+
+        # Mapping from user-defined entries to their ancestor entry present in
+        # `self.top_init_args`.
+        self.user_to_base_entry: Dict[str, str] = {}
+
+        # Mapping from reference string to corresponding namespace
+        self.ref_to_namespace: Dict[str, str] = {}
```
hunterhector: These shouldn't be called ``namespace``

forte/data/ontology/ontology_code_generator.py
```diff
@@ -0,0 +1,491 @@
...
+                        f"the attribute {property_name} already present in "
+                        f"the ontology, will be overridden")
+                self.allowed_types_tree[class_name].add(property_name)
+
+        return new_modules_to_import
+
+    def cleanup_generated_ontology(self, path, is_forced=False):
+        """
+        Deletes the generated ontology files.
+        """
```
hunterhector: Args missed in docstring.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -0,0 +1,491 @@
...
+                    os.rmdir(path)
+            log.info("Deleted %s.", path)
+        return is_empty
+
+    def parse_entry(self, ref_name: str, namespace: str, schema: Dict) \
+            -> Tuple[DefinitionItem, List[str]]:
+        """
+        Args:
+            ref_name: Reference name of the entry.
+            namespace: Full namespace of the entry.
```
hunterhector: Again not namespace.

forte/data/ontology/configs/example_ontology_config.json
```diff
@@ -1,25 +1,26 @@
...
                 },
                 {
-                    "attribute_name": "string_features",
-                    "attribute_description": "Miscellaneous string features",
-                    "attribute_type": "List",
-                    "element_type": "str"
+                    "name": "word_forms",
+                    "description": "To demonstrate that an attribute can have the entry it is contained in as it's type or item_type.",
+                    "type": "List",
+                    "item_type": "example_ontology.Word"
```
hunterhector: The test ontology is not quite sufficient. Could you try adding a couple more types? We could test the following:

1. Multiple entry definitions, this only contains one.
1. Entry definitions inherited from types form the same file (e.g Try extend word to something else?)
1. Try to use each type at least once  (i.e. str, int, bool, float, List)

You may be able to think of other points here. The point is, we would like to have a fairly comprehensive test, without assuming they would all work.

forte/data/ontology/configs/stanfordnlp_ontology.json
```diff
@@ -1,24 +1,20 @@
...
         },
         {
-          "attribute_name": "pos_tag",
-          "attribute_type": "str"
+          "attribute_name": "is_verb",
+          "attribute_type": "bool"
         },
         {
-          "attribute_name": "upos",
+          "attribute_name": "num_chars",
```
hunterhector: num_chars?

forte/data/ontology/configs/stanfordnlp_ontology.json
```diff
@@ -1,24 +1,20 @@
...
+      "parent_entry": "Annotation",
       "attributes": [
         {
           "attribute_name": "lemma",
           "attribute_type": "str"
         },
         {
-          "attribute_name": "pos_tag",
-          "attribute_type": "str"
+          "attribute_name": "is_verb",
```
hunterhector: Why are the example attributes copied here?

forte/data/ontology/configs/stanfordnlp_ontology.json
```diff
@@ -28,27 +24,27 @@
...
-          "element_type": "forte.data.ontology.stanfordnlp_ontology.Token"
+          "element_type": "stanfordnlp_ontology.Token"
         }
       ]
     },
     {
-      "entry_name": "forte.data.ontology.stanfordnlp_ontology.Document",
-      "parent_entry": "forte.data.ontology.top.Annotation"
+      "entry_name": "stanfordnlp_ontology.Document",
+      "parent_entry": "Annotation"
```
hunterhector: We can't accurately find the ``parent_entry`` without a full name, right?

mgupta1410: If it is a unique name, we can. For example, in this case Annotation is a class in `top`, and doesn't have a duplicate, otherwise, you need to provide the full name.  I will add a test case for (1) full name (2) duplicate entries.

forte/data/ontology/configs/stanfordnlp_ontology.json
mgupta1410: restored it - sorry about this, it was an error.

forte/data/ontology/configs/stanfordnlp_ontology.json
mgupta1410: restored it.

forte/data/ontology/configs/stanfordnlp_ontology.json
hunterhector: It seems that we have to make special cases to deal with this. And users can face errors when they are not careful about this approach. Let's simply use the full qualified name.

forte/data/ontology/configs/example_ontology_config.json
```diff
@@ -7,7 +7,7 @@
     "definitions": [
         {
             "entry_name": "example_ontology.Word",
-            "parent_type": "example_import_ontology.Token",
+            "parent_entry": "example_import_ontology.Token",
```
hunterhector: I have a question here. The parent entry reference did not include the prefix of the example_import_ontology (i.e. ``ft.onto``). How do we know that the full name of this ``parent_entry`` is ``ft.onto.example_import_ontology.Token``, what if the user creates another type with a different prefix, such as ``edu.cmu.example_import_ontology.Token``? Can the current semantics distinguish these?

mgupta1410: Changed as discussed. All the entries and types have full names. If additional names are to be defined, it is done through `additional_prefixes` field at the top level of ontology. See `forte/data/ontology/configs/example_multi_module_ontology_config.json` for an example of multi-module ontology.

forte/data/ontology/configs/stanfordnlp_ontology.json
mgupta1410: Changed as discussed. The full name is being used everywhere.

PR #77, opened 2019-11-13
Adding WikiPassageQA reader

mgupta1410:


forte/data/readers/wiki_passage_qa_reader.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            query = Query(data_pack)
+            query.query = doc_content[0]
+            if rel_docs is not None:
+                query.doc_ids = {"relevant_docs": rel_docs}
+            annotations = [query]
+
+        # add annotations to data_pack
+        for annotation in annotations:
+            data_pack.add_or_get_entry(annotation)
```
hunterhector: Let's don't use ``add_or_get_entry`` in the future, we are planning to deprecate them.

forte/data/readers/wiki_passage_qa_reader.py
```diff
@@ -0,0 +1,159 @@
...
+        if not is_query:
+            # add passages
+            for passage in doc_content:
+                annotations.append(Passage(data_pack, 0, len(passage)))
+
+            doc_text = os.linesep.join(doc_content)
+            annotations.append(Document(data_pack, 0, len(doc_text)))
+            data_pack.set_text(doc_text)
+        else:
+            query = Query(data_pack)
```
hunterhector: So here the ``Query`` type is used for the questions. I would not do that for a couple of reasons:

1. ``Query`` are objects we designed for IR, which should contain information such as embeddings and query statements. 
2. ``Query`` are not annotations, so they cannot contain text. I think the text of these questions are not stored.

hunterhector:
This PR contains way too much commits. And most of them are not related to the reader. Let's make the code gen PR through the following path:  mansi -> hector -> asyml   , jumping back and forth is confusing.

Issue #2, opened 2019-08-14
Associate embeddings to entries


hunterhector:
We need a consistent method to hook embeddings to entries. This can be done by one of the following, which can be applied in different cases:

1. Store the embeddings directly in the entries, this can be useful when the embedding is difficult to compute and cannot be looked up easily (say passage vector)
1. Store the vocabulary index to each entry and require a resource (embedding dictionary) to be available at the time of interpretation.
1. Store the functions of computing the embedding together with the entry, this requires we store both the function and the resource at the same time, which seem to be tricky to implement.

PR #78, opened 2019-11-13
Add Indexer into Forte

AvinashBukkittu:
This PR

- Adds Indexer module into Forte (currently supports EmbeddingBasedIndexer)
  - Users can create an index on a CPU or a GPU by setting "device" in HParams.

- Adds test cases for indexers.

- Introduces two types 
  - Embedding: An abstraction which is used to represent embedding for any entry type. 
    For e.g., this adds support for functionality like sentence.embed() (This PR does not add `embed()` method yet.)
  - Query: An entry type for IR related tasks. This type internally uses `Embedding` to represent queries.

- Tweaks the chatbot example to include the new entries added to the Forte type system.


forte/processors/query_creator.py
```diff
@@ -0,0 +1,84 @@
...
+        if "bot_utterance" in input_pack.pack_names:
+            bot_pack = input_pack.get_pack("bot_utterance")
+            context.append(bot_pack.text)
+
+        for text in context:
+            t = self.tokenizer.encode_text(text)
+            input_ids.append(t[0])
+            segment_ids.append(t[1])
+
+        input_ids = torch.LongTensor(input_ids).to(self.device)
```
hunterhector: These are the steps that's worth abstracting away into an ``Entry.get_embed`` method in the future.

hunterhector:
We now have copyright headers on other files. Could you also add those in the new files?

AvinashBukkittu:
> We now have copyright headers on other files. Could you also add those in the new files?

Fixed in 462e2b1

forte/processors/machine_translation_processor.py
```diff
@@ -0,0 +1,86 @@
...
+    def initialize(self, resources: Optional[Resources],
+                   configs: Optional[HParams]):
+        if configs:
+            self.src_language = configs.src_language
+            self.target_language = configs.target_language
+            self.in_pack_name = configs.in_pack_name
+            self.out_pack_name = configs.out_pack_name
+
+    # pylint: disable=no-self-use
+    def _define_input_info(self) -> DataRequest:
```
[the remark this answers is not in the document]
AvinashBukkittu: Fixed in 1e0aeba

forte/processors/query_creator.py
AvinashBukkittu: I guess this relates to https://github.com/asyml/forte/issues/2. This will be fixed as part of that issue.

hunterhector: Yes, that's the plan

PR #82, opened 2019-11-15
Refine Chatbot example

AvinashBukkittu:
This PR

- Improves chatbot example with contextual information. Uses the dataset here https://github.com/squareRoot3/Target-Guided-Conversation to finetune the model
- Uses Siamese BERT network structure to finetune BERT for sentence embedding
- Updates `SearchProcessor` processor to use the finetuned BERT model
- Splits `QueryCreator` into a `QueryProcessor` base class and a concrete `BertBasedQueryCreator` class for better design
- Updates the index with the new BERT model
- Adds scripts to download models from ASYML google account to run the example in inference only mode
- Renames `indexer` example to `chatbot`
- Updates the README file.


forte/processors/machine_translation_processor.py
```diff
@@ -49,6 +49,8 @@ def __init__(self) -> None:
     # pylint: disable=unused-argument
     def initialize(self, resources: Optional[Resources],
```
hunterhector: If this processor can only call MS translation, maybe it should be named more specifically. Or you can create an abstract MT processor that handle some basic operations.

forte/processors/query_creator.py
```diff
@@ -29,32 +31,43 @@
 
 
 class QueryCreator(MultiPackProcessor):
-    r"""This processor is used to search for relevant documents for a query
-    """
+    r"""This processor searches relevant documents for a query"""
```
hunterhector: Maybe we can also have an abstract QueryProcessor that exposes a ``build_query`` function, and a specific query processor can build the actual query.

examples/chatbot/data_utils.py
```diff
@@ -0,0 +1,127 @@
...
+            qa_with_history = [' '.join(qa_with_history[:-1]),
+                               qa_with_history[-1]]
+        else:
+            qa_with_history = current_qa
+
+        new_text_data.append(qa_with_history)
+    return new_text_data
+
+
+def create_dataset_with_history(conversations, num_line=2):
```
hunterhector: Some of these utilities seem to be useful for other people to build chatbots, shall we move some to the core?

examples/chatbot/prepare_chatbot_data.py
```diff
@@ -0,0 +1,156 @@
...
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+config_data = importlib.import_module(args.config_data)
+
+
+def main():
+    train_data = []
+    eval_data = []
+    test_data = []
+
```
hunterhector: Many codes for preparing the data could be boilerplates, let's consider how to standardize them and allow users to implement this with a few lines of code.

examples/chatbot/data_utils.py
AvinashBukkittu: At this point I would like to keep it in the examples. I am not sure if we want to make `data preparation` as part of Forte's `core`.

examples/chatbot/prepare_chatbot_data.py
AvinashBukkittu: This file is essentially doing the following

- Split the dataset into train, eval and test
- Augment the data with contextual information
- Generate Negative examples
- Saving it as RecordData

The functions for (1)-(3) are found in `data_utils.py`. We may impose some standard on the structure. But I am not sure if we want to move it as part of Forte core.

forte/processors/machine_translation_processor.py
AvinashBukkittu: I have renamed it to `MicrosoftBingTranslator` in ebab198

forte/processors/query_creator.py
AvinashBukkittu: I have created `QueryProcessor` base class in ebab198

forte/processors/bert_based_query_creator.py
```diff
@@ -65,6 +71,21 @@ def get_embeddings(self, inputs, sequence_length, segment_ids):
...
+        input_mask = torch.LongTensor(input_mask).unsqueeze(0).to(self.device)
+        sequence_length = (1 - (input_mask == 0)).sum(dim=1)
+        query_vector = self.get_embeddings(inputs=input_ids,
+                                           sequence_length=sequence_length,
+                                           segment_ids=segment_ids)
+        query_vector = torch.mean(query_vector, dim=0, keepdim=True)
+        query_vector = query_vector.cpu().numpy()
+        return query_vector
+
     def _process(self, input_pack: MultiPack):
```
hunterhector: Can this function in the abstract class?

AvinashBukkittu: We could. But base class `QueryProcessor` is supposed to work with both `PackProcessor` and `MultiPackProcessor` and `_process` in `BertBasedQueryCreator` uses multi-pack related methods while accessing the data pack. Hence I thought it is better to define this method here.

hunterhector: Huh, so we have hit the wall of multi-inherence so quick. That's something I am worrying about during the design phase.

AvinashBukkittu: Added a `_process_query` method in `QueryProcessor`. Subclasses of `QueryProcessor` should now implement this method.

forte/processors/base/query_processor.py
```diff
@@ -24,13 +26,44 @@
...
+             - If `input_pack` is a multipack, `query_pack` is one of its data
+             pack for which a `query` is generated.
+             - If `input_pack` is a datapack, `query_pack` is the `input_pack`.
+        """
+        raise NotImplementedError
+
+    def _process(self, input_pack: PackType):
+        query_pack, query_vector = self._process_query(input_pack)
+        query = Query(pack=query_pack, value=query_vector)
+        query_pack.add_or_get_entry(query)
```
hunterhector: let's simply use `add_entry` here, we would deprecate `add_or_get_entry` soon.

PR #97, opened 2019-12-02
Adding a simple corpus reader for MS Marco passages

mgupta1410:


forte/data/readers/corpus_reader.py
```diff
@@ -0,0 +1,86 @@
...
+from forte.data import DataPack
+from forte.data.readers.base_reader import PackReader
+from forte.common.resources import Resources
+
+__all__ = [
+    "CorpusReader"
+]
+
+
+class CorpusReader(PackReader):
```
hunterhector: Apparently, this class name, and the filename are too generic. Isn't this for MS-Macro only?

PR #103, opened 2019-12-27
Adding indexer+reranker inference pipeline; passage reranking bert model

mgupta1410:

PR #105, opened 2019-12-30
Elastic indexer

AvinashBukkittu:
This PR

- Adds evaluator support for the pipeline. This means, we can now add an evaluator to the pipeline and call `evaluate()` on the pipeline to evaluate on a dataset.
- Adds ElasticIndexer along with `ElasticSearchIndexProcessor` processor to index the documents.
- Adds ElasticSearchProcessor for searching documents in an elastic indexer
- Creates Passage Reranker example for MS Marco Dataset. Provides a baseline model for ranking using just Elastic Indexer
  - Adds an EvalReader to read MS Marco eval dataset
  - Adds MS Marco Eval script in the example


forte/processors/elastic_search_query_creator.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @staticmethod
+    def default_hparams() -> Dict[str, Any]:
+        return {
+            "size": 1000,
+            "field": "content",
+            "query_pack_name": "query"
+        }
+
+    def _process_query(self, input_pack: MultiPack):
```
hunterhector: Return type would help

forte/processors/elastic_search_processor.py
```diff
@@ -0,0 +1,72 @@
...
+    def default_hparams() -> Dict[str, Any]:
+        return {
+            "query_pack_name": "query",
+            "index_config": ElasticSearchIndexer.default_hparams(),
+            "field": "content"
+        }
+
+    def _process(self, input_pack: MultiPack):
+        query_pack = input_pack.get_pack(self.config.query_pack_name)
+        first_query = list(query_pack.get_entries(Query))[0]
```
hunterhector: It would help future developers if we comment why we use the first query here.

forte/processors/elastic_search_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        packs = {}
+        for idx, hit in enumerate(hits):
+            document = hit["_source"]
+            first_query.update_passage({document["doc_id"]: hit["_score"]})
+            pack = DataPack(doc_id=document["doc_id"])
+            content = document[self.config.field]
+            # TODO: add the BM score and the labels for retrieved documents
+            # which will be used by the reranker
+            # Convert `Document` to `Passage`
+            document = Document(pack=pack, begin=0, end=len(content))
```
hunterhector: This document entry doesn't not have meta data apart from the character range, what is its utility?

forte/processors/base/index_processor.py
```diff
@@ -0,0 +1,63 @@
...
+from forte.common import Resources
+from forte.data import DataPack
+from forte.processors.base.base_processor import BaseProcessor
+
+__all__ = [
+    "IndexProcessor"
+]
+
+
+class IndexProcessor(BaseProcessor[DataPack], ABC):
```
hunterhector: Let's have a class docstring here.

examples/passage_reranker/ms_marco_evaluator.py
```diff
@@ -0,0 +1,39 @@
+from texar.torch import HParams
```
hunterhector: Well, header is missing here.

examples/passage_reranker/reader.py
```diff
@@ -0,0 +1,38 @@
+from typing import Iterator
```
hunterhector: Header

forte/indexers/indexers.py
```diff
@@ -37,20 +40,147 @@ class BaseIndexer(ABC):
...
+                If None, value will be picked from hparams.
+
+        Returns:
+            A dict containing the documents matching the query along with
+            meta data of the search.
+        """
+        index_name = index_name if index_name else self.hparams.index_name
+        return self.elasticsearch.search(index=index_name, body=query)
+
+    def search_all(self, query: Dict[str, Any]):
```
hunterhector: What's the plan for this method? If we are not planning to implement this soon, maybe we should add not_implemented error here?

forte/processors/elastic_search_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        return {
+            "query_pack_name": "query",
+            "index_config": ElasticSearchIndexer.default_hparams(),
+            "field": "content"
+        }
+
+    def _process(self, input_pack: MultiPack):
+        query_pack = input_pack.get_pack(self.config.query_pack_name)
+        first_query = list(query_pack.get_entries(Query))[0]
+        results = self.index.search(first_query.value)
```
hunterhector: Now the design for `index.search` is that it takes the value of `first_query`. How about taking the whole `query` entry? That way if we have a more complex query we can represent them with the `query` object, instead of forcing everything inside `value`.

examples/passage_reranker/reader.py
AvinashBukkittu: Added in 2fc3e5f

examples/passage_reranker/ms_marco_evaluator.py
AvinashBukkittu: Added in 2fc3e5f

forte/processors/base/index_processor.py
AvinashBukkittu: Added in 4776d40

forte/processors/elastic_search_query_creator.py
AvinashBukkittu: Added in 4776d40

forte/processors/elastic_search_processor.py
AvinashBukkittu: Added in 4776d40

forte/processors/elastic_search_processor.py
AvinashBukkittu: Each retrieved result is added as `Document` type in the new pack. It is similar to how `SearchProcessor` was adding `Document` types. If the subsequent processors need to fetch the results, they can use `get_entries_by_type(...)` on this pack.

forte/indexers/indexers.py
AvinashBukkittu: Removed it in b7632b8

forte/processors/elastic_search_processor.py
AvinashBukkittu: Hmm, the `value` attribute of `Query` was designed to be generic enough to hold any complex query. `Query` object may hold other attributes like `rank_list` etc which maybe irrelevant to the indexer.

AvinashBukkittu:
Importing relevant comments from https://github.com/asyml/forte/pull/103

1) https://github.com/asyml/forte/pull/103#discussion_r361890259
> Let's have more typing here.

Added typing in MS Marco Evaluator

2) https://github.com/asyml/forte/pull/103#discussion_r361890348
> It would be better to store only some necessary information from the pack, here we only need the doc_id?

Simplified the logic of MS Marco Evaluator in b7632b8

3) https://github.com/asyml/forte/pull/103#discussion_r361890717
> passage is too specific as a name. how about rank_list?

Changed to `results` from `passages`

4) https://github.com/asyml/forte/pull/103#discussion_r361891432
> Add some docstring here to teach users to extend this method in order to create more complex queries.

Done in b7632b8

5) https://github.com/asyml/forte/pull/103#discussion_r361891867
> Do we need to benchmark the speed of the indexer? Hopefully, our wrapper won't decrease the speed a lot.

Added a benchmark testcase in b7632b8

6) https://github.com/asyml/forte/pull/103#discussion_r361892701
> The design of the _process_query seem to be a little difficult, especially on returning input_pack.

I agree. Essentially, the following three lines

```
query = Query(pack=query_pack)
query.set_value(value=query_value)
query_pack.add_entry(query)
``` 
are common in `QueryProcessor`. If we want to abstract away these details in `_process` method, I couldn't think of a better way other than to return the `query_pack` and the `query_value` from `_process_query`

7) https://github.com/asyml/forte/pull/103#discussion_r361892941
> at least in bulk mode, we should add a couple more documents.

Increased the limit to 10,000

PR #108, opened 2020-01-03
fix resources save none key bug

tcl326:
When `keys` argument is `None`, the `keys` is assigned with `self.resources.keys()`. The `keys` variable is supposed to be a `list`, however, in Python 3, `dict.keys()` returns `dict_keys` data type instead of `list`. In this pull request, we explicitly convert the `iterator` to `list` to fix this bug.


forte/common/test/test_resources.py
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Unit tests for Resources.
+"""
+import unittest
+
+from forte.common.resources import Resources
```
AvinashBukkittu: Can we use `from forte.common import Resources` instead? `common/__init__.py` imports everything inside `resources` into `common`

PR #114, opened 2020-01-08
Deserialize reader

AvinashBukkittu:
This PR

- Adds a Deserialize Reader and a test case


AvinashBukkittu:
> I am going to approve this now. But we'd better add some documentation for the reader on the expected input

I will make a note of this.

PR #117, opened 2020-01-11
Html reader with minor bug fixes and improvements

swapnull7:
Adds HTMLReader with a proposed template which supports reading from :

1. Directory
2. A specific file path
3. List of data
4. A single instance of data

Minor improvements in doc-strings and a small bug-fix to use self.set_text instead of pack.set_text


forte/data/readers/base_reader.py
```diff
@@ -50,15 +50,15 @@ class BaseReader(PipelineComponent[PackType], ABC):
 
     def __init__(self,
                  from_cache: bool = False,
-                 cache_directory: Optional[Path] = None,
```
swapnull7: Could go either ways on `Path`/ `str`

forte/data/readers/base_reader.py
```diff
@@ -50,15 +50,15 @@ class BaseReader(PipelineComponent[PackType], ABC):
...
         Args:
             from_cache (bool, optional): Decide whether to read from cache
-                if cache file exists. By default (``True``), the reader will
-                try to read an datapack from the first line of the caching file.
-                If ``False``, the reader will only read from the original file
-                and use the cache file path only for output.
+                if cache file exists. By default (``False``), the reader will
+                only read from the original file and use the cache file path
+                only for output. If ``True``, the reader will try to read a
+                datapack from the first line of the caching file.
```
hunterhector: Actually I don't understand "the first line of the caching file", I think that is not correct.

forte/data/readers/base_reader.py
```diff
@@ -50,15 +50,15 @@ class BaseReader(PipelineComponent[PackType], ABC):
...
         """
         Args:
             from_cache (bool, optional): Decide whether to read from cache
-                if cache file exists. By default (``True``), the reader will
-                try to read an datapack from the first line of the caching file.
-                If ``False``, the reader will only read from the original file
-                and use the cache file path only for output.
+                if cache file exists. By default (``False``), the reader will
+                only read from the original file and use the cache file path
+                only for output. If ``True``, the reader will try to read a
```
hunterhector: Thanks for fixing the docstring, but I realize the original docstring is not very accurate.

"use the cache file path only for output"?why is the cache file related to output? This parameter should  only affect whether this reader reads from the cache, but it shouldn't affect its behavior on its own caching.

forte/data/readers/base_reader.py
```diff
@@ -243,7 +242,6 @@ def cache_data(self,
         :func:`serialize_instance`).
 
         Args:
-            cache_directory: The directory used to cache the data.
             collection: The collection to be read as a DataPack, this collection
```
hunterhector: The collection is a piece of data from the ``_collect`` function,  to be read to produce DataPack(s). During caching, a cache key is computed based on the data in this collection.

forte/data/readers/html_reader.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class HTMLReader(StringReader):
+    """
+    :class:`HTMLReader` is designed to read in list of html strings.
+
+    It takes in list of html strings, cleans the HTML tags and stores the
+    cleaned text in pack.
+    """
+    # pylint: disable=no-self-use,unused-argument
```
hunterhector: no-self-use is disabled in the pylint setup so we don't need them here.

forte/data/readers/html_reader.py
```diff
@@ -0,0 +1,234 @@
...
+        Returns: List[Tuple[Span, str]]: the replacement operations
+
+        """
+        parser = ForteHTMLParser()
+        parser.feed(text)
+
+        return parser.spans
+
+    def _cache_key_function(self, collection):
+        return str(hash(collection)) + '.txt'
```
hunterhector: Hashing on the data content could be inefficient. We hope the _cache_key_function to be as simple as possible. This is also related to the fact you are extending a StringReader. If it takes a directory of html documents, then the cache_key can simply be those paths or file names, like here: `https://github.com/asyml/forte/blob/master/forte/data/readers/plaintext_reader.py`

Further, I think the `_cache_key_function` in StringReader is wrong. Either we can change that to "hash", or we shouldn't use caching there.

forte/data/readers/tests/html_reader_test.py
```diff
@@ -0,0 +1,175 @@
...
+            count_orig = count_orig + 1
+
+        num_files = len(os.listdir(self._cache_directory))
+
+        self.assertEqual(num_files, count_orig)
+
+        # Test Caching
+        count_cached = 0
+        for _ in self.pl2.process_dataset(value):
+            count_cached = count_cached + 1
```
hunterhector: So the test only counts the number of files but not the content? I think we should compare the content to ensure caching is correct.

forte/data/readers/base_reader.py
swapnull7: it's what the current documentation says..I've just rearranged it. will fix it.

forte/data/readers/base_reader.py
swapnull7: I think it meant that the contents of the reader are written to the cache file if the `cache_directory` is not None basically, it is not affecting the caching behaviour but letting the user know that the reader will not read from `cache_directory` but only write to it. Otherwise,  the cache file path(`cache_directory` + whatever `_cache_key_function` returns) is not used. I think the words used are cause of miscommunication. This is the expected behaviour right?

forte/data/readers/base_reader.py
swapnull7: Will update

forte/data/readers/base_reader.py
hunterhector: Yeah, just wording issues.

forte/data/readers/html_reader.py
swapnull7: okay.. I'll remove it from `PlainTextReader` and `BaseReader` as well? and possibly other readers that still have this.

hunterhector: Pengzhi have removed a few. Since this doesn't matter too much so we will just remove them when seeing them.

forte/data/readers/html_reader.py
swapnull7: Right.. I'll update the HTMLReader to read from files and yeah.. i agree..i'll make the key_func as hash..some caching is better than no caching :)

forte/data/readers/tests/html_reader_test.py
swapnull7: Good Point! I'll add that

forte/data/readers/tests/html_reader_test.py
```diff
@@ -0,0 +1,209 @@
...
+                                 append_to_cache=True)
+
+        self.pl1 = Pipeline()
+        self.pl1.set_reader(self.reader)
+
+        self.pl2 = Pipeline()
+        self.pl2.set_reader(HTMLReader(from_cache=True,
+                                       cache_directory=self._cache_directory))
+
+        self.data_dir = os.path.join(
```
AvinashBukkittu: Initialize the pipeline after creation. `self.pl1.initialize()` and `self.pl2.initialize()`

forte/data/readers/html_reader.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    def __init__(self):
+        super().__init__()
+        self.spans = []
+
+    def collect_span(self, begin, end):
+        self.spans.append((Span(begin, end), ''))
+
+    def goahead(self, end):
+        rawdata = self.rawdata
```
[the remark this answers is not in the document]
swapnull7: This is actually an overloaded method from the Original HTMLParser, the only thing it doesn't do is save the span information, which we overload this function and save the span info. Done by `collect_spans`. Would it help if I add this as a comment?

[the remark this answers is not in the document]
swapnull7: Added

PR #118, opened 2020-01-13
Tests

AvinashBukkittu:
This PR

- Improves the logic of pipeline processing. The new logic is explained in the comments sections of `_process_packs` method.
- Adds more state to Process Manager to maintain status of the job during pipeline processing
- Adds test cases to check the different combinations of Pipeline processing
- Modifies existing test cases which failed to call `initialize()` on the pipeline after its creation.


forte/process_manager.py
```diff
@@ -32,19 +34,116 @@ class ProcessManager:
     class __ProcessManager:
         def __init__(self):
             self.current_component: str = '__default__'
+            self.pipeline_length: int
```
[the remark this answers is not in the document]
AvinashBukkittu: Improved the docs here.

forte/processors/tests/nltk_processors_test.py
```diff
@@ -19,6 +19,7 @@ def setUp(self):
         self.nltk = Pipeline()
         self.nltk.set_reader(StringReader())
         self.nltk.add_processor(NLTKSentenceSegmenter())
+        self.nltk.initialize()
```
gpengzhi: Duplicated line. Should delete it.

AvinashBukkittu: Okay. Thanks for letting me know. I will fix it in another PR.

PR #119, opened 2020-01-14
Spacy processors

swapnull7:
Adds Spacy Processors designed similar to stanford NLP Processors. Come with some added tests.


docs/requirements.txt
```diff
@@ -11,6 +11,7 @@ jsonpickle
 sortedcontainers
 texar
 texar-pytorch
+spacy
```
hunterhector: We usually don't put 3rd packages to the core requirements. Users only need to install them when needed

forte/processors/__init__.py
```diff
@@ -18,3 +18,4 @@
 from forte.processors.text_generation_processor import *
 from forte.processors.vocabulary_processor import *
 from forte.processors.bert_based_query_creator import *
+from forte.processors.spacy_processors import *
```
hunterhector: Similarly, we currently did not find a way to put 3rd party processors in __init__, since they might not get installed.

setup.py
```diff
@@ -27,7 +27,8 @@
         'funcsigs',
         'mypy_extensions',
         'typed_ast',
-        'jsonschema'
+        'jsonschema',
+        'spacy'
```
hunterhector: Add a keyword in extra requires and place spacy there.

forte/processors/spacy_processors.py
```diff
@@ -0,0 +1,105 @@
...
+]
+
+
+class SpacyProcessor(PackProcessor):
+    """
+    A wrapper for spaCy processors
+    """
+    def __init__(self):
+        super().__init__()
+        self.processors = ""
```
hunterhector: Our convention is to add type annotation when possible.

forte/processors/spacy_processors.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # sentence parsing
+        sentences = self.nlp(doc).sents
+
+        for sentence in sentences:
+            begin_pos = doc.find(sentence[0].text, end_pos)
+            end_pos = doc.find(sentence[-1].text, begin_pos) + len(
+                sentence[-1].text)
+            sentence_entry = Sentence(input_pack, begin_pos, end_pos)
+            input_pack.add_or_get_entry(sentence_entry)
```
hunterhector: Use `add_entry`, trying to retire `add_or_get_entry`

forte/processors/spacy_processors.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _process(self, input_pack: DataPack):
+        doc = input_pack.text
+        end_pos = 0
+
+        # sentence parsing
+        sentences = self.nlp(doc).sents
+
+        for sentence in sentences:
+            begin_pos = doc.find(sentence[0].text, end_pos)
```
hunterhector: This should not be our common way to align the offsets, at least i can see from here (https://spacy.io/api/token), that  the `sent` attribute can give you the span.

forte/processors/spacy_processors.py
```diff
@@ -0,0 +1,105 @@
...
+
+            tokens: List[Token] = []
+
+            if "tokenize" in self.processors:
+                offset = sentence_entry.span.begin
+                end_pos_word = 0
+
+                # Iterating through spaCy token objects
+                for word in sentence:
+                    begin_pos_word = sentence_entry.text.find(word.text,
```
hunterhector: It might be acceptable for sentences, but this method will probably fail if we have the same word twice.
From the  [doc](https://spacy.io/api/token), the `idx` attribute can give you the span.

forte/processors/tests/spacy_processors_test.py
```diff
@@ -0,0 +1,96 @@
...
+            self.assertEqual(tokens, document.split())
+
+            if "pos" in value:
+                self.assertIsNotNone(pos)
+            else:
+                none_pos = [None] * len(pos)
+                self.assertListEqual(pos, none_pos)
+
+            if "lemma" in value:
+                self.assertIsNotNone(lemma)
```
hunterhector: It would be great to check for the outcome in addition to checking None or not.

docs/requirements.txt
swapnull7: docs fail if I don't do this. Any idea why that happens?

setup.py
swapnull7: 👍

forte/processors/spacy_processors.py
swapnull7: Oh.. didn't know this.. I'll use that instead.

forte/processors/spacy_processors.py
swapnull7: 👍

forte/processors/tests/spacy_processors_test.py
swapnull7: I agree... I'll add that.

swapnull7:
> Besides the comments, there are also sentence segmenter, dependency parser, and NER available, will they be added in a separate PR?

If `self.processors` is empty, it'll act as a sentence segmenter. and yeah, planning on making dependency_parser and NER as a different PR.

forte/processors/spacy_processors.py
swapnull7: 👍

PR #122, opened 2020-01-14
Add NLTKChunker

gpengzhi:


forte/processors/nltk_processors.py
```diff
@@ -97,6 +102,38 @@ def penn2morphy(penntag: str) -> str:
...
+    def _process(self, input_pack: DataPack):
+        for sentence in input_pack.get(Sentence):
+            token_entries = list(input_pack.get(entry_type=Token,
+                                                range_annotation=sentence,
+                                                component=self.token_component))
+            tokens = [(token.text, token.pos) for token in token_entries]
+            cs = self.chunker.parse(tokens)
+            iob_tagged = tree2conlltags(cs)
+            for token, iob_tag in zip(token_entries, iob_tagged):
+                token.set_fields(chunk=iob_tag[2])
```
hunterhector: Let's create a Phrase object in our base_ontology to hold the chunks, that way the end user don't need to go over the IOB tags.

forte/processors/nltk_processors.py
```diff
@@ -97,6 +102,48 @@ def penn2morphy(penntag: str) -> str:
...
+                if hasattr(chunk, 'label'):
+                    begin_pos = token_entries[index].span.begin
+                    end_pos = token_entries[index + len(chunk) - 1].span.end
+                    phrase = Phrase(input_pack, begin_pos, end_pos)
+                    kwargs_i = {"phrase_type": chunk.label()}
+                    phrase.set_fields(**kwargs_i)
+                    input_pack.add_or_get_entry(phrase)
+                    index += len(chunk)
+                else:
+                    index += 1
```
hunterhector: Is this assuming the len of the chunk is 1? Is it better if we do `index += len(chunk)` here as well?

gpengzhi: I think this is correct. When the chunk has no `label` attribute, it contains only one token.

PR #123, opened 2020-01-14
Add NLTKNER

gpengzhi:


forte/processors/nltk_processors.py
```diff
@@ -111,3 +112,24 @@ def _process(self, input_pack: DataPack):
...
+            token_entries = list(input_pack.get(entry_type=Token,
+                                                range_annotation=sentence,
+                                                component=self.token_component))
+            tokens = [(token.text, token.pos) for token in token_entries]
+            ne_tree = ne_chunk(tokens)
+
+            ners = [chunk.label() if hasattr(chunk, 'label') else ''
+                    for chunk in ne_tree]
+            for token, ner in zip(token_entries, ners):
+                token.set_fields(ner=ner)
```
hunterhector: Similar to the chunking case, setting another attribute to the token does not fit into Forte's span based design principle. You can find a proper implementation here.

forte/processors/nltk_processors.py
```diff
@@ -111,3 +112,32 @@ def _process(self, input_pack: DataPack):
...
+                if hasattr(chunk, 'label'):
+                    begin_pos = token_entries[index].span.begin
+                    end_pos = token_entries[index + len(chunk) - 1].span.end
+                    entity = EntityMention(input_pack, begin_pos, end_pos)
+                    kwargs_i = {"ner_type": chunk.label()}
+                    entity.set_fields(**kwargs_i)
+                    input_pack.add_or_get_entry(entity)
+                    index += len(chunk)
+                else:
+                    index += 1
```
hunterhector: Similarly, are you sure adding 1 is OK?

gpengzhi: Yes, I think it is correct.

PR #127, opened 2020-01-15
fix accumulation of tokens in spacy_processors

swapnull7:


forte/processors/spacy_processors.py
```diff
@@ -97,4 +97,4 @@ def _process(self, input_pack: DataPack):
                         token.set_fields(lemma=word.lemma_)
 
                     tokens.append(token)
```
AvinashBukkittu: If we are not using `tokens`, no need to create it.

PR #131, opened 2020-01-17
Add performance_test decorator

gpengzhi:


AvinashBukkittu:
Thanks for doing this! LGTM

PR #134, opened 2020-01-23
Remove unnecessary html code

gpengzhi:
Our repository has 57.5% HTML code which is misleading.


forte/data/data_utils.py
```diff
@@ -20,31 +20,22 @@
...
 
-# pylint: disable=unused-argument,function-redefined,missing-docstring
-
-
-def maybe_create_dir(dirname: str) -> bool:
-    r"""Creates directory if it does not exist.
 
-    Args:
-        dirname (str): Path to the directory.
+Py3 = sys.version_info[0] == 3
```
AvinashBukkittu: Where are we using this variable?

forte/utils/utils_io.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2019 The Forte Authors. All Rights Reserved.
```
AvinashBukkittu: I like this `utils_io.py` file here. However, there is another similarly named `io_utils.py` in `forte/data`. I am wondering if we can consolidate the IO related functions at one place.

forte/data/data_utils.py
gpengzhi: Deleted.

forte/utils/utils_io.py
gpengzhi: Delete `ensure_dir` and rename `io_utils` to `data_utils_io`

PR #136, opened 2020-01-27
Add embedding to Entry

gpengzhi:


gpengzhi:
```python
class Generic(Entry):
    def __init__(self, pack: PackType):
        super(Generic, self).__init__(pack=pack)
```

```python
class Entry(Generic[ContainerType]):
    def __init__(self, pack: ContainerType):
```

from Hector: "
2 main reasons: 1) Generic class use PackType in the _init_, which is s a more useful type
2) The code gen need to analyze the classes to inherited from, now it only need to analyze top.py. It would be more difficult to analyze also the core.py. (edited) 
The ContainerType argument in the Entry will be very annoying to analyze."

AvinashBukkittu:
> ```python
> class Generic(Entry):
>     def __init__(self, pack: PackType):
>         super(Generic, self).__init__(pack=pack)
> ```
> 
> ```python
> class Entry(Generic[ContainerType]):
>     def __init__(self, pack: ContainerType):
> ```
> 
> from Hector: "
> 2 main reasons: 1) Generic class use PackType in the _init_, which is s a more useful type
> 2) The code gen need to analyze the classes to inherited from, now it only need to analyze top.py. It would be more difficult to analyze also the core.py. (edited)
> The ContainerType argument in the Entry will be very annoying to analyze."

I see. I'm wondering if this is a good time to rename `Generic` to something less confusing? I am fine with lengthy ones as well as long as it won't clash with `typing.Generic`.

hunterhector:
> > ```python
> > class Generic(Entry):
> >     def __init__(self, pack: PackType):
> >         super(Generic, self).__init__(pack=pack)
> > ```
> > 
> > 
> > ```python
> > class Entry(Generic[ContainerType]):
> >     def __init__(self, pack: ContainerType):
> > ```
> > 
> > 
> > from Hector: "
> > 2 main reasons: 1) Generic class use PackType in the _init_, which is s a more useful type
> > 2) The code gen need to analyze the classes to inherited from, now it only need to analyze top.py. It would be more difficult to analyze also the core.py. (edited)
> > The ContainerType argument in the Entry will be very annoying to analyze."
> 
> I see. I'm wondering if this is a good time to rename `Generic` to something less confusing? I am fine with lengthy ones as well as long as it won't clash with `typing.Generic`.

I agree that we should rename it. Do you have any proposals? How about `General`, `Generics` or `Info`, ``

AvinashBukkittu:
> > > ```python
> > > class Generic(Entry):
> > >     def __init__(self, pack: PackType):
> > >         super(Generic, self).__init__(pack=pack)
> > > ```
> > > 
> > > 
> > > ```python
> > > class Entry(Generic[ContainerType]):
> > >     def __init__(self, pack: ContainerType):
> > > ```
> > > 
> > > 
> > > from Hector: "
> > > 2 main reasons: 1) Generic class use PackType in the _init_, which is s a more useful type
> > > 2) The code gen need to analyze the classes to inherited from, now it only need to analyze top.py. It would be more difficult to analyze also the core.py. (edited)
> > > The ContainerType argument in the Entry will be very annoying to analyze."
> > 
> > 
> > I see. I'm wondering if this is a good time to rename `Generic` to something less confusing? I am fine with lengthy ones as well as long as it won't clash with `typing.Generic`.
> 
> I agree that we should rename it. Do you have any proposals? How about `General`, `Generics` or `Info`, ``

I guess `Generics` sounds okay. I not able to find better alternatives.

AvinashBukkittu:
Looks good to me.

PR #138, opened 2020-01-27
Polish processor.base

gpengzhi:


gpengzhi:
> I still see many functions take Hparams as config, which one exactly did you fix?

This PR just fix some docstring. I will open another PR to fix `hparams` issue.

PR #144, opened 2020-01-30
Add unit test for EntryContainer

gpengzhi:


forte/data/container.py
```diff
@@ -11,13 +11,21 @@
...
 from forte.data.span import Span
 from forte.process_manager import ProcessManager
 
+__all__ = [
+    "IdManager",
+    "EntryContainer",
+]
+
 E = TypeVar('E')
 L = TypeVar('L')
```
AvinashBukkittu: I see that `L` and `G` are used in the `EntryContainer` class. But the types `L` and `G` are not used anywhere in the class. I am wondering do we need it?

hunterhector: I am actually not very familiar with Python's generic types. Here I am just trying to say there are 3 types, which is later extended by class BasePack(EntryContainer[EntryType, LinkType, GroupType]). If we can simplify this and everything still works, then it will be fine.

PR #148, opened 2020-02-01
Update Readme

AvinashBukkittu:
This PR

- Updates README: Adds description, example code and installation instructions. We need to add reference section later.


hunterhector:
Let me merge this now and let's keep improving it.

PR #156, opened 2020-02-03
Remove duplicated pytest

### LINKED ISSUE #155: Are we running tests twice in the CI?
Not fully sure but the 2 commands here run all the tests twice.
https://github.com/asyml/forte/blob/master/.travis.yml#L47-L50

The second command `coverage run -m pytest` is sufficient to run both the tests and the coverage. Can we remove the `pytest` command?

gpengzhi:
resolve #155


AvinashBukkittu:
OK

PR #157, opened 2020-02-03
Remaining changes for HParams -> configs

atif93:
Some of the changes to make things compatible with #140.


atif93:
There are some of the changes in batchers, model_utils which I'm not sure about. Can you guys make sure those places also make sense?

AvinashBukkittu:
#140 should be causing several issues - Reason being most of the processors do not define `default_configs` in the way they should be used. The point of concern is examples. I think that if the CI is passing, we can let this merge.

PR #162, opened 2020-02-05
Add sentiment analyzer using VaderSentiment

hunterhector:
This PR adds a sentiment analyzer to Forte using VaderSentiment:

1. Add a sentiment field to the Sentence entry
1. The processor will add a dictionary of scores into the sentiment field.
1. Added the Vader dependency in travis
1. Tested by comparing the results of the sentiment scores, inside a pipeline.


forte/processors/sentiment_analysis.py
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__all__ = [
+    "VaderSentimentProcessor",
+]
+
+from texar.tf import HParams
```
AvinashBukkittu: We can also use `texar.torch` here. I think CI may be failing because of this since we don't install texar-tf in the CI.

AvinashBukkittu:
Looks good. Can be merged once the CI passes.

forte/processors/sentiment_analysis.py
hunterhector: fixed in 88a8b2b6920e1216d489fda5640cb7be699a4f7c

PR #163, opened 2020-02-07
Remove the unnecessary functions in BaseReader 

gpengzhi:
Delete `serialize_instance` and `deserialize_instance`.


forte/data/readers/base_reader.py
```diff
@@ -29,12 +30,12 @@
 from forte.data.multi_pack import MultiPack
 from forte.pipeline_component import PipelineComponent
 from forte.process_manager import ProcessManager
-from forte.utils import get_full_module_name
+from forte.utils.utils import get_full_module_name
```
AvinashBukkittu: Why did we make `get_full_module_name` go one-level deeper?

forte/data/readers/base_reader.py
```diff
@@ -270,26 +237,24 @@ def cache_data(self,
...
+            cache_filename: Path to the cache file.
 
         Returns: List of cached data packs.
-
         """
         logger.info("reading from cache file %s", cache_filename)
         with cache_filename.open("r") as cache_file:
             for line in cache_file:
-                pack = self.deserialize_instance(line.strip())
+                pack = jsonpickle.decode(line.strip())
```
AvinashBukkittu: I am wondering instead of `jsonpickle.decode(...)` we can provide a class method inside `MultiPack` and `DataPack` for deserialize. The reason is - if we serialize packs using `pack.serialize(...)` we would like to deserialize it using `DataPack.deserialize(...)` or `MultiPack.deserialize(...)` because the serialization and deserialization should be handled by the pack.

forte/data/readers/deserialize_reader.py
```diff
@@ -36,4 +38,4 @@ def _collect(self, *args, **kwargs) -> Iterator[str]:
         yield from data_packs
 
     def _parse_pack(self, data_source: str) -> Iterator[DataPack]:
-        yield self.deserialize_instance(data_source)
+        yield jsonpickle.decode(data_source)
```
AvinashBukkittu: Same logic here.

forte/data/readers/base_reader.py
gpengzhi: We made a `utils` folder and moved `utils.py` into that folder.

AvinashBukkittu: Yeah, I am wondering if we can pull it up by one level through `__init__.py`. `forte.utils.utils` is a bit verbose.

gpengzhi: When I import something in our repo, I try to import it from its own file so that cycling import won't be triggered.

AvinashBukkittu: I see. If this could trigger cyclic imports, we can leave it as is.

gpengzhi: I do not think this one could trigger cyclic imports, but I think this is a good import style when we import something in our repo.

PR #164, opened 2020-02-07
Fix the documentation issues in reader module

gpengzhi:


AvinashBukkittu:
@gpengzhi I noticed that all the tests in `forte/data/readers` are moved out of `tests` in this PR. This may be a bit confusing. After discussing with @hunterhector, it makes sense to have a separate folder for them. This is followed for `forte/processors`. Would you mind moving them to `tests/` in `readers` directory in this same PR?

PR #169, opened 2020-02-11
Move nltk from install_requires to extras_require

gpengzhi:
`nltk` is a 3rd party package such that we should not put it in `install_requires`.


atif93:
Are we sure that `nltk` was only being used in the processors?

gpengzhi:
> Are we sure that `nltk` was only being used in the processors?

Currently, we only import `nltk` in `nltk_processors.py`.

setup.py
```diff
@@ -37,6 +36,7 @@
         'texar-pytorch',
```
AvinashBukkittu: I see that we install `texar-pytorch` and `texar`. To my knowledge, we use only `texar-pytorch`. For simplicity, we can remove `texar`

PR #173, opened 2020-02-27
Ontology features: Managing imports, fixing getter and setter, dictionary support

mgupta1410:


hunterhector:
Will merge through a different way.

PR #183, opened 2020-03-17
Adding __init__.py to generated packages

mgupta1410:


tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -101,13 +94,51 @@ def test_dry_run_false(self):
...
+        gen_files = sorted(utils.get_generated_files_in_dir(folder_path))
+
+        # Assert the generated python files
+        exp_file_path = ['ft/__init__',
+                         'ft/onto/__init__',
+                         'ft/onto/example_import_ontology']
+        exp_files = sorted([f"{os.path.join(folder_path, file)}.py"
+                           for file in exp_file_path])
+
+        self.assertCountEqual(gen_files, exp_files)
```
hunterhector: it is more accurate to compare the list content (after sorted)

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -101,13 +94,51 @@ def test_dry_run_false(self):
...
+        self.generator = OntologyCodeGenerator()
+        folder_path = self.generator.generate(temp_filename, folder_path,
+                                              is_dry_run=False,
+                                              include_init=False)
+        gen_files = sorted(utils.get_generated_files_in_dir(folder_path))
+
+        # Assert the generated python files after removing the corrupted file
+        # which should not have been regenerated
+        exp_files = [file for file in exp_files if file != corrupted_path]
+        self.assertCountEqual(gen_files, exp_files)
```
hunterhector: it is more accurate to compare the list content (after sorted)

forte/ontology_specs/wikipedia_ontology.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "wikipedia",
   "imports": [
-    "ft.onto.base_ontology"
+    "base_ontology.json"
```
hunterhector: Is this .json?

hunterhector: Further, how can one distinguish a package from a file path? From the string surface, it is impossible. I would rather suggest use two separate keys.

mgupta1410: Yes, it should be .json, in this case, we would want to use base_ontology.json from the same directory, and not some other base_ontology version that might be installed. 

In the next commit, I have added separate keys for filepath and package.

tests/forte/data/ontology/ontology_code_generator_test.py
mgupta1410: Done.

tests/forte/data/ontology/ontology_code_generator_test.py
mgupta1410: Done.

hunterhector:
Let's use another PR to move the "installed" version of the "qa_ontology" out.

PR #184, opened 2020-03-18
Solve some problems in serialization

hunterhector:
The previously generated ontology codes make mistakes in "get_state" and "set_state", which breaks serialization. This PR fixes this with the following:

1. Correcting the get_state and set_state
1. Do not call `set_fields` directly now, replaced all `set_fields` call with proper getter and setter
1. Correct corresponding generated files
1. Write a test for serialization and deserialization


hunterhector:
Prevous "Sqaush and merge" makes Git unaware of all the commits that are already merged. This time we will create an actual merge commit instead.

PR #186, opened 2020-03-21
Adding search (indexer+reranker) inference pipeline, passage reranking bert model

mgupta1410:


examples/passage_reranker/bert.py
```diff
@@ -0,0 +1,237 @@
...
+                         cache_dir=self.cache_dir)
+
+        self.init_pretrained_weights(prefix=self._hparams.prefix)
+
+    def default_hparams(self):
+        params = super().default_hparams()
+        params.update(self.kwargs)
+        return params
+
+    def load_pretrained_config(self,
```
hunterhector: This looks complex even if we are using Texar. Why is this? Can Texar handle most of the model loading?

examples/passage_reranker/config.yml
```diff
@@ -15,7 +18,71 @@ indexer:
     hosts: "localhost:9200"
     algorithm: "bm25"
 
+
```
hunterhector: So we don't have to demonstrate the NER part here again, I guess. We have demonstrated in several parts that we can combine ranking + analysis already. Let's just focus on ranking in this case.

examples/passage_reranker/indexer_reranker_inference_pipeline.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Indexing and Re-ranking
+    nlp.add_processor(processor=ElasticSearchQueryCreator(),
+                      config=config.query_creator)
+    nlp.add_processor(processor=ElasticSearchProcessor(),
+                      config=config.indexer)
+    nlp.add_processor(processor=BertRerankingProcessor(),
+                      config=config.reranker)
+
+    # NER Tagging
```
hunterhector: Let's focus on the reranking demo here and leave out the NER part, in this example

PR #217, opened 2020-08-01
Add the Clinical NER example.

ha-lins:
Adds the example of NER for clinical text record. This example

1. uses custom ontology for medical application.
2. can employ ner for clinical text.


examples/Cliner/download_model.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+wget -P ../CliNER/models \ https://drive.google
```
hunterhector, line 3: Will these files be deleted in the future otherwise we can move it to a permanent place.

In addition, I think it is better to upload all binary files here, any pickle or wheel files.

examples/Cliner/generation.json
```diff
@@ -0,0 +1 @@
..."}}, {\"py/object\": \"ft.onto.clinical.ClinicalEntityMention\", \"py/state\": {\"_span\": {\"py/object\": \"forte.data.span.Span\", \"begin\": 2968, \"end\": 2984}, \"_tid\": 49, \"cliner_type\": \"treatment\"}}, {\"py/object\": \"ft.onto.clinical.ClinicalEntityMention\", \"py/state\": {\"_span\": {\"py/object\": \"forte.data.span.Span\", \"begin\": 3002, \"end\": 3019}, \"_tid\": 50, \"cliner_type\": \"treatment\"}}, {\"py/object\": \"ft.onto.clinical.ClinicalEntityMention\", \"py/state\": {\"_span\": {\"py/object\": \"forte.data.span.Span\", \"begin\": 3048, \"end\": 3057}, \"_tid\": 51, \"cliner_type\": \"treatment\"}}, {\"py/object\": \"ft.onto.clinical.ClinicalEntityMention\", \"py/state\": {\"_span\": {\"py/object\": \"forte.data.span.Span\", \"begin\": 3081, \"end\": 3088}, \"_tid\": 52, \"cliner_type\": \"treatment\"}}, {\"py/object\": \"ft.onto.clinical.ClinicalEntityMention\", \"py/state\": {\"_span\": {\"py/object\": \"forte.data.span.Span\", \"begin\": 3126, \"end\": 3137}, \"_tid\": 53, \"cliner_type\": \"treatment\"}}], \"generics\": [], \"replace_back_operations\": [], \"processed_original_spans\": [], \"orig_text_len\": 3337, \"serialization\": {\"next_id\": 54}}}"
```
hunterhector, line 1: You have serialized the file wrong. A correctly serialized JSON file will not have all the backslashes.

forte/ontology_specs/clinical.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "clinical",
+  "imports": [
+    "forte/ontology_specs/base_ontology.json"
```
hunterhector, line 4: I think the import like the following would work:
`
 "imports": [
    "base_ontology.json"
  ],
`

forte/ontology_specs/clinical.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "definitions": [
+    {
+      "entry_name": "ft.onto.clinical.ClinicalEntityMention",
```
hunterhector, line 8: It is better to extend `ft.onto.base_ontology.EntityMention`, and then you don't have to provide the attribute `cliner_type`, simply use the `ner_type` attribute from the root project.

.travis.yml
```diff
@@ -3,63 +3,65 @@ language: python
 dist: xenial
 cache: pip
 python:
```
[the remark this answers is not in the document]
ha-lins: Yes, I'll restore it. 

But I wonder if I can add [this line](https://github.com/ha-lins/forte/blob/CliNER/.travis.yml#L32) in this file to install the hypothesis package, which is in order to enable passing the pytest.

docs/conf.py
```diff
@@ -240,17 +240,17 @@
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {
-# The paper size ('letterpaper' or 'a4paper').
-#'papersize': 'letterpaper',
+    # The paper size ('letterpaper' or 'a4paper').
```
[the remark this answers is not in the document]
ha-lins: Fixed.

.travis.yml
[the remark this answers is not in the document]
hunterhector: So how to add the dependency actually depends on the case. But in this case, I think you are trying to add a dependency during running the example. I would suggest to add it in the setup.py, more specifically, the following line:

https://github.com/ha-lins/forte/blob/CliNER/setup.py#L47

forte/ontology_specs/clinical.json
hunterhector: Could you extend `ft.onto.base_ontology.EntityMention` instead?

docs/conf.py
```diff
@@ -12,23 +12,21 @@
...
-# from unittest.mock import MagicMock
+import os
+#from unittest.mock import MagicMock
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
+from forte import __version__
+from forte.version import VERSION_SHORT as __version_short__
```
hunterhector: Please group the imports together at the top of the file.

And please do not make changes to all the lines that are you haven't really touched below, such as those spaces after the #.

examples/Cliner/CliNER/code/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+__version__ = '0.1dev'
```
hunterhector, line 3: I think you should remove the version here.

forte/ontology_specs/clinical.json
ha-lins: Fixed. Pls check again.

docs/conf.py
ha-lins: Yes, I removed them in an early commit possibly in order to pass the pylint check. And I restore them in the latest commit.

examples/Cliner/CliNER/.gitattributes
```diff
@@ -0,0 +1,22 @@
+# Auto detect text files and perform LF normalization
```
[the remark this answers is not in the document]
ha-lins: I removed these git files, including .gitattributes and .gitattributes.

examples/Cliner/CliNER/.gitignore
```diff
@@ -0,0 +1,50 @@
+
```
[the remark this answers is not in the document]
ha-lins: I have removed it in CliNER.

examples/Cliner/CliNER/requirements.txt
```diff
@@ -0,0 +1,7 @@
+nltk
```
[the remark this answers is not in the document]
ha-lins: I have removed these dependencies in `setup.py` in the latest commit.

examples/Cliner/generation.json
ha-lins: It is updated and the backslashes have gone.

forte/ontology_specs/clinical.json
ha-lins: Yes, fixed.

examples/Cliner/pipeline.py
```diff
@@ -28,13 +27,10 @@
 
 
 def do_process(input_pack_str: str):
-    # Let's assume there is a JSON string for us to use.
     datapack: DataPack = pipeline.process([input_pack_str])
-    # You can get the JSON form like this.
     data_json = datapack.serialize()
-    # Let's write it out.
     with open('generation.json', 'w') as fo:
-        fo.write(json.dumps(data_json, indent=2))
```
ha-lins: It seems that those backslashes are caused by the indent parameter.

tests/forte/processors/cliner_processor_test.py
```diff
@@ -0,0 +1,77 @@
...
+                  'w') as fo:
+            fo.write(document)
+        pack = self.cliner.process(document)
+
+        entities_entries = list(pack.get(entry_type=ClinicalEntityMention))
+
+        entities_text = [x.text for x in entities_entries]
+        entities_type = [x.ner_type for x in entities_entries]
+
+        self.assertEqual(entities_text, [
```
ha-lins, line 68: The test case passed in my current environment yet I haven't try to run it in a new environment.

hunterhector: We can consider it passed if travis passes. Thanks for adding the test.

However, the traivs seems to failed for something else:

https://travis-ci.org/github/asyml/forte/builds/718284267

examples/Cliner/pipeline.py
hunterhector: I don't think so. This is most likely that you encode json twice. Once you have converted the object into json string, if you encode it again, the json library will escape all special characters.

examples/Cliner/pipeline.py
```diff
@@ -0,0 +1,51 @@
...
+from forte.pipeline import Pipeline
+
+# Let's create a pipeline first.
+pipeline = Pipeline[DataPack]()
+
+
+def do_process(input_pack_str: str):
+    datapack: DataPack = pipeline.process([input_pack_str])
+    data_json = datapack.serialize()
+    with open('generation.json', 'w') as fo:
```
hunterhector, line 32: So here you basically implement the writer yourself.  I think this is probably the reason for double json encoding as well. 

Actually, for serialization you can directly call a writer without doing this. You can follow the example here:

https://github.com/asyml/forte/blob/master/examples/serialization/serialize_example.py#L104

examples/Cliner/CliNER/code/predict.py
```diff
@@ -0,0 +1,104 @@
...
+
+import glob
+import os
+import pickle
+import sys
+
+from examples.Cliner.CliNER.code import tools
+from examples.Cliner.CliNER.code.notes.documents import Document
+
+import helper_dataset as hd
```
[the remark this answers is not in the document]
ha-lins: Yes, I think it can work.

examples/Cliner/Cliner.py
```diff
@@ -40,7 +44,8 @@ def _process(self, input_pack: DataPack):
 
         self.model.predict()
 
-        con = codecs.open(os.path.join(self.output, 'ex_doc.con'),
+        fname = os.path.splitext(os.path.basename(self.txt))[0] + '.' + 'con'
+        con = codecs.open(os.path.join(self.output, fname),
```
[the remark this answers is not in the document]
ha-lins: Fixed with `with open(...) as ...`.

examples/Cliner/reader.py
```diff
@@ -30,17 +30,18 @@ def _parse_pack(self, collection) -> Iterator[DataPack]:
         txt_path = collection
 
         pack = self.new_pack(pack_name='Cliner_input')
-        doc = codecs.open(txt_path, "r", encoding="utf8")
+        doc = codecs.open(txt_path, "rb+", encoding="utf8")
```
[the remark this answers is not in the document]
ha-lins: Fixed with `with open(...) as ...`.

examples/Cliner/pipeline.py
ha-lins: I have changed it to call the writer, while it is written to the output/cliner_input.json instead of a new generation.json. I think it doesn't matter.

I will check the code again according to the json and then open a new branch.

hunterhector: So the writer, when provided correct parameters, will write out documents in that directory using its "pack_name". So if you set the `pack_name` somewhere it will result in that file name.  https://github.com/asyml/forte/blob/dbf04edb3e7e58e59022e324b87a182c26c93de7/forte/processors/writers.py#L31

hunterhector:
I think most problems are gone. As discussed before, let's have a different PR. I will close this PR when that one is merged.

PR #235, opened 2020-09-07
Add the SST2(Stanford Sentiment Analysis) reader

### LINKED ISSUE #224: Create Reader for Stanford Sentiment Treebank
This feature will create a simple reader to read the SST-2 dataset. https://nlp.stanford.edu/sentiment/treebank.html

1. Add sentiment attribute to the parse tree node ontology.
2. Read in the treebank data as Forte format.
(Some readers examples are here: https://github.com/asyml/forte/tree/master/forte/data/readers)
3. Try to create a pipeline by chaining an existing sentiment analysis system. (https://github.com/asyml/forte/blob/master/forte/processors/sentiment_analysis.py)
A pipeline example can be found here: 
https://github.com/asyml/forte/blob/master/examples/pipelines/process_dataset_example.py
4. Write test cases for this reader. Here is an example:
https://github.com/asyml/forte/blob/master/tests/forte/data/readers/race_multi_choice_qa_reader_test.py

N/A

**Additional context**

haoyuLucas:
This PR fixes #224 . 

### Description of changes
This PR creates a reader for the SST2(Stanford Sentiment Treebank) dataset. 
1. Includes SST2 dataset to the data_samples
2. Adds a ConstituentNode ontology to base_ontology.json & base_ontology.py
3. Writes a reader which builds the constituency tree from parent pointer list and align with the text spans & sentiments
4. Test the reader with specific testcases

The Constituency Tree Node ontology can be used in the future.

### Test Conducted
The testcase is an input sentence from the dataset:
"Effective but too-tepid biopic".


hunterhector (changes requested):
1. If you include some sample data, you can include just a small sample. Here you seem to put the whole dataset here. This is not advised due to multiple reasons. The main problem is that these files are actually quite large. Note that our codebase is only a few MBs and these data will double our repository size. In addition, once you commit data to Github, you cannot get rid of them because Git will keep the whole history. So now, even if we remove these files from this branch, they are likely to be in the GIt history forever.  Here is what you need to do **please create a new branch from master,  add you commits without the data set, and make a new PR. We will close this PR.**
2. If you want to use the whole dataset, you can create a download script to do that. Here is an example: https://github.com/asyml/forte/blob/master/examples/passage_ranker/download_data.py
3. Let's get used to the `Typing` feature provided by Python 3. This is one of our coding styles and we will use automatic tools to check if we have added enough typing. Please annotate the variable types as much as possible, most importantly the method signature. and also most of the variables that you are using. Also, avoid using generic types such as `Any`. Our general rule-of-thumb is only not annotating the type when you cannot do so.

Please re-create the PR without the dataset, and setup the PR again, remember to update the master branch first, the master branch now has the `Constituent` code.

hunterhector:
Closing this because this will be merged in https://github.com/asyml/forte/pull/236

PR #238, opened 2020-09-09
Ag news reader

### LINKED ISSUE #226: Create a reader for the AG Corpus
This feature will create a simple reader to read AG Corpus. 
http://groups.di.unipi.it/~gulli/AG_corpus_of_news_articles.html

1. Create an appropriate metadata field for the AG fields using the ontology.
1. Read in the treebank data as Forte format.
(Some readers examples are here: https://github.com/asyml/forte/tree/master/forte/data/readers). You will save the sentiment in one of the attributes. Currently, we have a sentence level sentiment attribute (https://github.com/asyml/forte/blob/master/ft/onto/base_ontology.py#L100)
1. Write test cases for this reader. Here is an example:
https://github.com/asyml/forte/blob/master/tests/forte/data/readers/race_multi_choice_qa_reader_test.py

N/A

**Additional context**

jrxk:
This PR fixes #226 . 

### Description of changes
This PR creates a reader for the [AG News](https://github.com/mhjabreel/CharCnn_Keras/tree/master/data/ag_news_csv) dataset

Add ontology and reader for AG News.

### Test Conducted
Add a unit test for the reader with samples from the AG News dataset.


forte/data/readers/ag_news_reader.py
```diff
@@ -0,0 +1,107 @@
...
+        Args:
+            csv_file: A CSV file path.
+
+        Returns: Iterator of each line in the csv file.
+        """
+        with open(csv_file, "r") as f:
+            for line_id, line in enumerate(f):
+                yield (line_id, line)
+
+    def _cache_key_function(self, data_pack: DataPack) -> str:
```
hunterhector, line 60: See discussion here https://github.com/asyml/forte/pull/237#discussion_r485792965

forte/ontology_specs/ag_news.json
```diff
@@ -0,0 +1,32 @@
...
+    "imports": [
+      "base_ontology.json"
+    ],
+    "definitions": [
+      {
+        "entry_name": "ft.onto.ag_news.Article",
+        "parent_entry": "ft.onto.base_ontology.Document",
+        "attributes": [
+            {
+              "name": "class_id",
```
hunterhector, line 12: I don't think we should have a `class_id`. We can derive this id from the name, so this information is a bit redundant.

forte/ontology_specs/ag_news.json
```diff
@@ -0,0 +1,32 @@
...
+              "description": "The class ids are numbered 1-4 where 1 represents World, 2 represents Sports, 3 represents Business and 4 represents Sci/Tech.",
+              "type": "int"
+            },
+            {
+                "name": "class_name",
+                "description": "The name of the class.",
```
hunterhector, line 18: The description is not informative enough, this basically provides the same amount of information as the `name` itself.

forte/ontology_specs/ag_news.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "ag_news",
+    "imports": [
+      "base_ontology.json"
+    ],
+    "definitions": [
+      {
+        "entry_name": "ft.onto.ag_news.Article",
```
hunterhector, line 8: Interesting, so you choose to extend the base ontology. This is a good idea! One thing I would suggest is that we can actually add a `document_class` attribute the the `base_ontology`, since it is common to have a category for a document.  To make things general, you may consider to use a list of strings for that variable. In the AG news corpus, we happen to use just one of them.

forte/data/readers/ag_news_reader.py
```diff
@@ -0,0 +1,107 @@
...
+        Description(pack, description_start, description_end)
+
+        pack.pack_name = line_id
+        yield pack
+
+    @classmethod
+    def default_configs(cls):
+        config: dict = super().default_configs()
+
+        # Add OIE dataset file extension. The default is '.oie'
```
hunterhector, line 98: Your comment here is wrong.

forte/data/readers/ag_news_reader.py
```diff
@@ -0,0 +1,107 @@
...
+        config.update({
+            'class_names': {
+                1: 'World',
+                2: 'Sports',
+                3: 'Business',
+                4: 'Sci/Tech'
+            }
+        })
+        return config
```
hunterhector, line 107: New line at the end of file is a recommended practice.

hunterhector (changes requested):
Overall looks good, it seems like you know how to write a reader now.

Please resolve some comments, update your branch, and fix formatting errors raised by Travis. You can run pyling, mypy, flake8 yourself to see what happen.

forte/ontology_specs/ag_news.json
jrxk: Okay. I will just store the class names in the config, then.

hunterhector (approved):
I think this looks good to me overall. I think it would be OK if the travis passes.

forte/data/readers/ag_news_reader.py
```diff
@@ -0,0 +1,103 @@
...
+from ft.onto.base_ontology import Document
+
+__all__ = [
+    "AGNewsReader",
+]
+
+
+class AGNewsReader(PackReader):
+    r""":class:`AGNewsReaderr` is designed to read in AG News
+    text classification dataset.
```
hunterhector, line 33: Include the links and citations for this dataset, so people know where to find them. You can also include how to run the reader on the downloaded data (which directory, file to pass into the reader, for example.)

forte/data/readers/ag_news_reader.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2019 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: It is 2020 now.

hunterhector (changes requested):
Just need to improve the documentation for the reader.

forte/ontology_specs/ag_news.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "ag_news",
+    "imports": [
+      "base_ontology.json"
+    ],
+    "definitions": [
+      {
+        "entry_name": "ft.onto.ag_news.Title",
```
hunterhector, line 8: I think `Title` can be a common type that can be included in `base_ontology`.

tests/forte/data/readers/ag_news_reader_test.py
```diff
@@ -0,0 +1,84 @@
...
+        self.pipeline.set_reader(AGNewsReader())
+        self.pipeline.initialize()
+
+        self.sample_file: str = os.path.abspath(os.path.join(
+            os.path.dirname(os.path.realpath(__file__)),
+            *([os.path.pardir] * 4),
+            'data_samples/ag_news/sample.csv'))
+
+        self.expected_content: Dict[int, str] = {}
+        with open(self.sample_file, 'r') as file:
```
hunterhector, line 43: If you simply one sample here, sometimes it is easier to put the expected text in raw string.  This can help avoid you making the same mistake in the reader and the test case.

hunterhector (changes requested):
Some minor comments.

hunterhector:
Please fix the residual comments and let's merge this.

forte/ontology_specs/ag_news.json
jrxk: Sure, I'll fix that

tests/forte/data/readers/ag_news_reader_test.py
jrxk: Thanks for the suggestion. But here I have multiple samples in one file. Since the expected content is generated differently from the data reader, it's not likely to make the same mistake.

hunterhector: OK, I don't think this is a crucial point to fix anyway.

hunterhector (approved):

PR #245, opened 2020-09-11
Data Augmentation Base Class

### LINKED ISSUE #241: Design the interface of base augmenter
This request will create a base class for data augmenter.

The base data augmenter takes in a data pack and returns an augmented multi-pack. It will only define abstract methods for subclasses to implement.

haoyuLucas:
This PR fixes #241 . 

### Description of changes
This PR creates a base class of the data augmentation. It inherits from the class of Caster.

This base class will be inherited by different kinds of data augmenter.


forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2019 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: 2020 now

hunterhector (changes requested):
I cannot see that this class is anything different from the Boxer. Further, if you define the `cast` function to be abstract, then the child function will overwrite this, it means that the implementation in this `cast` function is not doing anything.

forte/processors/base/data_augment_processor.py
```diff
@@ -27,8 +28,11 @@
 
 class BaseDataAugmentProcessor(Caster[DataPack, MultiPack]):
     r"""The base class of processors that augments the data.
+    Pass the input data_pack to the "cast" method
+    and it will return an output Multipack.
+
+    To inherit from this class, please override the augment function.
```
hunterhector, line 34: The docstring here can still be improved. 

1. Users don't have to pass in data pack to `cast` themselves.
2. To refer a function in the docstring, see here https://stackoverflow.com/questions/11168178/how-do-i-reference-a-documented-python-function-parameter-using-sphinx-markup

The docstring here should explain what this processor do. "Augments the data" is too vague.  "return an output Multipack" is also not informational

forte/processors/base/data_augment_processor.py
```diff
@@ -40,11 +44,15 @@ def cast(self, pack: DataPack) -> MultiPack:
...
+            p.add_pack_(data_pack, pack_name)
         return p
 
-    @classmethod
-    def default_configs(cls):
-        return {
-            'pack_name': 'default'
-        }
+    @abstractmethod
+    def augment(self, pack: DataPack) -> Iterable[Tuple[str, DataPack]]:
```
hunterhector, line 54: The docstring here only says the programming behavior. But didn't explain it conceptually. Why it returns tuples of pack name and pack? What are the packs? What are the strings? 

You also haven't mentioned that this is an `Iterator`.

hunterhector (changes requested):
The docstring here is not helpful for the developer who tries to implement this. In fact, the docstring for a base class is far more important than normal methods. 

Take a look at some good docstring examples here.

https://github.com/asyml/forte/blob/master/forte/processors/base/batch_processor.py

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+        r"""
+        This function parse the augment algorithm and instantiate an augmenter.
+        The replacement level of processor must be supported by the augmenter.
+
+        For example, we cannot do sentence replacement with a DictionaryReplacementAugmenter,
+        because it is based on synonyms substitution.
+        :return: an instance of data augmenter.
+        """
+        algorithm = self.configs.augment_algorithm
+        if algorithm == "DictionaryReplacement":
```
hunterhector, line 51: Instead of using a string here, we should construct the augmented by directly pass in the an Augmenter object, so you don't have to import specific augmenters.

forte/processors/data_augment/algorithms/base_augmenter.py
```diff
@@ -0,0 +1,42 @@
...
+    For example, the replacement_level of synonym replacement is ["word"],
+    that of back-translation is ["word", "sentence"].
+    """
+    @property
+    @abstractmethod
+    def replacement_level(self) -> List[str]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def augment(self, input: str, additional_info: Dict[str, str]) -> str:
```
hunterhector, line 41: Like we discussed, this interface is not very meaningful, let's discuss.

hunterhector (changes requested):

hunterhector (commented):
See comments.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,92 @@
...
+            - other_entry_policy: the policy for other entries that
+            is not the "augment_entry".
+            If "Delete", all the other entries will be erased.
+            Otherwise, they will be kept as they were,
+            but they might become invalid after the augmentation.
+        """
+        config = super().default_configs()
+        config.update({
+            'augment_entry': "ft.onto.base_ontology.Sentence",
+            'other_entry_policy': "Delete"
```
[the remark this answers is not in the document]
haoyuLucas: Yes. Maybe we can change this to "Keep".

forte/processors/data_augment/algorithms/base_augmenter.py
```diff
@@ -0,0 +1,53 @@
...
+        self.configs = configs
+
+
+class ReplacementDataAugmenter(BaseDataAugmenter):
+    r"""
+    Most data augmentation algorithms can be considered as replacement-based
+    methods on different levels.
+    """
+    @abstractmethod
+    def augment(self, input: Entry, *args, **kwargs) -> str:
```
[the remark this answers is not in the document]
haoyuLucas: Resolved.

hunterhector (commented):
I think about your `base_augmenter`.  So if the purpose of this class is simply replacing the text, then it is not an `augmenter`, because `augmenter` need to handle other things, especially the labels. 

But I would say you can still keep that class, but probably rename it to something like `text_replacement_ops`. And then a full augmentation algorithm is done by combining these `text_replacement_ops` (can be more than one) and deal with the labels.

When calling the text_repacement_ops, the augmented should probably call via a helper function `replace(replacement_ops, begin, end)`, this helper function will do two things:

1. call  `replacement_ops.replace(text[begin: end]) -> replaced_text`
2. record the change of length here automatically, which can be used to readjust the pack information later.

This helper function can be implemented in `ReplacementDataAugmentProcessor`. The function the realign other entries can also be implemented here.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,92 @@
...
+            If "auto_align", the span will be automatically modified according
+            its original location. However, some spans might become invalid
+            after the augmentation, for example, the tokens within a replaced
+            sentence may disappear.
+            If "delete", the entry will not be copied to the new data pack. It
+            is the default operation for entries not specified in this dict.
+
+            Example: {
+                "ft.onto.base_ontology.Sentence": "auto_align",
+                "ft.onto.base_ontology.Document": "delete"
```
hunterhector, line 83: I think users won't need to define all entries, we simply not copy the items that the users don't provide a resolution here.

forte/processors/data_augment/algorithms/text_replacement_op.py
```diff
@@ -0,0 +1,47 @@
...
+]
+
+
+class TextReplacementOp(ABC):
+    r"""
+    The base class holds the data augmentation algorithm.
+    We leave the :func: replace method to be implemented
+    by subclasses.
+    """
+    def __init__(self, configs: Dict[str, Any]):
```
hunterhector, line 33: Let's use the `Config` object in Forte (which is the Hparam object in Texar) to keep interface slightly more consistent.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,92 @@
...
+        self.replaced_spans: List[Tuple[Entry, str]] = []
+
+    def replace(self, replacement_op: TextReplacementOp, input: Entry):
+        """
+        This is a wrapper function to call the replacement op. After
+        getting the augmented text, it will register the input & output
+        for later batch process of building the new data pack.
+        """
+        replaced_text: str = replacement_op.replace(input)
+        self.replaced_spans.append((input, replaced_text))
```
hunterhector, line 62: Now maybe we can use this information to do the alignment. It will be great if this processor calls the function automatically so that the implementation can focus on the algorithm logic.

By the way, we have a somewhat similar functionality implemented [here](https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L180), which is used when the users remove "noisy text" from the original data (i.e. HTML tags). Maybe it can help you with this function.

hunterhector (commented):
A few more comments.

hunterhector:
By the way, even though this is the base class, test cases are needed, do you have a plan for that?

forte/processors/base/data_augment_processor.py
haoyuLucas: Yes. As in the comment we said, the "delete" is the default operation for entries not specified in the dict.

hunterhector: Now this logic is not clear.

First, we discuss that we will not use the word "Delete", because 1) this make users feel there would be data loss 2) this is not the same as our code behavior. What the code is doing is simply "Not Copy".

Second, there is no need to provide a key for "not going to copy" entries. And in the documentation, we shouldn't say the default behavior is "Delete", we should simply say we will copy the entries that you specify here. When would the user ever need to specify "Delete"?

haoyuLucas: That makes sense. In this case, I don't think we need the dict as well. We only need a list, and all the entries in the list should be "auto_align". For entries not on the list, we do not copy them.

forte/processors/base/data_augment_processor.py
haoyuLucas: The batch replacement function(I am still working on) may be called in the _process function of the processor. I was also thinking about testing the base classes. I guess we can test it after finishing the batch replacement function. But maybe I will create another PR for that function, so that we can merge the base classes ASAP.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,115 @@
...
+            data_pack: Datapack holding the annotations to be replaced.
+            replaced_annotations: A list of tuples(annotation, new string).
+            The text for annotation will be updated with the new string.
+
+        Returns:
+            A new data_pack holds the text after replacement. The annotations
+            in the original data pack will be copied and auto-aligned as
+            instructed by the "other_entry_policy".
+
+        """
```
hunterhector, line 85: Oh, can you do this? Do you need to put a pass keyword here?

hunterhector (approved):
I don't have much, just one small comment

forte/processors/base/data_augment_processor.py
haoyuLucas: What is a pass keyword? I thought we can just access the config and get the "other_entry_policy" from it.

hunterhector: No, I am just wondering if  it is OK to leave the function empty. Normally I will use the `pass` statement: https://docs.python.org/2.0/ref/pass.html

hunterhector: Yeah, I just have a try, so having the docstring will serve the purpose, we don't need the pass statement.

forte/processors/base/data_augment_processor.py
```diff
@@ -83,6 +83,7 @@ def auto_align_annotations(
             instructed by the "other_entry_policy".
 
         """
+        pass
```
hunterhector, line 86: well I just learn that we don't need this, but it won't hurt as well. Feel free to merge after travis, I already approved

hunterhector (approved):

PR #258, opened 2020-09-22
Dictionary-based replacement

### LINKED ISSUE #242: Implement a data augmentation processor for text generation data and dictionary-based word replacement augmenter
1. Creates a DA processor for text generation data where each multipack consists
of two datapacks: input source and output target.
2. Create an augmenter based on dictionary synonym replacement.

1. The TextGenerationAugmentProcessor inherits the base data augment processor. It replaces the char/word/sentence and rebuild the datapack, inserting them to the original multipack.
2. This augmenter will load a dictionary(such as wordnet) and replace the word with synonyms.

haoyuLucas:
This PR fixes #242 . 

### Description of changes
1. Creates a DA processor for text generation data where each multipack consists of two datapacks: input source and output target. The TextGenerationAugmentProcessor inherits the base data augment processor. It replaces the char/word/sentence and rebuildsthe datapack, inserting them to the original multipack.
2. Create an augmenter based on dictionary synonym replacement. This augmenter will load a dictionary(such as wordnet) and replace the word with synonyms.


forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+from forte.processors.base.pack_processor import MultiPackProcessor
+__all__ = [
+    "BaseDataAugmentProcessor"
+]
+
+
+class BaseDataAugmentProcessor(MultiPackProcessor):
+    r"""The base class of processors that augment data.
+    This processor instantiates an augmenter where specific
+    data augmentation algorithms are embedded. The augmenter
```
hunterhector, line 30: "embedded" has some special meaning in ML, let's just say "implemented".

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+    will run the algorithms and the processor will pack the
+    strings.
```
hunterhector, line 32: "pack the strings"? This description is not very accurate and clear.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+
+    The DA methods can all be considered as replacement-based
+    methods with different levels: character, word, sentence.
```
hunterhector, line 35: I am a little not sure about whether they can "all be considered". Maybe we should not be so sure here. Btw, if you want to use the abbreviation form "DA", make sure you declare it before like this: Data Augmentation (DA)

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+        algorithm = self.configs.augment_algorithm
+        if algorithm == "DictionaryReplacement":
+            lang = self.configs.get("lang", "eng")
+            augmenter = DictionaryReplacementAugmenter(configs={"lang": lang})
+        else:
+            raise ModuleNotFoundError("The augment algorithm {} is not implemented!".format(algorithm))
+
+        if self.configs.replacement_level not in augmenter.replacement_level:
+            raise LookupError("The replacement level is not supported by the provided augmenter!")
+        return augmenter
```
hunterhector, line 59: Interesting, I see that you start to have a couple pre-defined configurations, such as algorithm name, and replacement_level, now it is a good time to start have documentation for them

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+    @classmethod
+    def default_configs(cls):
+        """
+        :return: A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - augment_algorithm: defines the augmenter to use
+            - augment_ontologies: defines the ontologies that will be returned
```
hunterhector, line 68: You don't return an "ontology", so what exactly is being returned here?

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+            raise LookupError("The replacement level is not supported by the provided augmenter!")
+        return augmenter
+
+
+    @classmethod
+    def default_configs(cls):
+        """
+        :return: A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - augment_algorithm: defines the augmenter to use
```
hunterhector, line 67: Here it is a good place to start writing down what algorithms are available.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Processors that augment the data.
+"""
+from abc import abstractmethod
+from forte.common.resources import Resources
+from forte.common.configuration import Config
+from forte.processors.data_augment.algorithms.dictionary_replacement_augmenter import DictionaryReplacementAugmenter
```
hunterhector, line 20: Now you are importing a specific Augmenter into the parent Augmenter, this will probably cause cyclic import. We shouldn't have concrete implementations here.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    def default_configs(cls):
+        """
+        :return: A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - augment_algorithm: defines the augmenter to use
+            - augment_ontologies: defines the ontologies that will be returned
+            - replacement_prob: defines the probability of replacement
+            - replacement_prob: defines the type of replacement(char/word/sentence),
```
hunterhector, line 70:
```suggestion
            - replacement_level: defines the type of replacement(char/word/sentence),
```

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+            must align with(is included in) the replacement levels of the augmenter.
```
hunterhector, line 71: OK, now this is difficult to understand, what does "align with" mean?

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+            - kwargs: augmenter-specific parameters
+        """
+        config = super().default_configs()
+        config.update({
+            'augment_algorithm': "DictionaryReplacement",
+            'augment_ontologies': ["Sentence", "Document"],
+            'replacement_prob': 0.1,
```
hunterhector, line 78: The `replacement_prop` is too concrete and should not appear in the base class.

forte/processors/base/data_augment_processor.py
```diff
@@ -0,0 +1,83 @@
...
+            'replacement_level': 'word',
+            'type': "",
+            'kwargs': {}
+        })
+        return config
```
hunterhector, line 83: New line.

forte/processors/data_augment/algorithms/base_augmenter.py
```diff
@@ -0,0 +1,42 @@
...
+    For example, the replacement_level of synonym replacement is ["word"],
+    that of back-translation is ["word", "sentence"].
+    """
+    @property
+    @abstractmethod
+    def replacement_level(self) -> List[str]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def augment(self, input: str, additional_info: Dict[str, str]) -> str:
```
hunterhector, line 41: I don't think this design is sufficient for all data augmentation algorithms.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,129 @@
...
+        then insert the new datapacks to the multipack.
+        """
+        aug_input_pack: DataPack = self._process_pack(multipack.get_pack(self.configs.input_pack_name))
+        aug_output_pack: DataPack = self._process_pack(multipack.get_pack(self.configs.output_pack_name))
+        multipack.update_pack({
+            self.configs.aug_input_pack_name: aug_input_pack,
+            self.configs.aug_output_pack_name: aug_output_pack
+        })
+
+    def _process_pack(self, pack: DataPack):
```
hunterhector, line 67: return type missing

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,129 @@
...
+        r"""
+        This function process a single datapack with an augmenter.
+        It processes one sentence at a time.
+        :param pack: a datapack with original texts and annotations
```
hunterhector, line 71: Please don't mix the documentation style, here you are using `reStructured` style (:param) but we are using Google style all along.

hunterhector (commented):
Good work on try to set up the replacement based augmentation. I think there is some work to be done on the interface design and class inheritance. Let's iterate over this a bit.

Please see my other comments inline. Importantly, we have to be extra clear when writing documentation, don't assume the users have any prior knowledge, it is better to be verbose than not clear. Always use full sentences, capitalize on every sentence, and use periods.

I haven't fully reviewed the implementation part, let's first make sure the overall design is OK first.

forte/processors/base/data_augment_processor.py
haoyuLucas: Where shall we write the docs?

forte/processors/base/data_augment_processor.py
haoyuLucas: Something like the annotations(token, sentence, document) or links. I am not sure how do we call them.

forte/processors/base/data_augment_processor.py
haoyuLucas: The parent is a processor, while the child is an augmenter. They are two types of classes.

forte/processors/base/data_augment_processor.py
haoyuLucas: The augmenter's algorithm has limitations on the replacement. Some are word-only(synonym replacement) while some can replace both word and sentence (back-translation). The user has to specify a replacement level for the processor, and it should be included in the augmenter's allowed replacement level. For example, we cannot replace a sentence with a dictionary-based augmenter(in each single augment operation).

forte/processors/base/data_augment_processor.py
haoyuLucas: Shall we add the algorithms during implementation or add them all at the very beginning?

forte/processors/base/data_augment_processor.py
hunterhector: Yeah, but we are still adding all the specific augmented instance to a base class. That means we have to change the code of the base class everytime we have a new algorithm, which does not sound right.

forte/processors/base/data_augment_processor.py
hunterhector: A couple of places to consider, one is the class's docstring, one is the configuration function. If you write them in the configuration functions, then it is better to refer to it from the class's docstring.

forte/processors/base/data_augment_processor.py
hunterhector: They are called "Entry"

forte/processors/base/data_augment_processor.py
hunterhector: Class level docstring and the default config.

forte/processors/base/data_augment_processor.py
```diff
@@ -60,8 +60,11 @@ def replace(self, replacement_op: TextReplacementOp, input: Entry):
         getting the augmented text, it will register the input & output
         for later batch process of building the new data pack.
         """
-        replaced_text: str = replacement_op.replace(input)
-        self.replaced_spans.append((input, replaced_text))
+        replaced_text: str
+        is_replace: bool
+        is_replace, replaced_text = replacement_op.replace(input)
+        if is_replace:
+            self.replaced_spans.append((input, replaced_text))
```
hunterhector, line 67: I am wondering is it sufficient to record the length of the input and replaced_text?

hunterhector (commented):
I guess the auto_align is not in the PR. Probably we should merge that PR first?

forte/processors/base/data_augment_processor.py
haoyuLucas: For the input entry, we need the begin and end span. For the replaced_text, we need the text string to build the new datapack. BTW, the data_augment_processor.py here is just a placeholder to pass the Travis test. The real file is in the auto_align_replace PR.

tests/forte/processors/data_augment/algorithms/dictionary_replacement_op_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+import unittest
+from forte.processors.data_augment.algorithms.dictionary_replacement_op \
+    import DictionaryReplacementOp
+
+from ft.onto.base_ontology import Token
+from forte.data.data_pack import DataPack
+
+
+class TestDictionaryReplacementOp(unittest.TestCase):
```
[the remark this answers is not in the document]
haoyuLucas: This PR only tests the replacement op. Operations related to the entry and datapack are processed in another [PR](https://github.com/asyml/forte/pull/285)

forte/processors/data_augment/algorithms/dictionary.py
```diff
@@ -0,0 +1,224 @@
...
+    Given an input word and its pos_tag(optional), the dictionary
+    will outputs its synonyms, antonyms, hypernyns and hyponyms.
+    """
+    def get_synonyms(
+            self,
+            word: str,
+            pos_tag: str = "",
+            lang: str = "eng"
+    ) -> List[str]:
+        # pylint: disable=unused-argument
```
hunterhector, line 38: If you put this pylint comment at a higher scope, it will be applied to the whole scope.

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
```diff
@@ -0,0 +1,64 @@
...
+        - lang: The language of the text.
+    """
+    def __init__(self, configs: Config):
+        super().__init__(configs)
+        self.dictionary = create_class_with_kwargs(
+            configs["dictionary"],
+            class_args={}
+        )
+
+    def replace(self, token: Entry) -> Tuple[bool, str]:
```
hunterhector, line 46: Use `Token` instead of `Entry`

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
```diff
@@ -0,0 +1,64 @@
...
+        Returns:
+            - A boolean value, True if the replacement happens.
+            - A synonym of the word.
+        """
+        # If the replacement does not happen, return False.
+        if random.random() > self.configs["prob"]:
+            return False, token.text
+        word = token.text
+        pos_tag = token.pos
+        lang = self.configs["lang"]
```
hunterhector, line 60: Did we specify that users need to provide this `lang` in the config?

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
```diff
@@ -0,0 +1,64 @@
...
+
+    The config should contain the following fields:
+        - dictionary: The full path to the dictionary class.
+        - prob: The probability of replacement, should fall in [0, 1].
+        - lang: The language of the text.
+    """
+    def __init__(self, configs: Config):
+        super().__init__(configs)
+        self.dictionary = create_class_with_kwargs(
+            configs["dictionary"],
```
hunterhector, line 42: So here you mean to provide the full qualified class name of the dictionary? Can you change the docstring and the config name to make sure this is clear?

1. This should not be named "path", but "full qualified name".
2. The config keyword is named `dictinary` now, maybe we should rename it to `dictionary_class`?

hunterhector (changes requested):

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
haoyuLucas: Yes. It is in the doc of DictionaryReplacementOp.

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
haoyuLucas: It seems the mypy does not allow that because the replace function in the base class takes Entry as input. Btw, we don't necessarily need a Token, a Sentence will also be good.

hunterhector (commented):
Just one small comment.

Could we review together how the generated doc looks like in the meeting?

forte/processors/data_augment/algorithms/dictionary_replacement_op.py
```diff
@@ -0,0 +1,70 @@
...
+    This class is a replacement op utilizing the dictionaries,
+    such as WORDNET, to replace the input word with an synonym.
+    Part-of-Speech(optional) can be provided to the wordnet for
+    retrieving synonyms with the same POS. It will sample from a
+    Bernoulli distribution to decide whether to replace the input,
+    with `prob` as the probability of replacement.
+
+    The config should contain the following fields:
+
+    Args:
```
hunterhector, line 38: I think you can remove this `Args` line.

hunterhector (approved):
I think this should be OK

hunterhector (approved):

Issue #259, opened 2020-09-24
Create a data augmentation processor for ranking data


cz9779:
**Is your feature request related to a problem? Please describe.**
Create a data augmentation processor for ranking data, where each multipack consists of a query and multiple documents.

**Describe the solution you'd like**
The RankingAugmentProcessor inherits the base data augment processor. It augments the query text and the documents' texts. It replaces the char/word/sentence and adds the new texts into the datapack. The augmented datapacks are added to the original multipack.

PR #260, opened 2020-09-24
create ranking data augmentation processor

cz9779:
This PR fixes [#259 ]. 

### Description of changes
Create a data augmentation processor for ranking data, where each multipack consists of a query and multiple documents. The RankingAugmentProcessor inherits the base data augment processor. It augments the query text and the documents' texts. It replaces the char/word/sentence and adds the new texts into the datapack. The augmented datapacks are added to the original multipack.

This PR builds on top of the base data augmenter PR #245 and the dictionary word-replacement data augmentation algorithm PR #242. There are duplicate files and codes of those 2 PRs, so any code reviews related to those files should go directly to the corresponding PR. Any changes in those PRs may not be up-to-date in this PR. 
Currently, query ranking augment processor is the one to review.

### Test Conducted
Create a test for RankingAugmentProcessor with bert_query_creator in the pipeline

### DIFF
--- a/forte/processors/base/__init__.py
+++ b/forte/processors/base/__init__.py
@@ -18,3 +18,4 @@ from forte.processors.base.index_processor import *
 from forte.processors.base.pack_processor import *
 from forte.processors.base.query_processor import *
 from forte.processors.base.writers import *
+from forte.processors.base.data_augment_processor import *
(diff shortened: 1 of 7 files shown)


hunterhector (changes requested):
Overall, for each particular direction/type of augmenter, we should create a base class as well. For this one, you should create a base class for ranking augmenter, and allow new implementations to be easily plugged in.

PR #261, opened 2020-09-24
Embedding similarity replacement op

### LINKED ISSUE #248: Create a word replacement from embedding similarity
This feature will create an augmenter by extending the base augmenter class.

This augmenter will replace words from embedding similarity. Will need to use the embeddings from Texar.

jrxk:
This PR fixes #248, #273 . 

### Description of changes
Create an embedding similarity data augmentation replacement op. It utilizes Texar to read a pre-trained word embedding file and finds replacement words based on embedding similarity.


forte/processors/data_augment/algorithms/dictionary_replacement_augmenter.py
```diff
@@ -0,0 +1,95 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import random
+from typing import Dict, List
+from forte.processors.data_augment.algorithms.base_augmenter import BaseDataAugmenter
```
hunterhector, line 17: We follow PEP8 that group the imports into 3 groups, and add new line between the groups. Check here: https://www.python.org/dev/peps/pep-0008/#imports

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
```diff
@@ -0,0 +1,104 @@
...
+import numpy as np
+from tensorflow import gfile
+from texar.tf.data import Embedding, load_glove
+from forte.processors.data_augment.algorithms.base_augmenter import BaseDataAugmenter
+
+__all__ = [
+    "EmbeddingSimilarityAugmenter",
+]
+
+random.seed(0)
```
hunterhector, line 26: Please put this seed setup into the class/function scope, and you may allow users to provide seeds.

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
```diff
@@ -0,0 +1,104 @@
...
+
+emb_types = ['glove']
+
+
+def load_glove_vocab(filename):
```
hunterhector, line 31: It is slightly ad-hoc to implement embeddng vocab here, let's do it formally.

You can use the Texar embedder class to wrap these pre-trained models: https://github.com/asyml/texar-pytorch/blob/master/texar/torch/modules/embedders/embedders.py

And you can follow the examples here to allow automatically download the Glove vector if possible: https://github.com/asyml/texar-pytorch/blob/master/texar/torch/modules/pretrained/pretrained_base.py#L39

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
```diff
@@ -0,0 +1,104 @@
...
+    word with another word with similar word embedding.
+    By default, the replacement word is randomly chosen from the
+    top k words with the most similar embeddings.
+
+    Args:
+        emb_path: Path to the pretrained embedding file
+        emb_type: Type of embedding. E.g. glove
+        emb_dim: Dimension of the embedding
+    """
+    def __init__(self, emb_path, emb_type, emb_dim, configs: Dict[str, str]):
```
hunterhector, line 62: 1. We should be consistent in terms of the configuration, you can put everything in the configs dict, or you can put all configs as named parameters. The latter method should be better because you can pass a dict as named parameters.
2. We will implement the embedded using Texar, hence we should pass in the embedder here directly.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,132 @@
...
+    the original multipack as augmented source and target.
+    """
+    def initialize(self, resources: Resources, configs: Config):
+        super().initialize(resources, configs)
+        self.tokenizer = TreebankWordTokenizer()
+        self.detokenizer = TreebankWordDetokenizer()
+        try:
+            nltk.pos_tag("apple")
+        except:
+            nltk.download('averaged_perceptron_tagger')
```
hunterhector, line 53: I think it is better to explicitly add nltk requirement in docstring.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,132 @@
...
+    r"""
+    The data augmentation processor for text generation data.
+    The multipack contains two datapacks: source and target.
+    This class augment both datapacks and add the new datapacks to
+    the original multipack as augmented source and target.
+    """
+    def initialize(self, resources: Resources, configs: Config):
+        super().initialize(resources, configs)
+        self.tokenizer = TreebankWordTokenizer()
+        self.detokenizer = TreebankWordDetokenizer()
```
hunterhector, line 49: So here you are requiring the users to use NLTK tokenizer, this is not what we want. We expect the system would use whatever tokenization created from the previous processors, not to force the users to use NLTK here.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,132 @@
...
+        try:
+            nltk.pos_tag("apple")
+        except:
+            nltk.download('averaged_perceptron_tagger')
+
+    def _process(self, multipack: MultiPack):
```
hunterhector, line 55: This function looks like a boilerplate, it should be in the parent class.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,132 @@
...
+from ft.onto.base_ontology import (
+    Token, Sentence, Document
+)
+
+
+__all__ = [
+    "TextGenerationDataAugmentProcessor"
+]
+
+random.seed(0)
```
hunterhector, line 37: Put this inside class scope.

forte/processors/data_augment/text_generation_augment_processor.py
```diff
@@ -0,0 +1,132 @@
...
+    def default_configs(cls):
+        config = super().default_configs()
+        config.update({
+            'input_pack_name': 'input_src',
+            'output_pack_name': 'output_tgt',
+            'aug_input_pack_name': 'aug_input_src',
+            'aug_output_pack_name': 'aug_output_tgt',
+            'replacement_prob': 0.1,
+        })
+        return config
```
hunterhector, line 132: New line at the end of the file.

hunterhector (changes requested):
Besides the interface problem, there are a bit other issues to fix, let's work on them while the interface is being consolidated. Please look at the comments, and why is `forte/processors/data_augment/dictionary_replacement_processor.py` empty?

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
jrxk: Thanks for the comment. However, I am a bit confused about what to do here:

I didn't find a method in the `Embedder` class that can take a word embedding file to initialize the embeddings. 

However, there are methods in the `texar.tf.data.Embedding` class that can do this, which is why I chose to use that class.

Also, it seems like these two aren't connected: I didn't find a way to use an `Embedding` to initialize `Embedder`, or vice versa. Also, `Embedding` uses Numpy array, and `Embedder` uses Pytorch tensor.

I think what I could do is: initialize `Embedding` first, then use that to initialize the `Embedder`.  However, I don't see the value of this additional wrapper because we will need to use the entire embedding matrix anyway.

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
jrxk: 2. Okay, I see. In this class, we assume the Embedder/Embedding is already initialized. 

So do we separate the "reading pre-trained word embedding" part out of this class, and give the user some util functions (e.g. whose input is a Glove format text file) to do that themselves?

Since our input is text, we also need a word-to-index mapping and an index-to-word mapping, right?

hunterhector: Yes, we can directly pass in an `embedder` object, which the user can freely choose. We shouldn't implement again though,  
 for word embedding, we should use the Texar's Embedding class (https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/embedding.py#L103). In the future we may need to use contextual embeddings such as Bert, we should use the Texar Embedder class (https://github.com/asyml/texar-pytorch/tree/master/texar/torch/modules/pretrained).

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
hunterhector: I see, I just realize you already using Texar Embedding. But what I mean here is we should do it outside the class so user can pass in his own embedding class. For word embedding I agree that we should use the Texar Embedding class, Embedder can be used for contextual embeddings.

forte/processors/data_augment/algorithms/embedding_similarity_augmenter.py
hunterhector: For the word-to-index and index-to-word mapping, yes we should have something for that, that is called the Vocabulary. So unfortunately I find that the Vocab implementation in texar-pytorch is not very complete, it can only load a vocabulary file. https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/vocabulary.py

However, I don't want to create too many copies of a simple `Vocabulary` implementation. What do you think here?

jrxk: Oh okay. Maybe we could make another util function to read in Glove or Word2vec file and generate a vocabulary file. Would that be a good solution?

forte/processors/data_augment/algorithms/dictionary_replacement_augmenter.py
jrxk: This was part of Yu's commit, but I'll keep that in mind.

hunterhector:
Is this PR still active?

hunterhector:
The embedding sim data-aug can still be used by extending the data augmentation framework we have now?

hunterhector:
updates?

forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py
```diff
@@ -0,0 +1,84 @@
...
+from forte.common.configuration import Config
+from forte.processors.data_augment.algorithms.text_replacement_op \
+    import TextReplacementOp
+from forte.processors.data_augment.utils.utils import l2_norm
+
+__all__ = [
+    "EmbeddingSimilarityReplacementOp",
+]
+
+random.seed(0)
```
hunterhector, line 32: This may change the random seed when the class is imported, don't add this line here.

forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py
```diff
@@ -0,0 +1,84 @@
...
+
+emb_types = ['glove']
```
hunterhector, line 34: This line seems not used.

forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py
```diff
@@ -0,0 +1,84 @@
...
+    top k words with the most similar embeddings.
+
+    Args:
+        embedding: A texar.tf.data.Embedding object. Can be initialized
+            from pre-trained embedding file using helper functions
+            E.g. forte.processors.data_augment.utils.utils.load_glove_embedding
+        vocab: A texar.torch.data.Vocab object. Can be initialized from
+            pre-trained embedding file using helper functions
+            E.g. forte.processor.data_augment.utils.utils.load_glove_vocab
+        configs: The config should contain `top_k`,
```
hunterhector, line 52: Let's use the standard forte way of adding `default_configs`. Don't add it here.

forte/processors/data_augment/utils/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        word_vecs: 2d numpy array of word vectors.
+
+    Returns:
+        Normalized vectors.
+    """
+    norm = np.sqrt((word_vecs * word_vecs).sum(axis=1))
+    return word_vecs / norm[:, np.newaxis]
+
+
+def load_glove_vocab(glove_filename: str) -> Vocab:
```
hunterhector, line 44: I think the original `Vocab` function has this method: https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/vocabulary.py#L105. Let's not re-do it

forte/processors/data_augment/utils/utils.py
```diff
@@ -0,0 +1,86 @@
...
+from texar.torch.data import Embedding, load_glove
+from texar.torch.data import Vocab
+
+__all__ = [
+    "l2_norm",
+    "load_glove_vocab"
+]
+
+
+def l2_norm(word_vecs):
```
hunterhector, line 31: Maybe `numpy` also have the norm implemented: https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/vocabulary.py#L105

forte/processors/data_augment/utils/utils.py
```diff
@@ -0,0 +1,86 @@
...
+                    continue
+                word = vec[0]
+                vocab_list.append(word)
+
+        with open(vocab_filename, "w") as output_file:
+            output_file.write('\n'.join(vocab_list))
+    return Vocab(vocab_filename)
+
+
+def load_glove_embedding(glove_filename: str, emb_dim: int, vocab: Vocab)\
```
hunterhector, line 69: https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/embedding.py#L72

forte/processors/data_augment/utils/utils.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: Most functions here do not need to be re-implmented, please try to use the wheels, not reinventing them.

tests/forte/processors/data_augment/algorithms/embedding_similarity_replacement_op_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.embedding = load_glove_embedding(abs_vocab_path, 50, self.vocab)
+        self.esa = EmbeddingSimilarityReplacementOp(
+            self.embedding,
+            self.vocab,
+            configs={
+                "top_k": 5
+            }
+        )
+
+    def test_replace(self):
```
hunterhector, line 46: Can we add another test where we are doing the correct pipeline style and see if that works?

forte/processors/data_augment/utils/utils.py
jrxk: Yes. In fact, this function is just a helper that wraps the code in the link to initialize an `Embedding`.

Do you suggest to remove this helper function and let the user initialize the `Embedding` themselves?

hunterhector: I don't see the need to have such a function. It cannot support a lot of embedding types anyway. The function here is also not really re-used, so just use it in the example

jrxk: Sure. Then I can get rid of `utils.py` entirely. Thanks!

forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py
jrxk: I'm a bit confused. The `replacement_ops` cannot be directly initialized by `initialize` since it is a parameter to the processors. I looked at other `replacement_ops` and they all use a `configs` parameter instead of `default_configs`.

hunterhector: I see. Then we simply follow the other's convention now. I may have forgotten some details about this design.

One thing about the config is whether we provide a default_config (which restricts what config parameters to be provided). Here we may also add a default config (manually)?

jrxk: We pass parameters to `replacement_ops` using the `configs`. In an actual pipeline, this op would be initialized by passing the `data_aug_op_config` to the processor's config (see the pipeline test)

tests/forte/processors/data_augment/algorithms/embedding_similarity_replacement_op_test.py
jrxk: It's done

forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py
```diff
@@ -0,0 +1,80 @@
...
+    Args:
+        configs:
+            The config should contain the following key-value pairs:
+            embedding: A texar.tf.data.Embedding object. Can be initialized
+                from pre-trained embedding file
+            vocab: A texar.torch.data.Vocab object. Can be initialized from
+                a vocabulary file
+            top_k: the number of k most similar words to choose from
+    """
+    def __init__(self, configs: Config):
```
hunterhector, line 47: I think about this parameter set-up a little bit more. Now, this stores whole python objects into a Config. I think this in the long term will hurt the system. Configs are normally not used this way. I think even sending the objects directly is better than this method. Maybe we should simply do the following?

```suggestion
    def __init__(self, embedding: Embedding, vocab: Vocab):
```

jrxk: I agree. In fact the code was changed from the latter format.

But currently, the `ReplacementDataAugmentProcessor` requires a `data_aug_op_config` for it to work. It needs to initialize the data_aug_op from that Config object.

Maybe we can instead pass the parameters to initialize the `Embedding` and `Vocab` objects, then initialize those objects in this `__init__` function?

hunterhector: We can still pass the required `data_aug_op_config` to fit the class requirement, but use handle our own Embeddings here?

jrxk: Okay. I updated the Config parameters

PR #275, opened 2020-10-05
Serialize

hunterhector:
This PR adds methods to read data from the SQLite database (created from Stave).

Description of changes
Add a utility to dump data onto the disk from the database.

Currently, it is not easy to directly read from the database since the ids of the documents are not stored in the table.

Possible influences of this PR.

Test Conducted
Added a test case for this function.


hunterhector:
Will update this PR to directly ingest data from SQLite.

Issue #247, opened 2020-09-14
Create a word replacement augmenter from distributions


cz9779:
**Is your feature request related to a problem? Please describe.**
This feature will create an augmenter by extending the base augmenter class.

**Describe the solution you'd like**
This augmenter will replace words from distributions (eg. unigram).

PR #278, opened 2020-10-08
create DA algorithm - replacement by distribution

cz9779:
This PR fixes [#247 ]. 

### Description of changes
Create a distribution replacement op `DistributionReplacementOp` that inherits `TextReplacementOp` for data augmentation.
A base `Sampler` class is defined, uniform distribution sampler and unigram distribution sampler are supported.

### Test Conducted
Create tests for `Sampler` and `DistributionReplacementOp`.


forte/processors/data_augment/algorithms/distribution_replacement_augmenter.py
```diff
@@ -0,0 +1,112 @@
...
+    Todo: More distriubtion types can be added.
+    """
+    def __init__(self, configs: Dict[str, str]):
+        random.seed(0)
+        super().__init__(configs)
+        try:
+            self.vocab = set(words.words())
+            self.stop_words = set(stopwords.words(self.configs['lang']))
+        except LookupError:
+            nltk.download('words')
```
hunterhector, line 46: Let's try to move the distribution part out, i.e., we should allow users to easily change the distribution, instead of relying on NLTK.

forte/processors/data_augment/algorithms/distribution_replacement_augmenter.py
```diff
@@ -0,0 +1,112 @@
...
+            nltk.download('stopwords')
+            self.vocab = set(words.words('en'))
+            self.stop_words = set(stopwords.words(self.configs['lang']))
+
+    @property
+    def replacement_level(self) -> List[str]:
+        return ["word"]
+
+    def _uniform_sample(self) -> str:
```
hunterhector, line 55: And also make these sample methods into individual classes so that we can pass them in.

forte/processors/data_augment/algorithms/distribution_replacement_augmenter.py
```diff
@@ -0,0 +1,112 @@
...
+        Following are the keys for this dictionary:
+            - distribution: defines what types of distribution to be used.
+                Options are uniform, unigram, etc.
+            - unigram: An input file with unigram stats of the specific dataset.
+                if distribution type is set to "unigram" and "unigram" is not provided,
+                then simply use _generate_unigram function to generate one
+                from the given text.
+        """
+        config = super().default_configs()
+        config.update({
```
hunterhector, line 108: So the better design is to pass in a "Sampler", and this class don't need to understand the Sampler type, it only knows that it can draw samples from the "Sampler".

hunterhector (changes requested):
We should include this method in some use cases and test out how it works. We should use some real distributions for this purpose.

hunterhector:
Is this PR still active? If yes, can you fix the comments and re-request review?

forte/processors/data_augment/algorithms/sampler.py
```diff
@@ -0,0 +1,108 @@
...
+    r"""
+    A base sampler, an abstract class.
+    """
+    def __init__(self):
+        random.seed()
+
+    @abstractmethod
+    def sample(self) -> str:
+        raise NotImplementedError
+
```
hunterhector, line 39: The sampler design seems restricted.

For a uniform sampler, it should probably simply take a word list, instead of hardcoding with 'nltk'. And after all, 'nltk' is not a distribution.

Similarly, the `UniformSampler` can also simply take a `Dict`.  We don't have to assume a file containing these counts.

forte/processors/data_augment/algorithms/sampler.py
```diff
@@ -0,0 +1,108 @@
...
+        r"""
+        A different distribution type can be given in the configs.
+        The default distribution is NLTK.
+        """
+        super().__init__()
+        self.configs = configs if configs is not None \
+            else self.default_config()
+
+        # set distribution
+        if self.configs["distribution"] == "nltk":
```
hunterhector, line 56: `nltk` is not a distribution.

forte/processors/data_augment/algorithms/distribution_replacement_augmenter.py
```diff
@@ -0,0 +1,48 @@
...
+    # pylint: disable=unused-argument
+    def replace(self, input: Entry) -> Tuple[bool, str]:
+        r"""
+        This function replaces a word from sampling a distribution.
+        Args:
+        Returns:
+            a word sampled from distribution
+        """
+        word: str = self.sampler.sample()
+        return True, word
```
hunterhector, line 48: Does this means that "every word" in the file will be changed? I don't think this is a good choice. You should set another probability to choose to replace only a small portion of the words.

hunterhector (changes requested):

docs/conf.py
```diff
@@ -40,7 +40,7 @@
     'sphinx.ext.extlinks',
     'sphinx.ext.napoleon',
     'recommonmark',
-    'sphinxcontrib.spelling',
+    #'sphinxcontrib.spelling',
```
hunterhector, line 43: why is this commented?

hunterhector (commented):
Only a few small comments. But two things:

1. Remember to fix the branch conflicts.
2. Does this PR depend on another? Should we merge that one first?

docs/conf.py
cz9779: oops this is a mistake, will revert back to the original

hunterhector (approved):
LGTM. You can merge after travis passing.

PR #279, opened 2020-10-08
examples: Add BC5CDR NER data reading example.

LucasWangZH:
examples: Add BC5CDR NER data reading example.

### Description of changes
Add an example as NER reader. The dataset is BC5CDR data.

### Test Conducted
Run on real dataset and get normal results.

### DIFF
--- a/examples/bc5cdr/config_data.yml
+++ b/examples/bc5cdr/config_data.yml
@@ -1,0 +1,1 @@
+train_path: "data/train/"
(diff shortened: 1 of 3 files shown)


hunterhector (changes requested):

hunterhector:
Can we finish this PR? We can probably re-use the ontonotes reder https://github.com/asyml/forte/blob/master/forte/data/readers/ontonotes_reader.py for this.

hunterhector:
If this is simply reading CONLL files, then we don't need another reader.

Issue #251, opened 2020-09-21
Write a demo converter for NER and its unit test


Jiaqiang-Ruan:
**Describe the solution you'd like**
Understand functions under VocabularyProcessor, Trainer, Predictor, Evaluator. 
Think about how to separate user code and library code in those functions.

PR #280, opened 2020-10-08
add conll03_converter and its unit test

Jiaqiang-Ruan:
This PR fixes [#251]. 

### Description of changes
This pull add a demo converter for NER example and used for a better designed converter module.

Related to the converter design for the framework

### Test Conducted
A unit test is implemented.

### DIFF
--- a/forte/data/converter/conll03_converter.py
+++ b/forte/data/converter/conll03_converter.py
@@ -1,0 +1,77 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The convert that handle conll03 data.
+"""
+import codecs
+import logging
+import os
+from typing import Iterator, Any
+
+from forte.data.data_pack import DataPack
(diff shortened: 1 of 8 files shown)


forte/data/converter/conll03_converter.py
```diff
@@ -0,0 +1,77 @@
...
+            for p in self.positions:
+                self.labelpos2id[(l, p)] = curid
+                curid += 1
+
+        # "O" label
+        for p in self.positions:
+            self.labelpos2id[('O', p)] = curid
+
+    def convert(self, pack: DataPack):
+        for sentence in pack.get(Sentence):
```
jasonyanwenl, line 59: We probably could use `pack.get_data(data_request)` to easily query different entries instead of doing this explicitly here.

forte/data/converter/conll03_converter.py
```diff
@@ -0,0 +1,77 @@
...
+            else:
+                prev_label = None
+                for cur_label in ners:
+                    if cur_label == prev_label:
+                        ners_ids.append(self.labelpos2id[(cur_label, 'I')])
+                        prev_label = cur_label
+                    else:
+                        ners_ids.append(self.labelpos2id[(cur_label, 'B')])
+                        prev_label = cur_label
+            yield token_ids, ners_ids
```
jasonyanwenl, line 77: Currently, in my design, the output from converter should be a list of tensors. One entry inside this list corresponding to one instance. Please take a look at this doc for more details: https://hackmd.io/@wcqb9QuTQjiwSLy8TexlLg/SkStvAlLP/edit

jasonyanwenl (changes requested):
Generally, I think the logic is good. We probably need to sync this implementation with the design. You could find design doc here: https://hackmd.io/@wcqb9QuTQjiwSLy8TexlLg/SkStvAlLP/edit. 
Also, the design is not finalized annd we will discuss this in the meeting.
After we finalized our design, we could change this implementation based on that.

Jiaqiang-Ruan:
Design changes.

Issue #288, opened 2020-10-15
Create a data selector for data augmentation


cz9779:
**Is your feature request related to a problem? Please describe.**
Create a data selector that selects a subset of datapacks from the original dataset, then passes to the data augmenter.

**Describe the solution you'd like**
Design an elastic search based data selector class.
Add select options on top of it.

PR #289, opened 2020-10-15
create a data selector class for data augmentation

cz9779:
This PR fixes [#288 ]. 

### Description of changes
Create a base data selector class `BaseElasticSearchDataSelector` for data augmentation. 
This class inherits `PackReader`, which is a reader that searches for relevant documents from an elastic search indexer.
Currently, there are two subclasses of data selector: `QueryDataSelector` and `RandomDataSelector`, which support search by query and random search. Note that the query selector requires an input of a query file.

The elastic search indexer needs to be created before using the data selectors. 
A `CreateIndexerBasePipeline` pipeline is composed of a specific dataset `Reader` and an `ElasticSearchIndexProcessor`, which will create an indexer with the documents of that particular dataset.
Currently, there is one subclass of the base pipeline: `CreateIndexerMSMacroPipeline`.


forte/processors/base/index_processor2.py
```diff
@@ -0,0 +1,76 @@
...
+from forte.data.data_pack import DataPack
+from forte.processors.base.pack_processor import PackProcessor
+
+
+__all__ = [
+    "IndexProcessor2"
+]
+
+
+class IndexProcessor2(PackProcessor, ABC):
```
hunterhector, line 31: First, just put this into `index_processor.py`. And second, `IndexProcessor2` is not a good name, please rename it to be more meaningful.

forte/processors/base/index_processor2.py
```diff
@@ -0,0 +1,76 @@
...
+        return config
+
+    def _bulk_process(self):
+        r"""Subclasses of :class:`IndexProcessor` should implement this method
+        to bulk add the documents into the index.
+        """
+        raise NotImplementedError
+
+    def _process(self, input_pack: DataPack):
+        compressed_datapack: str = input_pack.serialize()
```
hunterhector, line 62: What is compressed?

forte/processors/data_augment/data_select_index_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+# pylint: disable=useless-super-delegation
+
+"""
+This class is used for data selection of the data augmentation tasks.
+For each specific dataset, we first create a reader that reads
+in the document text as Datapack.
+Then we pipeline the reader and the ElasticSearchIndexProcessor
+to create a indexer,
+which is defined in forte/data/indexers/EleasticSearchIndexer.
+This class is then used by the DataSelector class to search for documents.
```
hunterhector, line 23: You mean `this index`?

forte/processors/data_augment/data_select_index_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+        self.create_pipeline()
+        self.create_index()
+
+    def _set_reader(self) -> PackReader:
+        """
+        Return: A dataset Reader.
+        """
+        raise NotImplementedError
+
+    def _set_data_path(self) -> str:
```
hunterhector, line 58: Why don't we just put this as the `default_config`? And when we use this we simply change the config?

hunterhector: Actually, I don't find data source is the best way to use this. The source of data (input parameters) should depend on the implementation of the reader. So we cannot assume that a path is enough. Maybe we can expose the parameters in the `create_index` function?

forte/processors/data_augment/data_select_index_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+    def create_pipeline(self):
+        self.nlp: Pipeline[DataPack] = Pipeline()
+        self.nlp.set_reader(self.reader)
+        self.nlp.add(DataSelectorIndexProcessor(), config=self.config)
+        self.nlp.initialize()
+
+    def create_index(self):
+        for idx, _ in enumerate(self.nlp.process_dataset(self.datapath)):
+            if idx + 1 > 0 and (idx + 1) % 10000 == 0:
+                print(f"Indexed {idx + 1} packs")
```
hunterhector, line 73: Use logging instead of print.

forte/processors/data_augment/data_select_index_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+__all__ = [
+    "CreateIndexerBasePipeline",
+    "CreateIndexerMSMacroPipeline",
+]
+
+
+class CreateIndexerBasePipeline:
+
+    def __init__(self):
```
hunterhector, line 45: Why don't we directly pass in reader in the parameter?

forte/processors/data_augment/data_select_index_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+        self.reader = self._set_reader()
+        self.datapath = self._set_data_path()
+        self.config = Config(self.default_config(), default_hparams=None)
+        self.create_pipeline()
+        self.create_index()
```
hunterhector, line 50: Here you directly create the index? Can we let the user to call this so they control when to start?

forte/processors/data_augment/data_selector_index_processor.py
```diff
@@ -0,0 +1,103 @@
...
+                "other_kwargs": {
+                    "request_timeout": 10,
+                    "refresh": False
+                }
+            }
+        })
+        return config
+
+    def _bulk_process(self):
+        documents = [{self.config.fields[0]: document[0],
```
hunterhector, line 99: May be you can use better iteration instead of hardcoding the indexes (0,1,2)

forte/processors/base/data_selector.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+__all__ = [
+    "BaseElasticSearchDataSelector",
+    "RandomDataSelector",
+    "QueryDataSelector",
+]
+
+
+class BaseElasticSearchDataSelector(PackReader):
```
hunterhector, line 42: Can we have a base class to be BaseDataSelector, where the actual implementation such as ElasticSearch is not attached here?

forte/processors/base/data_selector.py
```diff
@@ -0,0 +1,162 @@
...
+    r"""
+    The base data selector reader for data augmentation.
+    This class creates an ElasticSearchIndexer and search for document
+    according to the search key. It then generate Datapacks.
```
hunterhector, line 46: I think you need to improve the language of the docstring, making it more formal and clear.

hunterhector (changes requested):
In addition to the comments inline, please add test cases.

hunterhector:
Updates?

cz9779:
> In addition to the comments inline, please add test cases.

I'm wondering how do we add tests for data selector task because the processors are all related to the elastic search indexer that needs to be running...

hunterhector:
> > In addition to the comments inline, please add test cases.
> 
> I'm wondering how do we add tests for data selector task because the processors are all related to the elastic search indexer that needs to be running...

I think the functionality here is to make sure you can extract some expected results right? So you can create a small index, and try different configurations of your selector (such as random and query based), and make sure the results are correct.

hunterhector (approved):

PR #291, opened 2020-10-15
remove the abstract keyword for new pack

haoyuLucas:
remove the abstract keyword for new pack

### DIFF
--- a/forte/processors/base/base_processor.py
+++ b/forte/processors/base/base_processor.py
@@ -38,7 +38,6 @@ class BaseProcessor(PipelineComponent[PackType], ABC):
         super().__init__()
         self.selector = DummySelector()
 
-    @abstractmethod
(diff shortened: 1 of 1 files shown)


forte/processors/base/base_processor.py
```diff
@@ -38,7 +38,6 @@ def __init__(self):
         super().__init__()
         self.selector = DummySelector()
 
-    @abstractmethod
     def new_pack(self, pack_name: Optional[str] = None) -> PackType:
```
hunterhector, line 41: You cannot simply remove the keyword, you also need to replace "NotImplemented" with "pass" or something like that.

hunterhector (changes requested):

hunterhector:
This is solved at https://github.com/asyml/forte/pull/304

PR #294, opened 2020-10-29
Create a text classifier for the IMDB large movie dataset

### LINKED ISSUE #293: Create a text classifier for the IMDB large movie dataset
Add a modularized text classifier to Forte.

jrxk:
This PR fixes #293. 

### Description of changes
Added a text classifier for the IMDB large movie dataset based on Texar TF and BERT. The model expects CSV file inputs with columns (content, label, id), which can be generated from a Forte pipeline. 

### Test Conducted
Added an example and trained locally & got ~84% accuracy.


hunterhector (changes requested):
When we said we use Texar to implement the classifier, it means that we won't use tensorflow or pytorch directly. Texar is the layer that hides the actual implementation. And then you can easily switch between pytorch and tensorflow.

Please implement this using Texar-Pytorch instead.

jrxk:
> When we said we use Texar to implement the classifier, it means that we won't use tensorflow or pytorch directly. Texar is the layer that hides the actual implementation. And then you can easily switch between pytorch and tensorflow.
> 
> Please implement this using Texar-Pytorch instead.

Sorry, I'm a bit confused. Why do we have to use TensorFlow? Is it because we are using PyTorch for the rest of Forte?

I was referring to Texar TF's BertClassfier for this. I will modify the Texar-PyTorch version of the BertClassifier instead. Does that sound good?

hunterhector:
> > When we said we use Texar to implement the classifier, it means that we won't use tensorflow or pytorch directly. Texar is the layer that hides the actual implementation. And then you can easily switch between pytorch and tensorflow.
> > Please implement this using Texar-Pytorch instead.
> 
> Sorry, I'm a bit confused. Why do we have to use PyTorch? Is it because we are using PyTorch for the rest of Forte?
> 
> I was referring to Texar TF's BertClassfier for this. I will modify the Texar-PyTorch version of the BertClassifier instead. Does that sound good?

1. We would like to use more general interfaces. In this implementation, there are still some specific tensorflow calls.
2. And most of the other implementation examples are texar-pytorch, so let's stick with that.

jrxk:
Reimplement with texar-pytorch.

hunterhector:
Could you open this as another PR and make sure this file is not included here directly? 

examples/text_classification/data/IMDB_raw/train_id_list.txt

Files in GitHub will be there permanently once committed. So this file would increase the size of the project permanently. Can you simply provide a download link?

When you open another PR, make sure this file is not in your commit history of the branch. This means that you should create a fresh branch from master, add your changes to that branch, and do not include `train_id_list`.

Note that `git rm` on the current branch will not reduce the size of this PR.

PR #301, opened 2020-10-30
Update the cache

ziqian98:
### Description of changes
Map each datapack and instance to a npy file

Cache on the disk and do not use run time memory any more.

### Test Conducted
Tested locally


hunterhector:
@jasonyanwenl Do you think we should close this PR?

jasonyanwenl:
Yes, I agree.

PR #310, opened 2020-11-15
Add BioBERT NER Predictor

mingkaid:
This PR adds BioBERT NER Predictor and appropriate infrastructure to the repo

### Description of changes
- Added BioBERTNERPredictor class to processors/ner_predictor.py module
- Added transformers_processor.py module with BERTTokenizer class, which produces WordPiece tokenization used in BERT-based models
- Added Subword ontology to ft/onto/base_ontology.py. TODO: Fix generate_ontology script

No influence on other modules / classes expected

### Test Conducted
Tested on with BioBERT cased v1.1 model fine-tuned on NCBI-disease dataset, evaluated on a number of sentences from CORD-19 dataset.


examples/biobert_ner/ner_predict_example.py
```diff
@@ -0,0 +1,63 @@
...
+from forte.data.readers import StringReader
+from forte.pipeline import Pipeline
+from forte.processors import BERTTokenizer, BioBERTNERPredictor
+from forte.processors.nltk_processors import NLTKSentenceSegmenter
+
+from ft.onto.base_ontology import Subword, Sentence, EntityMention
+
+config = yaml.safe_load(open("config.yml", "r"))
+
+config = Config(config, default_hparams=None)
```
hunterhector, line 30: I don't see the reason for writing config at the module level. But this is fine for an example class.

forte/processors/ner_predictor.py
```diff
@@ -348,3 +354,183 @@ def default_configs(cls):
 
         configs.update(more_configs)
         return configs
+
+    
+class BioBERTNERPredictor(FixedSizeBatchProcessor):
```
hunterhector, line 359: We currently use Python typing for most variables. It would be great if you can add the typing to the variables in this class.

hunterhector (commented):
This is a pretty good PR considering this is your first time contributing to a public OSS project. Great Job! There are a few small format issues and missed typing here, but due to some time constraint, I would rather merge this and fix the typing later. 

In Forte, we use type annotation extensively to improve readability and reduce error.

hunterhector (approved):
Will pass this although the dependency for documentation is not done (reason being dependency on transformers). I will move the class out to the example to fix that later.

forte/processors/ner_predictor.py
mingkaid: Got it. I can add typing to variables where appropriate. 

When I observed the `CoNLLNERPredictor` class, I noticed that only variables that we expect to be passed outside a function gets typing. Would you describe that as the guiding principle to typing in this repo?

hunterhector: Actually, I have merged this and we don't have to worry about the typing right now, since I am considering converting this as an example.

`CoNLLNERPredictor` is a bit old so some convention is not fully enforced.

examples/biobert_ner/ner_predict_example.py
mingkaid: What is "writing config at the module level"? I was just copying what `examples/pipeline/process_string_example.py` did. What's the best practice for this?

hunterhector: The config is at the top level of this module (the current file).  In this example, it can actually be added to the `main`  function. The problem of writing at the module level is that if someone else import this module, then the `config` line will be executed at import time. This is fine since this is just an example script.

forte/processors/ner_predictor.py
mingkaid: I see. Thanks for providing the context

examples/biobert_ner/ner_predict_example.py
mingkaid: I see - That makes a lot of sense. I'll be careful about this in the next example script

PR #316, opened 2020-11-18
Support memory cached data pack inside reader

### LINKED ISSUE #315: Support caching parsed data pack in memory
When running pipeline multiple times during training phase, we need to retreive exactly same data packs once for each epoch. If we do not cache the parsed data pack and ask reader to re-parse the data pack during each epoch, it will be time-consuming.

Inside `BaseReader` class, add memory caching mechanism.

Doing the cache manually outside of the reader. But this has conflicts with the usage of pipeline which assumes reader is a pipeline component and as long as the reader is put into pipeline, we should not interact with pipeline anymore. This means it is hard to do caching around pipeline.

**Additional context**
This issue is related to the current development of model interface project issue #297.

jasonyanwenl:
This PR fixes #315.

### Description of changes
Track parsed data pack in side reader and retrieve those data packs instead of re-parsing from disk file when the `iter` method is called the second time.

Default is do not cache in memory so it should have no side-effect for current readers.

### Test Conducted
N/A

### DIFF
--- a/data_samples/base_reader_test/0.txt
+++ b/data_samples/base_reader_test/0.txt
@@ -1,0 +1,35 @@
+1. All you need to do is pick up the pen and begin.
+2. The knives were out and she was sharpening hers.
+3. So long and thanks for the fish.
+4. Rock music approaches at high velocity.
+He didn't heed the warning and it had turned out surprisingly well.
+She advised him to come back at once.
+A glittering gem is not enough.
+She could hear him in the shower singing with a joy she hoped he'd retain after she delivered the news.
+Check back tomorrow; I will see if the book has arrived.
+He had accidentally hacked into his company's server.
+All you need to do is pick up the pen and begin.
+She borrowed the book from him many years ago and hasn't yet returned it.
+There was no ice cream in the freezer, nor did they have money to go to the store.
+She always speaks to him in a loud voice.
+There is a fly in the car with us.
+It was the best sandcastle he had ever seen.
+The memory we used to share is no longer coherent.
+Italy is my favorite country; in fact, I plan to spend two weeks there next year.
+My Mum tries to be cool by saying that she likes all the same things that I do.
+How do you like RandomWordGenerator.com?
+A song can make or ruin a person’s day if they let it get to them.
+Oh, how I'd love to go!
+A glittering gem is not enough.
+The pizza smells delicious.
+I hope that, when I've built up my savings, I'll be able to travel to Mexico.
+Not all people who wander are lost.
+Yeah, I think it's a good environment for learning English.
+He didn't heed the warning and it had turned out surprisingly well.
+I would have gotten the promotion, but my attendance wasn’t good enough.
+I want to buy a onesie… but know it won’t suit me.
+It was getting dark, and we weren’t there yet.
+Did you know that, along with gorgeous architecture, it's home to the largest tamale?
+Joe made the sugar cookies; Susan decorated them.
+Brad came to dinner with us.
+So long and thanks for the fish.
--- a/data_samples/base_reader_test/1.txt
+++ b/data_samples/base_reader_test/1.txt
@@ -1,0 +1,35 @@
+1. All you need to do is pick up the pen and begin.
+2. The knives were out and she was sharpening hers.
+3. So long and thanks for the fish.
+4. Rock music approaches at high velocity.
+He didn't heed the warning and it had turned out surprisingly well.
+She advised him to come back at once.
+A glittering gem is not enough.
+She could hear him in the shower singing with a joy she hoped he'd retain after she delivered the news.
+Check back tomorrow; I will see if the book has arrived.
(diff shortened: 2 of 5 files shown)


hunterhector (commented):
I am wondering what we should do when we call the `initialize` function? Shall we clear the cache? This depends on whether you want to call `initialize` during your training loop.

jasonyanwenl:
> I am wondering what we should do when we call the `initialize` function? Shall we clear the cache? This depends on whether you want to call `initialize` during your training loop.

In my current implementation, I do not call `initialize` of reader and it does not matter whether I clear cache or not because `reader` is created only once and I will not use the functionality of clearing the memory cache. However, we probably need to support clearing memory cache in case some other use cases need this functionality. I guess here are two ways to do this:

* Like you mentioned, we can do this in `BaseReader::initialize` method. This means that if a user wants to clear the memory cache, he needs to either call `BaseReader::initialize` or `Pipeline::initialize` (which internally will call `BaseReader::initialize`) but I guess this is probably not intuitive to a user?

* Add another method in `BaseReader` called `clear_memory_cache` that allows users explicitly clear memory cache.

Which way do you think is better?

hunterhector:
> > I am wondering what we should do when we call the `initialize` function? Shall we clear the cache? This depends on whether you want to call `initialize` during your training loop.
> 
> In my current implementation, I do not call `initialize` of reader and it does not matter whether I clear cache or not because `reader` is created only once and I will not use the functionality of clearing the memory cache. However, we probably need to support clearing memory cache in case some other use cases need this functionality. I guess here are two ways to do this:
> 
> * Like you mentioned, we can do this in `BaseReader::initialize` method. This means that if a user wants to clear the memory cache, he needs to either call `BaseReader::initialize` or `Pipeline::initialize` (which internally will call `BaseReader::initialize`) but I guess this is probably not intuitive to a user?
> * Add another method in `BaseReader` called `clear_memory_cache` that allows users explicitly clear memory cache.
> 
> Which way do you think is better?

As you mentioned, `initialize` is a special function of `PipelineComponent` that will be called by the pipeline when the user calls the pipeline's `initialize` function. So it is important that you know that this behavior is OK for you.

jasonyanwenl:
> > > I am wondering what we should do when we call the `initialize` function? Shall we clear the cache? This depends on whether you want to call `initialize` during your training loop.
> > 
> > 
> > In my current implementation, I do not call `initialize` of reader and it does not matter whether I clear cache or not because `reader` is created only once and I will not use the functionality of clearing the memory cache. However, we probably need to support clearing memory cache in case some other use cases need this functionality. I guess here are two ways to do this:
> > 
> > * Like you mentioned, we can do this in `BaseReader::initialize` method. This means that if a user wants to clear the memory cache, he needs to either call `BaseReader::initialize` or `Pipeline::initialize` (which internally will call `BaseReader::initialize`) but I guess this is probably not intuitive to a user?
> > * Add another method in `BaseReader` called `clear_memory_cache` that allows users explicitly clear memory cache.
> > 
> > Which way do you think is better?
> 
> As you mentioned, `initialize` is a special function of `PipelineComponent` that will be called by the pipeline when the user calls the pipeline's `initialize` function. So it is important that you know that this behavior is OK for you.

Yes, this is OK for me. But still do we need to add clearing cache logic in `BaseReader::initialize` in this PR?

hunterhector:
> > > > I am wondering what we should do when we call the `initialize` function? Shall we clear the cache? This depends on whether you want to call `initialize` during your training loop.
> > > 
> > > 
> > > In my current implementation, I do not call `initialize` of reader and it does not matter whether I clear cache or not because `reader` is created only once and I will not use the functionality of clearing the memory cache. However, we probably need to support clearing memory cache in case some other use cases need this functionality. I guess here are two ways to do this:
> > > 
> > > * Like you mentioned, we can do this in `BaseReader::initialize` method. This means that if a user wants to clear the memory cache, he needs to either call `BaseReader::initialize` or `Pipeline::initialize` (which internally will call `BaseReader::initialize`) but I guess this is probably not intuitive to a user?
> > > * Add another method in `BaseReader` called `clear_memory_cache` that allows users explicitly clear memory cache.
> > > 
> > > Which way do you think is better?
> > 
> > 
> > As you mentioned, `initialize` is a special function of `PipelineComponent` that will be called by the pipeline when the user calls the pipeline's `initialize` function. So it is important that you know that this behavior is OK for you.
> 
> Yes, this is OK for me. But still do we need to add clearing cache logic in `BaseReader::initialize` in this PR?

There are other readers that will extend `BaseReader` so it is not only related to this implementation. Overall, I think the semantics of the `initialize` function should set all the states to the initial one. So clearing the cache now make more sense.

hunterhector (commented):
Should be OK. But we probably need a test for it.

hunterhector (dismissed):
LGTM

hunterhector (approved):

PR #329, opened 2020-12-07
Update predicate argument span start/end detection and read multiple documents in one file - Ontonotes Reader

jasonyanwenl:
### Description of changes
Original, there are two issues for [OntonotesReader](https://github.com/asyml/forte/blob/master/forte/data/readers/ontonotes_reader.py):

* The `Predicate Argument` span start/end detection is not currect for nested span cases like `(ARG1(R-ARG2*)`

* There could be multiple documents in a single dataset file. However, the reader will only parse the first document in the file and ignore all subsequent documents in this file.

Regarding the first issue, a workaround is used to ignore all inner spans.
Regarding the second issue, now each document in a file will be parsed into a single `datapack`.

N/A

### Test Conducted
Will add unit testing for those two issue cases.


hunterhector (commented):
I can pass this if you make sure the test cases are OK.

jasonyanwenl:
> I can pass this if you make sure the test cases are OK.

Sorry, please wait a moment. I haven't checked the issue thoroughly. I will ping you when I have done it.

hunterhector (approved):

Issue #328, opened 2020-12-06
A Classification Example in Forte Pipeline


ziqian98:
**Is your feature request related to a problem? Please describe.**
Provide Forte with a classification example

**Describe the solution you'd like**
1. Using the reader to parse the dataset as Sentence Datapack and Token Datapack

2. Using the Text Extractor on the input to Build the word embedding table from a Token level and using the Attribute Extractor again on the output to get the label information from the Sentence Level.

3. Build a CNN sentence classifier for the sentiment classification task on the IMDB movie review dataset.
(https://www.kaggle.com/lakshmi25npathi/imdb-dataset-of-50k-movie-reviews)

**Describe alternatives you've considered**
Attribute Extractor in the output maybe replaced by other extractors.

**Additional context**
View the idea design at:
(https://drive.google.com/file/d/17kcwu-XOVzk-8lEmVDWWASuAohPIA-PQ/view?usp=sharing)

PR #330, opened 2020-12-08
A new Classification Example using Forte Components on IMDB dataset

ziqian98:
This PR fixes [https://github.com/asyml/forte/issues/328]. 

### Description of changes
1. Write a reader for building the Sentence and Token DataPack.
2. Build the word embedding table by attribute_extractor on Token level
3. Build a CNN sentence classifier for the sentiment classification
4. Instead of using PyTorch Dataset class in reading, the new pipeline use TrainPreprocessor class, takes in the reader and extractors as inputs and return a train batch iterator. The train batch iterator contains the word index matrix and labels. The CNN model only take  word index matrix as inputs and calculates loss between its outputs and labels.

Provide the user with an example in classification task

### Test Conducted
Passed Unit Test for IMDB Reader


hunterhector:
So let's make sure the following points for a pull request:
1. It should contain only the work related to this feature, please do not commit other stuff.
2. It should not contain any binary, these binaries will increase the repository size permanently and we cannot undo it.
3. I read the "classification_new" a little, it looks like you are still doing embeddings, padding and convolution yourself. I don't think you really understand the project idea.

ziqian98:
> So let's make sure the following points for a pull request:
> 
> 1. It should contain only the work related to this feature, please do not commit other stuff.
> 2. It should not contain any binary, these binaries will increase the repository size permanently and we cannot undo it.
> 3. I read the "classification_new" a little, it looks like you are still doing embeddings, padding and convolution yourself. I don't think you really understand the project idea.

Like we discussed in the slack, I will close this PR and I will open up a new cleaner PR for review.

PR #331, opened 2020-12-09
Support prediction for NER new task.

Jiaqiang-Ruan:
The code will use the extractor framework and will need the output from train process to do the prediction.

### Test Conducted
Small dataset validation is done.


examples/ner_new/main_predict.py
```diff
@@ -0,0 +1,65 @@
...
+    print("---- pack ----")
+    for instance in pack.get(Sentence):
+        sent = instance.text
+        ner_tags = []
+        for entry in pack.get(EntityMention, instance):
+            ner_tags.append((entry.text, entry.ner_type))
+        print('---- example -----')
+        print("sentence: ", sent)
+        print("ner_tags: ", ner_tags)
+    print(evaluator.get_result())
```
hunterhector, line 65: Can we get the whole dataset result at the end?

hunterhector (commented):
I am a bit curious on how do we know the model class here? I have the same question as the SRL example.

examples/ner_new/main_predict.py
```diff
@@ -0,0 +1,65 @@
...
+from forte.pipeline import Pipeline
+from forte.data.readers.conll03_reader_new import CoNLL03Reader
+from forte.predictor import Predictor
+from ft.onto.base_ontology import Sentence, EntityMention
+from examples.ner_new.ner_evaluator import CoNLLNEREvaluator
+
+
+def predict_forward_fn(model, batch):
+    '''Use model and batch data to predict ner tag.'''
+    word = batch["text_tag"]["tensor"]
```
jasonyanwenl, line 27: It would be better if you can add typing hint for all variables here.

hunterhector (commented):
This PR should be good. Do we have the models downloadable somewhere? And do we have a README file associated?

hunterhector (approved):

PR #332, opened 2020-12-09
Support prediction for SRL task.

Jiaqiang-Ruan:
### Description of changes
This PR support the prediction for SRL task. It adds user code built upon extractor framework.

### Test Conducted
The code is runnable. But the performance still need to be validated on the small dataset.


examples/srl/main_predict.py
```diff
@@ -0,0 +1,67 @@
...
+
+predictor = Predictor(batch_size=10,
+                        model=saved_model,
+                        predict_forward_fn=predict_forward_fn,
+                        feature_resource=train_state["feature_resource"])
+
+pl = Pipeline()
+pl.set_reader(reader)
+pl.add(predictor)
+pl.initialize()
```
hunterhector, line 62: As an example, it is better to create a writer to write out the results, or show some pack content.

examples/srl/main_predict.py
```diff
@@ -0,0 +1,67 @@
...
+from torch import Tensor
+from ft.onto.base_ontology import Sentence
+from forte.data.converter.feature import Feature
+from forte.pipeline import Pipeline
+from forte.data.readers.ontonotes_reader import OntonotesReader
+from forte.predictor import Predictor
+
+
+reader = OntonotesReader()
+saved_model = torch.load("model.pt")
```
hunterhector, line 26: Is this just the weights, or the whole model class?

examples/srl/main_predict.py
```diff
@@ -0,0 +1,67 @@
...
+    for feature in raw_text_features:
+        text.append(feature.unroll()[0])
+
+    output = model.decode(text=text,
+                          char_batch=char_tensor,
+                          char_masks=char_masks,
+                          text_batch=text_tensor,
+                          text_mask=text_mask,
+                          srl_features=srl_features)
+    print(output)
```
hunterhector, line 48: you probably don't want to print this in real production

hunterhector (changes requested):
I cannot fully review this without seeing work and seeing other part of the SRL system. A few comments:
1. The code probably hide a little too much, for example, users cannot see the model class anymore
2. If this is an real example, we should provide downloadable weights and other things for people to try.

Issue #334, opened 2020-12-10
Require a better way to initialize reader by given config


jasonyanwenl:
**Is your feature request related to a problem? Please describe.**
Currently `BaseReader::__init__` method have several arguments like `from_cache`, `cache_directory`, etc. and tose parameters are optional. This will introduce some potential risks:

* Some child readers inherited from base reader will overwrite `__init__` without provide any base reader level argument (and no bug detected by modern IDE as those args are optional)
* reader class also provides a `initialize` method to do initialization. Those two entry points will introduce confusion.
* This is not flexible as if we want to suport more option, we need to add another arg in `BaseReader::__init__`.

It would be better if we entirely leave `BaseReader::__init__` as empty and move all args into `default_configs`. And then we only do initialization in `initialize` method.

However, this also means any developer who wants to write a new reader need to keep in mind that the `super().initialize()` should be called in `initialize` and `super().default_configs` should be called in `default_configs`. This is more error-prone.

**Describe the solution you'd like**
we need a cleaner and less error-prone way to do reader initialization.

**Describe alternatives you've considered**
Currently all child readers should not overwrite `BaseReader::__init__` or overwrite it but need to accept all arguments in `BaseReader::__init__`.

hunterhector:
Is this issue resolved?

jasonyanwenl:
> Is this issue resolved?

No. This issue is still existing and to solve this we need to carefully come up with a better solution that balances the trade-off mentioned above.

PR #335, opened 2020-12-10
Move column format logic to `initialize` method

jasonyanwenl:
This PR fixes the issue that `OntonotesReader` does not accept `BaseReader` level init arguments. And this is partially related to this [issue](https://github.com/asyml/forte/issues/334)

### Description of changes
Move initialization logic to `initialize` method. And move default `column format` to `default_config`.

Another PR (main train SRL which will be raised soon) will use this feature.

### Test Conducted
The change should not affect current unit testings

### DIFF
--- a/forte/data/readers/ontonotes_reader.py
+++ b/forte/data/readers/ontonotes_reader.py
@@ -128,6 +128,39 @@ class OntonotesReader(PackReader):
 
     @classmethod
     def default_configs(cls):
+        r"""
+        Returns a dictionary of default hyperparameters.
+
+        .. code-block:: python
+
+            {
+                "name": "reader",
+                "column_format": [
+                    "document_id",
+                    "part_number",
+                    None,
+                    "word",
+                    "pos_tag",
+                    None,
+                    "lemmatised_word",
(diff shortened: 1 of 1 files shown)


forte/data/readers/ontonotes_reader.py
```diff
@@ -117,6 +126,15 @@ def __init__(self, column_format: Optional[List[Optional[str]]] = None):
             if field not in seen_fields:
                 raise ValueError(f"'{field}' field is required")
 
+    @classmethod
+    def default_configs(cls):
+        config: dict = super().default_configs()
+
+        config.update({
+            "column_format": cls._DEFAULT_FORMAT
```
hunterhector, line 134: it would be helpful to add the description of the column format as the docstring for this class.

hunterhector (commented):
The fix should be OK. A small docstring modification would be nice.

forte/data/readers/ontonotes_reader.py
jasonyanwenl: Done

hunterhector (approved):
lgtm

PR #337, opened 2020-12-14
RL-DA model alone without example

### LINKED ISSUE #305: create a DA_RL trainer module
Add [Learning Data Manipulation for Augmentation and Weighting](https://arxiv.org/pdf/1910.12795.pdf) model.

Implement a data augmentation and reinforcement learning trainer.
Code reference [here](https://github.com/tanyuqian/learning-data-manipulation/)

cz9779:
This PR fixes #305 

### Description of changes
Add two models for reinforcement learning data augmentation: `MetaAugmentationWrapper` and `MetaModel`.
Add documentation.

### Test Conducted
Add unit tests for the above two models, respectively.


forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+from forte.models.da_rl.magic_model import MetaModule
+
+
+__all__ = [
+    "MetaAugmentationWrapper"
+]
+
+
+class MetaAugmentationWrapper:
+    # pylint: disable=line-too-long
```
hunterhector, line 36: probably not disable this.

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+                `segment_ids` is a tensor of shape `[seq_len]`.
+                `label_ids` is a tensor of shape `[seq_len]`.
+            num_aug: The number of samples from the augmentation model.
+
+        Returns:
+            aug_probs: A tensor of shape `[num_aug, seq_len, token_size]`.
+                It is the augmented bert token soft embedding.
+        """
+
+        init_ids, input_mask, segment_ids, _ = \
```
hunterhector, line 114: Type annotations here would be needed.

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+    "MetaAugmentationWrapper"
+]
+
+
+class MetaAugmentationWrapper:
+    # pylint: disable=line-too-long
+    r"""
+    A wrapper adding data augmentation to a Bert model with arbitrary tasks.
+    This is used to perform reinforcement learning for joint data augmentation
+    learning and model training.
```
hunterhector, line 40: Although the example is not included, some code examples of how to use this class would be very nice.

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+            [ap.unsqueeze(0) for ap in aug_probs_all], axis=0)
+
+        return aug_probs
+
+    def augment_instance(self, features: Tuple[torch.Tensor, ...]) \
+            -> Tuple[torch.Tensor, ...]:
+        r"""Augment a training instance.
+
+        Args:
+            features: A tuple of Bert features of one training instance.
```
hunterhector, line 156: What exactly is a "bert feature"?

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+
+            `input_probs_aug` is a tensor of soft Bert embeddings,
+            distributions over vocabulary.
+            It has shape `[num_aug, seq_len, token_size]`.
+            It keeps :math:`\phi` as variable so that after passing it as an
+            input to the classifier, the gradients of :math:`\theta` will
+            also apply to :math:`\phi`.
+
+            `input_mask_aug`, `segment_ids_aug`, `label_ids_aug` are all
+            tensors of shape `[num_aug, seq_len, token_size]`.
```
hunterhector, line 180: But what are they?

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: I don't think the name of this file should be `generator_trainer` though. First this class mainly contains the wrapper, secondly, `generator` and `trainer` are both very general names, we don't know what to expect from such a file.

forte/models/da_rl/generator_trainer.py
```diff
@@ -0,0 +1,398 @@
...
+                 augmentation_optimizer: Optimizer,
+                 input_mask_ids: int, device: torch.device, num_aug: int):
+        self._aug_model = augmentation_model
+        self._aug_optimizer = augmentation_optimizer
+        self._input_mask_ids = input_mask_ids
+        self._device = device
+        self._num_aug = num_aug
+
+    def reset_model(self):
+        self._aug_model.train()
```
hunterhector, line 86: Do we need an eval mode for this model? If yes, how do we use that?

forte/models/da_rl/magic_model.py
```diff
@@ -0,0 +1,160 @@
...
+import torch.nn as nn
+
+
+__all__ = [
+    "MetaModule"
+]
+
+
+class MetaModule(nn.Module):
+    # pylint: disable=line-too-long
```
hunterhector, line 32: Let's not disable this thing.

forte/models/da_rl/magic_model.py
```diff
@@ -0,0 +1,160 @@
...
+    and performs parameter updates locally.
+
+    This code is adapted from:
+    https://github.com/tanyuqian/learning-data-manipulation/blob/master/magic_module.py
+
+    It implements the calculation:
+    :math:`L(\theta - \nabla_{\theta} L_{train}(\theta, \phi))`.
+
+    Args:
+        module: A pytorch nn module.
```
hunterhector, line 43: you can specify the class name `torch.nn.Module`

forte/models/da_rl/magic_model.py
```diff
@@ -0,0 +1,160 @@
...
+
+    In order to perform :meth:`forward` the same way as the input module,
+    we need to copy into this class the helper functions that are called by the
+    nested :meth:`forward` of the input module.
+
+    For example, if the input module is tx.modules.BERTClassifier,
+    :meth:`_get_noise_shape`, :meth:`_split_heads`, :meth:`_combine_heads`
+    are needed to be exposed in this class, so that this :meth:`forward` can
```
hunterhector, line 51: Hmm, copying? So this is not a general wrapper.

setup.py
```diff
@@ -48,12 +48,12 @@
...
         'ir': ['faiss-cpu>=1.6.1', 'elasticsearch==7.5.1'],
         'spacy': ['spacy==2.3.0'],
-        'allennlp': ['allennlp==1.1.0', 'allennlp-models==1.1.0',
+        'allennlp': ['allennlp==1.2.0', 'allennlp-models==1.2.0',
                      'torch>=1.5.0'],
         'cliner': ['marisa-trie==0.7.4', 'scipy==1.5.0',
                    'scikit-learn==0.23.1', 'repoze.lru==0.7',
                    'tensorflow-gpu==1.12.0', 'python-crfsuite==0.9.7'],
-        'augment': ['elasticsearch==7.5.1']
+        'augment': ['elasticsearch==7.5.1', 'transformers>=3.1']
```
hunterhector, line 56: Just to make sure, you didn't use `transformers`, this is from another PR?

hunterhector (changes requested):
I think you need to copy the functions from `tx.modules.embedders` and `tx.modules.encoders` because you don't extend them. Since now this module is not generally applicable anyway, we don't we simply extend the same interface as tx Bert? Copying those functions is not a very good idea.

forte/models/da_rl/magic_model.py
cz9779: I disabled this because I have some URLs in the comments and they should be in one single line? I made sure that all the other lines are <80 char long.

setup.py
cz9779: I used `transformers` in my tests for `MetaAugmentationWrapper` because their `BertForMaskedLM` is the intended augmentation model we use. But yes I did copy this from Yu Hao's back translation PR.

forte/models/da_rl/generator_trainer.py
cz9779: same as the other reply I made, it's solely used because of the URL in the comment. All the other lines are <80 char long.

setup.py
hunterhector: I though we are using the Bert model from Texar.

forte/models/da_rl/generator_trainer.py
hunterhector: Understood. I think we may have methods to break URL lines somehow, but this is not very important.

setup.py
cz9779: The classifier is from Texar but the aug model is from Transformer. Texar doesn't support it.

hunterhector: Interesting to see Transformer and Texar work together well.

forte/models/da_rl/aug_wrapper.py
```diff
@@ -0,0 +1,488 @@
...
+                    aug_instance_features = \
+                        aug_wrapper.augment_example(instance_features)
+                    # Model is the downstream Bert model.
+                    model.zero_grad()
+                    loss = model(aug_instance_features)
+                    meta_model = MetaModule(model)
+                    meta_model = aug_wrapper.update_meta_classifier(
+                        meta_model, loss, model, optim)
+
+                    # Compute grads of aug model on eval data
```
hunterhector, line 95: validation data

hunterhector (commented):
The `MetaModule` looks nice now. Hope it can be applied in many places.

Just one small comment on the documentation.

hunterhector (approved):

PR #339, opened 2020-12-18
Fix `initialize` in Readers

DianLiI:
### Description of changes
Remove unnecessary `initialize` from Readers or add super calls in them when necessary.

N/A

### Test Conducted
N/A


hunterhector (approved):
Thanks for cleaning up these details. LGTM

Issue #344, opened 2020-12-19
Create an example for data augmentation with data selector 


cz9779:
**Describe the solution you'd like**
An example to use the data selector for data augmentation.

PR #345, opened 2020-12-19
add data selector example & rename

cz9779:
This PR fixes [#344 ]. 

### Description of changes
Create an example for the data selector of data augmentation task.

### Test Conducted
A README that describes the usage.


examples/data_augmentation/data_select/data_select_and_augment_example.py
```diff
@@ -0,0 +1,81 @@
...
+                "lang": "eng",
+            }
+        },
+        'augment_pack_names': {
+            'kwargs': {
+                'input': 'augmented_input'
+            }
+        }
+    }
+
```
hunterhector, line 64: How about we move the configs to another file and load them with YAML? We can keep the example focus on the main structure.

forte/processors/base/data_selector_for_da.py
```diff
@@ -31,19 +31,19 @@
 
 
 __all__ = [
-    "BaseElasticSearchDataSelector",
-    "RandomDataSelector",
-    "QueryDataSelector",
+    "BaseElasticSearchDataSelectorForDA",
```
hunterhector, line 34: I suppose you added 'ForDA' to avoid confusing with `Selector`? I think you have already named them as `DataSelector` it is actually OK now.

forte/processors/base/data_selector_for_da.py
```diff
@@ -31,19 +31,19 @@
...
+    "RandomDataSelectorForDA",
+    "QueryDataSelectorForDA",
 ]
 
 
-class BaseDataSelector(PackReader):
+class BaseDataSelectorForDA(PackReader):
```
hunterhector, line 40: I don't think `DataSelector have to "searches from an indexer", it just return data from a document set. So you can change the docstring to make it more general

forte/processors/base/data_selector_for_da.py
cz9779: Do you mean the previous names are ok and we can change them back?

hunterhector: Yeah. I think the base `Selector` name will be confusing. These concrete classes look ok.

hunterhector (approved):
LGTM

PR #349, opened 2020-12-24
add data selector & rl-model tutorial

cz9779:
This PR adds data selector and rl-model tutorials.


hunterhector (approved):
I guess there's definitely some room to improve. But anyway, happy chrismas.

PR #350, opened 2020-12-24
Add attribute extractor

Jiaqiang-Ruan:
### Description of changes
This PR add AttributeExtractor which inherit from BaseExtractor.

NLP tasks like NER or POS would use this class in our framework.

### Test Conducted
Test on NER dataset using this extractor is done.

### DIFF
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -55,14 +55,15 @@ class AttributeExtractor(BaseExtractor):
         return state
 
     @classmethod
-    def from_state(cls, state: Dict) -> "AttributeExtractor":
-        obj = super().from_state(state)
+    def from_state(cls, state: Dict) \
+        -> Union["BaseExtractor", "AttributeExtractor"]:
+        obj = BaseExtractor.from_state(state)
         additional = {
             "attribute": state["attribute"]
         }
         obj.config = Config(obj.config, additional,
                             allow_new_hparam=True)
-        obj = cls(obj)
+        obj.__class__ = cls
         return obj
 
     @staticmethod
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -46,29 +46,9 @@ class AttributeExtractor(BaseExtractor):
             raise AttributeError("attribute needs to be specified in "
                                 "the configuration of an AttributeExtractor.")
 
-    @property
-    def state(self) -> Dict:
-        state = super().state
-        state.update({
-            "attribute": self.config.attribute
-        })
-        return state
-
-    @classmethod
-    def from_state(cls, state: Dict) \
-        -> Union["BaseExtractor", "AttributeExtractor"]:
-        obj = BaseExtractor.from_state(state)
-        additional = {
-            "attribute": state["attribute"]
-        }
-        obj.config = Config(obj.config, additional,
-                            allow_new_hparam=True)
-        obj.__class__ = cls
-        return obj
-
     @staticmethod
     def get_attribute(entry: Entry, attr: str) -> Any:
-        r"""Functionality: Get the attribute from entry. You can
+        r"""Get the attribute from entry. You can
         overwrite this function if you have sepcial way to get the
         attribute from entry.
 
@@ -84,7 +64,7 @@ class AttributeExtractor(BaseExtractor):
 
     @staticmethod
     def set_attribute(entry: Entry, attr: str, value: Any):
-        r"""Functionality: Set the attribute of an entry to value.
+        r"""Set the attribute of an entry to value.
         You can overwrite this function if you have sepcial way to
         set the attribute.
 
@@ -100,7 +80,7 @@ class AttributeExtractor(BaseExtractor):
         setattr(entry, attr, value)
 
     def update_vocab(self, pack: DataPack, instance: Annotation):
-        r"""Functionality: Get all attributes of one instance and
+        r"""Get all attributes of one instance and
         add them into the vocabulary.
 
         Args:
@@ -120,7 +100,7 @@ class AttributeExtractor(BaseExtractor):
             self.add(element)
 
     def extract(self, pack: DataPack, instance: Annotation) -> Feature:
-        r"""Functionality: Extract attributes of one instance.
+        r"""Extract attributes of one instance.
         For example, the text of tokens in one sentence.
 
         Args:
@@ -148,7 +128,7 @@ class AttributeExtractor(BaseExtractor):
                        vocab=self.vocab)
 
     def pre_evaluation_action(self, pack: DataPack, instance: Annotation):
-        r"""Funcationality: Remove attributes of one instance. For
+        r"""Remove attributes of one instance. For
         example remove all pos tags of tokens in one sentence, if the
         entry_type is Token and the attribute is pos.  This function is
         called before the evaluation on a pack. After features are removed,
@@ -165,7 +145,7 @@ class AttributeExtractor(BaseExtractor):
 
     def add_to_pack(self, pack: DataPack, instance: Annotation,
                     prediction: Iterable[Union[int, Any]]):
-        r"""Functionality: Add the prediction for attribute to the
+        r"""Add the prediction for attribute to the
         instance. If the prediction is an iterable object, we assume
         each of the element in prediction will correspond to one entry.
         If the prediction is only one element, then we assume there will
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -48,7 +48,7 @@ class AttributeExtractor(BaseExtractor):
 
     def get_attribute(self, entry: Entry, attr: str) -> Any:
         r"""Get the attribute from entry. You can
-        overwrite this function if you have sepcial way to get the
+        overwrite this function if you have special way to get the
         attribute from entry.
 
         Args:
@@ -63,7 +63,7 @@ class AttributeExtractor(BaseExtractor):
 
     def set_attribute(self, entry: Entry, attr: str, value: Any):
         r"""Set the attribute of an entry to value.
-        You can overwrite this function if you have sepcial way to
+        You can overwrite this function if you have special way to
         set the attribute.
 
         Args:
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -1,0 +1,133 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Dict, Any, Union, Iterable
+from collections import abc
+from ft.onto.base_ontology import Entry, Annotation
+from forte.common.configuration import Config
+from forte.data.data_pack import DataPack
+from forte.data.converter.feature import Feature
+from forte.data.extractor.base_extractor import BaseExtractor
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "AttributeExtractor"
+]
+
+
+class AttributeExtractor(BaseExtractor):
+    r"""AttributeExtractor extracts feature from the attribute of entry.
+
+    Args:
+        config: An instance of `Dict` or
+            :class:`forte.common.configuration.Config`
+
+            attribute: str. Required. The attribute name of the
+                entry from which features will be extracted. For
+                example, "text" attribute of Token.
+    """
+    def __init__(self, config: Union[Dict, Config]):
+        super().__init__(config)
+
+        if "attribute" not in self.config:
+            raise AttributeError("attribute needs to be specified in "
+                                "the configuration of an AttributeExtractor.")
+
+    @property
+    def state(self) -> Dict:
+        state = super().state
+        state.update({
+            "attribute": self.config.attribute
+        })
+        return state
+
+    @classmethod
+    def from_state(cls, state: Dict) -> "AttributeExtractor":
+        obj = super().from_state(state)
+        additional = {
+            "attribute": state["attribute"]
+        }
+        obj.config = Config(obj.config, additional,
+                            allow_new_hparam=True)
+        obj = cls(obj)
+        return obj
+
+    @staticmethod
+    def get_attribute(entry: Entry, attr: str) -> Any:
+        r"""Functionality: Get the attribute from entry. You can
+        overwrite this function if you have sepcial way to get the
+        attribute from entry.
+        """
+        return getattr(entry, attr)
+
+    @staticmethod
+    def set_attribute(entry: Entry, attr: str, value: Any):
+        r"""Functionality: Set the attribute of an entry to value.
+        You can overwrite this function if you have sepcial way to
+        set the attribute.
+        """
+        if attr == "text":
+            raise AttributeError("text attribute of entry cannot "
+                                "be changed.")
+        setattr(entry, attr, value)
+
+    def update_vocab(self, pack: DataPack, instance: Annotation):
+        r"""Functionality: Get all attributes of one instance and
+        add them into the vocabulary.
+        """
+        for entry in pack.get(self.config.entry_type, instance):
+            self.add(self.get_attribute(entry, self.config.attribute))
+
+    def extract(self, pack: DataPack, instance: Annotation) -> Feature:
+        r"""Functionality: Extractor the attributes of one instance.
+        For example, the text of tokens in one sentence.
+        """
+        data = []
+        for entry in pack.get(self.config.entry_type, instance):
+            value = self.get_attribute(entry, self.config.attribute)
+            rep = self.element2repr(value) if self.vocab else value
+            data.append(rep)
+
+        meta_data = {"pad_value": self.get_pad_value(),
+                     "dim": 1,
+                     "dtype": int if self.vocab else Any}
+
+        return Feature(data=data,
+                       metadata=meta_data,
+                       vocab=self.vocab)
+
+    def remove_from_pack(self, pack: DataPack, instance: Annotation):
+        r"""Funcationality: Remove attributes of one instance. For
+        example remove all pos tags of tokens in one sentence, if the
+        entry_type is Token and the attribute is pos.
+        """
+        for entry in pack.get(self.config.entry_type, instance):
+            self.set_attribute(entry, self.config.attribute, None)
+
+    def add_to_pack(self, pack: DataPack, instance: Annotation,
+                    prediction: Iterable[Union[int, Any]]):
+        r"""Functionality: Add the prediction for attribute to the
+        instance. If the prediction is an iterable object, we assume
+        each of the element in prediction will correspond to one entry.
+        If the prediction is only one element, then we assume there will
+        only be one entry in the instance."""
+        instance_entry = list(pack.get(self.config.entry_type, instance))
+
+        if not isinstance(prediction, abc.Iterable):
+            prediction = [prediction]
+        values = [self.id2element(int(x)) for x in prediction]
+        for entry, value in zip(instance_entry, values):
+            self.set_attribute(entry, self.config.attribute, value)
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -46,8 +46,7 @@ class AttributeExtractor(BaseExtractor):
             raise AttributeError("attribute needs to be specified in "
                                 "the configuration of an AttributeExtractor.")
 
-    @staticmethod
-    def get_attribute(entry: Entry, attr: str) -> Any:
+    def get_attribute(self, entry: Entry, attr: str) -> Any:
         r"""Get the attribute from entry. You can
         overwrite this function if you have sepcial way to get the
         attribute from entry.
@@ -62,8 +61,7 @@ class AttributeExtractor(BaseExtractor):
         """
         return getattr(entry, attr)
 
-    @staticmethod
-    def set_attribute(entry: Entry, attr: str, value: Any):
+    def set_attribute(self, entry: Entry, attr: str, value: Any):
         r"""Set the attribute of an entry to value.
         You can overwrite this function if you have sepcial way to
         set the attribute.
--- a/forte/data/extractor/attribute_extractor.py
+++ b/forte/data/extractor/attribute_extractor.py
@@ -138,7 +138,8 @@ class AttributeExtractor(BaseExtractor):
             rep = self.element2repr(value) if self.vocab else value
             data.append(rep)
 
-        meta_data = {"pad_value": self.get_pad_value(),
+        meta_data = {"need_pad": self.config.need_pad,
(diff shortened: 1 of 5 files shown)


forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+    def state(self) -> Dict:
+        state = super().state
+        state.update({
+            "attribute": self.config.attribute
+        })
+        return state
+
+    @classmethod
+    def from_state(cls, state: Dict) \
+        -> Union["BaseExtractor", "AttributeExtractor"]:
```
Jiaqiang-Ruan, line 59: Inheriting class method is a little bit confuse. I am not sure whether there is a better way to write this.
At the same time, `Union["BaseExtractor", "AttributeExtractor"]` is my workaround for mypy type checking.

hunterhector: I see. This may be tricky for users to development. When I look at this, it seems this can be implemented in the `BaseExtractor` because most code here looks generic?

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+            raise AttributeError("text attribute of entry cannot "
+                                "be changed.")
+        setattr(entry, attr, value)
+
+    def update_vocab(self, pack: DataPack, instance: Annotation):
+        r"""Functionality: Get all attributes of one instance and
+        add them into the vocabulary.
+        """
+        for entry in pack.get(self.config.entry_type, instance):
+            self.add(self.get_attribute(entry, self.config.attribute))
```
hunterhector, line 93: Vocabulary need to take `Hashable` but `get_attribute` is `Any`. So what happens if we extract something that is not hashable?

To solve this problem, we should think about whether everything need to be placed into vocabulary, can something be skipped from the vocab? And how can users specify that?

Maybe one way to do this, is when the attribute is skipped from vocabulary, we may return the raw value?

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def extract(self, pack: DataPack, instance: Annotation) -> Feature:
+        r"""Functionality: Extractor the attributes of one instance.
```
hunterhector, line 96: Extractor -> Extract

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+        -> Union["BaseExtractor", "AttributeExtractor"]:
+        obj = BaseExtractor.from_state(state)
+        additional = {
+            "attribute": state["attribute"]
+        }
+        obj.config = Config(obj.config, additional,
+                            allow_new_hparam=True)
+        obj.__class__ = cls
+        return obj
+
```
hunterhector, line 68: Most of the docstring are missing arguments.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+                     "dtype": int if self.vocab else Any}
+
+        return Feature(data=data,
+                       metadata=meta_data,
+                       vocab=self.vocab)
+
+    def remove_from_pack(self, pack: DataPack, instance: Annotation):
+        r"""Funcationality: Remove attributes of one instance. For
+        example remove all pos tags of tokens in one sentence, if the
+        entry_type is Token and the attribute is pos.
```
hunterhector, line 116: Please mention the purpose of this function (is to before evaluation)

Further, I would suggest renaming `remove_from_pack` to `pre_evaluation_action`

hunterhector (changes requested):

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: I think we need to overwrite this function for each extractor. Because there will be some special field that you need to load for different extractor and you cannot assume they exists in the BaseExtractor.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it. I will add later.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: 1. Right we need to make sure the element added to vocabulary is Hashable type. So I think we need to make the assumption here or do an assertion here that if the user call update_vocab, they should be aware that they are using hashable attribute.
2. For the case, get_attribute return unhashable element. I think the user is better to set "vocal_method" to be "raw", so that there will not be vocabulary and update_vocab should not be called.

Jiaqiang-Ruan: https://github.com/asyml/forte/blob/900cf93880e83178a62c4d183f264268e3ea44c6/forte/data/extractor/base_extractor.py#L89

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: https://github.com/asyml/forte/blob/900cf93880e83178a62c4d183f264268e3ea44c6/forte/data/extractor/base_extractor.py#L172

forte/data/extractor/attribute_extractor.py
hunterhector: The implementation you provided here actually is class agnostic right? Which part of this is class dependent?

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+        additional = {
+            "attribute": state["attribute"]
+        }
+        obj.config = Config(obj.config, additional,
+                            allow_new_hparam=True)
+        obj.__class__ = cls
+        return obj
+
+    @staticmethod
+    def get_attribute(entry: Entry, attr: str) -> Any:
```
hunterhector, line 70: We may actually move these staticmethod to a util file, like here https://github.com/asyml/forte/blob/master/forte/data/data_utils.py

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+from forte.data.extractor.base_extractor import BaseExtractor
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "AttributeExtractor"
+]
+
+
+class AttributeExtractor(BaseExtractor):
```
hunterhector, line 31: One thing I forgot to mention.

For future parallel programming purposes. We need to reduce the amount of internal states. One thing you notice that is shared across the whole pipeline is `Resources` in the processor initialize. If we put all shared states to `Resources` that would be much easier for us to parallel them in the future.

Could you pass in the `Resources` class here? And put the `self.vocab` from Extractor into `Resources`? However, you need to use a specific key when you are adding to the resource. It cannot be the class name because there may be multiple `attribute_extractor`.

Another option is to make the object `Vocabulary` parallel in the future.

forte/data/extractor/base_extractor.py
```diff
@@ -257,17 +257,25 @@ def extract(self, pack: DataPack,
...
         pass
 
-    def remove_from_pack(self, pack: DataPack,
+    def pre_evaluation_action(self, pack: DataPack,
                 instance: Annotation):
         r"""Functionality: Remove the existing feature of the instance
         in a pack. We might remove an attribute under an entry or remove
         the entry itself directly, which will depend on the different
-        type of extractors.
+        type of extractors. This function is called before the evaluation
```
hunterhector, line 265: So `removal` won't be the main functionalities. You can simply say this function is called before evaluation. The users can use this chance to remove existing gold standards or do anything else. We don't assume that they will only do the deletion. We can suggest that one common thing to do is deletion.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+                instance.
+            instance (Annotation): The instance from which the
+                extractor will extractor feature.
+        """
+        for entry in pack.get(self.config.entry_type, instance):
+            self.set_attribute(entry, self.config.attribute, None)
+
+    def add_to_pack(self, pack: DataPack, instance: Annotation,
+                    prediction: Iterable[Union[int, Any]]):
+        r"""Functionality: Add the prediction for attribute to the
```
hunterhector, line 168: You can remove this `Functionality:` start.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+                     "pad_value": self.get_pad_value(),
+                     "dim": 1,
+                     "dtype": int if self.vocab else Any}
+
+        return Feature(data=data,
+                       metadata=meta_data,
+                       vocab=self.vocab)
+
+    def pre_evaluation_action(self, pack: DataPack, instance: Annotation):
+        r"""Funcationality: Remove attributes of one instance. For
```
hunterhector, line 151: You can remove this Functionality: start.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+            attr (str): The name of the attribute.
+            value (Any): The value to be set for the attribute.
+        """
+        if attr == "text":
+            raise AttributeError("text attribute of entry cannot "
+                                "be changed.")
+        setattr(entry, attr, value)
+
+    def update_vocab(self, pack: DataPack, instance: Annotation):
+        r"""Functionality: Get all attributes of one instance and
```
hunterhector, line 103: You can remove this `Functionality:` start.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+            if not isinstance(element, Hashable):
+                raise AttributeError("Only hashable element can be"
+                    "added into the vocabulary. Consider setting"
+                    "vocab_method to be raw and do not call update_vocab"
+                    "if you only need the raw attribute value without"
+                    "coverting them into index.")
+            self.add(element)
+
+    def extract(self, pack: DataPack, instance: Annotation) -> Feature:
+        r"""Functionality: Extract attributes of one instance.
```
hunterhector, line 123: You can remove this `Functionality:` start.

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,189 @@
...
+        For example, the text of tokens in one sentence.
```
hunterhector, line 124: `text` is a special feature we'd better mention this?

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: I think we might be talking different things here. Here is how I am planning to implement this function.

1. Any subclass from BaseExtractor will first call `BaseExtractor.from_state` to get a BaseExtractor object. 
2. Then this function tries to convert the BaseExtractor into an object of its own class.
     a. adding more fields
     b. change class by `obj.__class__`, but mypy checking cannot recognize this, so I have to write `Union["BaseExtractor", "AttributeExtractor"]` as return type.
3. in the future, if there are more than two layers of inheritance, let's say A(parent) -> B(subclass) -> C(subsbuclass). Inside the C.from_state, B.from_state should be first called and a B object is created. Then convert that B obj to a C obj.

hunterhector: So I think there are two parts in the state:

1. The `Config` object. For this, we should be able to dump and load with default python serialization? Just up the states from `Config` to a key `config`. So you don't need to write the key for each one. In this case, all subclasses don't have to do the same thing again.
2. The `Vocab` object, you can also do the same? 

At least for these objects, we shouldn't repeat the implementation. 

In fact, see my suggestion above of using "Resources", where we can put vocab inside, I actually suggest we put all other states there too, so we don't have to worry about states per class.

So the pseudo code in the base class is like:

```python
def __state__(self):
   return {
       'config': self.config.to_dict(),
       'vocab': self.vocab.state(),  # may remove if we use resource
   }

def __getstate__(self, state):
    self.config = Config.parse(state['config'])
    self.vocab = self.vocab.from_state(state['vocab']) # may remove if we use resource
```

By the way, why do we need to use classmethod here? Python's default pickle functions are these: https://docs.python.org/3/library/pickle.html#object.__getstate__

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Extractor currently is not a Processor, so there is no Resource and there is no initialize method. And I am not sure whether it is compatible to processor right now.

I take a look at Resource class. If we put self.vocab into resources directly, it will be saving the class object directly, which might conflict with our current design, turning class object into state and save the state. 

Can you be a little bit more specific how we are going to use this Resource class? For example,
1. create a resource class in __init__ or receive a resource class from outside
2. put self.vocab into resource or self.vocab.state into resource
3. what is the relation between this resource class and the state, from_state function?

hunterhector: 1. You don't have to make these Processors, you can simply pass a `Resources` from a processor.
2. We shouldn't create our own name `from_state`, `state`. Instead, if we implement `__getstate__`, `__setstate__` correctly, the pickle function will work automatically. [get_state](https://docs.python.org/3/library/pickle.html#object.__getstate__
). So you can put `self.vocab` to resources, and it will be load back automatically. Note that these are not classmethod, but instance method.

hunterhector: Take a look at DataPack serialisation as example:

https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L86

Jiaqiang-Ruan: I see. This seems to be a better solution. I will look into the __setstate__ and __getstate__ and update my implementation.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: I think we can only mention it in the `set_attribute`? Because it is `extract` here. Text behaves the same as other attribute.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: got it.

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: got it.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/attribute_extractor.py
hunterhector: ok, this is not a big deal.

Jiaqiang-Ruan:
1. Since the there is default `__setattr__` and `__getattr__`, there is no need to implement them in AttributeExtrator.
2. As for the Resource, after discussed with @jasonyanwenl, this is going to be implemented in next version.

forte/data/extractor/base_extractor.py
```diff
@@ -257,16 +257,25 @@ def extract(self, pack: DataPack,
         """
         pass
 
-    def remove_from_pack(self, pack: DataPack, instance: Annotation):
+    def pre_evaluation_action(self, pack: DataPack,
+                instance: Annotation):
         r"""Functionality: Remove the existing feature of the instance
         in a pack. We might remove an attribute under an entry or remove
         the entry itself directly, which will depend on the different
```
hunterhector, line 264: Removing attribute is just one of the possible actions, you can say when this function will be called, and then say what are the possible actions that could happen here.

forte/data/extractor/base_extractor.py
```diff
@@ -257,16 +257,25 @@ def extract(self, pack: DataPack,
...
-        type of extractors.
+        type of extractors. This function is called before the evaluation
+        on a pack. After features are removed, new features predicted
+        from the model will be added to the pack.
 
         Overwrite instruction:
             1. Get all entries from one instance in the pack.
             2. Remove entries or remove some attributes of the entry. You
                 might use pack.delete_entry function or setattr to do this.
```
hunterhector, line 272: Removing should not be the core instruction

tests/forte/data/extractor/attribute_extractor_test.py
```diff
@@ -0,0 +1,97 @@
...
+        for pack in pipeline.process_dataset(self.dataset_path):
+            features = []
+            for instance in pack.get(Sentence):
+                features.append(extractor.extract(pack, instance))
+
+            for feat in features:
+                recovered = [extractor.id2element(idx) for idx in feat.data[0]]
+                self.assertEqual(" ".join(recovered), sentence)
+
+        # Check add_to_pack and remove_from_pack.
```
hunterhector, line 64: Function `remove_from_pack` don't exist any more.

tests/forte/data/extractor/attribute_extractor_test.py
```diff
@@ -0,0 +1,97 @@
...
+        # Vocab_mathod is indexing, therefore the id of element
```
hunterhector, line 65: typo?

Jiaqiang-Ruan:
Here is a question on using a universal Resource to store the vocabulary of each extractor.

When we pickle the extractor, should we dump the resource? Because it it is going to a universal resource that save all vocabularies. So we should dump the universal resource as one file and each extractor with config into another file?

If so, we also need a function in extractor to set vocab from the loaded resource.

hunterhector (commented):
Besides the comments, there is a note for using adding data to the pack. If you add it through a processor, then it will do some automatic bookkeeping, such as remembering what processor created the pack. So we should not encourage these extractors to be called directly, but always through some processor.

hunterhector:
> Here is a question on using a universal Resource to store the vocabulary of each extractor.
> 
> When we pickle the extractor, should we dump the resource? Because it it is going to a universal resource that save all vocabularies. So we should dump the universal resource as one file and each extractor with config into another file?
> 
> If so, we also need a function in extractor to set vocab from the loaded resource.

Let's discuss this when we reach that part.

forte/data/extractor/base_extractor.py
```diff
@@ -257,16 +257,25 @@ def extract(self, pack: DataPack,
         """
         pass
 
-    def remove_from_pack(self, pack: DataPack, instance: Annotation):
+    def pre_evaluation_action(self, pack: DataPack,
+                instance: Annotation):
         r"""Functionality: Remove the existing feature of the instance
```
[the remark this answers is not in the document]
Jiaqiang-Ruan: Sorry. I should have deleted this one. It's a mistake.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

tests/forte/data/extractor/attribute_extractor_test.py
Jiaqiang-Ruan: Yep.

tests/forte/data/extractor/attribute_extractor_test.py
Jiaqiang-Ruan: Got it.

Jiaqiang-Ruan:
Travis check fails because of misspelled words.
https://travis-ci.org/github/asyml/forte/jobs/753310995

forte/data/extractor/attribute_extractor.py
```diff
@@ -0,0 +1,167 @@
...
+                     "pad_value": self.get_pad_value(),
+                     "dim": 1,
+                     "dtype": int if self.vocab else Any}
+
+        return Feature(data=data,
+                       metadata=meta_data,
+                       vocab=self.vocab)
+
+    def pre_evaluation_action(self, pack: DataPack, instance: Annotation):
+        r"""Remove attributes of one instance. For
```
hunterhector, line 129: "Remove attributes" is not the only possible action.

You can simply say:

"This function is performed on the pack before the evaluation stage, allowing one to perform some actions to before the evaluation. By default, this function will remove the attribute and you can overwrite this ......"

forte/data/extractor/base_extractor.py
```diff
@@ -231,16 +231,25 @@ def extract(self, pack: DataPack,
         """
         pass
 
-    def remove_from_pack(self, pack: DataPack, instance: Annotation):
-        r"""Functionality: Remove the existing feature of the instance
+    def pre_evaluation_action(self, pack: DataPack,
+                instance: Annotation):
+        r"""Remove the existing feature of the instance
```
hunterhector, line 236: The base extractor should be more general, please put something like this at the begining: "This function is performed on the pack before the evaluation stage, allowing one to perform some actions to before the evaluation."

Note that deleting stuff is just an example.

Also, mentioning that "by default, this function will not do anything"

forte/data/extractor/base_extractor.py
```diff
@@ -231,16 +231,25 @@ def extract(self, pack: DataPack,
...
+        type of extractors. This function is called before the evaluation
+        on a pack to do some pre_evaluation action. For example, you can
+        remove entries or remove some attributes of the entry. So that
+        they are not mixed with predicted entries or predicted attributes.
 
         Overwrite instruction:
             1. Get all entries from one instance in the pack.
-            2. Remove entries or remove some attributes of the entry. You
-                might use pack.delete_entry function or setattr to do this.
+            2. Conduct pre_evaluation_action on entries.
```
hunterhector, line 246: `pre_evaluation_action` is the function itself, this is confusing.

forte/data/extractor/base_extractor.py
```diff
@@ -231,16 +231,25 @@ def extract(self, pack: DataPack,
...
+        r"""Remove the existing feature of the instance
         in a pack. We might remove an attribute under an entry or remove
         the entry itself directly, which will depend on the different
-        type of extractors.
+        type of extractors. This function is called before the evaluation
+        on a pack to do some pre_evaluation action. For example, you can
+        remove entries or remove some attributes of the entry. So that
+        they are not mixed with predicted entries or predicted attributes.
 
         Overwrite instruction:
```
hunterhector, line 244: I don't think you need this instruction. You just need to tell users what this function should do.

hunterhector (commented):
I guess `extractor` will not be called directly by the users, by from our framework mostly. Can we specify this in some part of the documentation?

forte/data/extractor/attribute_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: I wil remove the instruction.

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

Jiaqiang-Ruan:
I add some documentation at the very beginning of base extractor, saying that this class will be used internally.

hunterhector (approved):

hunterhector (approved):

PR #361, opened 2021-01-07
[vocab/extractor] use pickle to dump and load Vocabulary and Extractor

Jiaqiang-Ruan:
### Description of changes
use `__getstate__` and `__setstate__` to save vocabulary instead of `state` and `from_state` function.

The way to save vocabulary will be changed. After this PR, users can use pickle to save vocabulary directly.

### Test Conducted
Pickle load and dump is tested.

### DIFF
--- a/forte/data/extractor/base_extractor.py
+++ b/forte/data/extractor/base_extractor.py
@@ -119,19 +119,6 @@ class BaseExtractor(ABC):
             "need_pad": True,
         }
 
-    def __getstate__(self):
-        r"""For serializaion."""
-        state = {
-            "config": self.config,
-            "vocab": self.vocab
-        }
-        return state
-
-    def __setstate__(self, state):
-        r"""For deserialization."""
-        self.config = state["config"]
-        self.vocab = state["vocab"]
-
     @property
     def entry_type(self) -> Type[Annotation]:
         return self.config.entry_type
--- a/forte/data/extractor/base_extractor.py
+++ b/forte/data/extractor/base_extractor.py
@@ -119,31 +119,18 @@ class BaseExtractor(ABC):
             "need_pad": True,
         }
 
-    @property
-    def state(self) -> Dict:
-        return {
-            "vocab_method": self.config.vocab_method,
-            "need_pad": self.config.need_pad,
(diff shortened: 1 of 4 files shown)


forte/data/vocabulary.py
```diff
@@ -116,13 +116,14 @@ def __init__(self, method: str, need_pad: bool, use_unk: bool):
...
     def __getstate__(self):
         r"""For serialization."""
-        state = {}
-        state["method"] = self.method
-        state["need_pad"] = self.need_pad
-        state["use_unk"] = self.use_unk
-        state["element2id_dict"] = self.element2id_dict
-        state["id2element_dict"] = self.id2element_dict
-        state["next_id"] = self.next_id
+        state = {
```
hunterhector, line 119: Is this function simply return everything of this vocab class into a dict? If so, you don't need to implement __getstate__ and __setstate__ at all. Python's default `__dict__` will be used (https://docs.python.org/3/library/pickle.html#object.__getstate__)

forte/data/extractor/base_extractor.py
```diff
@@ -119,31 +119,18 @@ def default_configs():
...
-        return {
-            "vocab_method": self.config.vocab_method,
-            "need_pad": self.config.need_pad,
-            "vocab_use_unk": self.config.vocab_use_unk,
-            "entry_type": self.config.entry_type,
-            "vocab": self.vocab.state if self.vocab else None,
+    def __getstate__(self):
+        r"""For serializaion."""
+        state = {
+            "config": self.config,
```
hunterhector, line 125: Similar to the other comment, if your serialization step actually saves all class fields, without special treatment, you don't need to implement __getstate__ and __setstate__

hunterhector (commented):
I think these classes can be directly pickled

forte/data/extractor/base_extractor.py
Jiaqiang-Ruan: Got it.

forte/data/vocabulary.py
Jiaqiang-Ruan: That's good to know. Updated.

hunterhector (approved):
should be find

PR #367, opened 2021-01-18
extended allennlp processors to include srl

jieralice13:
This PR fixes [issue link]. 

### Description of changes
Extended Allennlp processors to include Semantic Role Labeling by adding a new func _create_srl

Make users have the option of using AllenNLP SRL predictor

### Test Conducted
Tested in pipeline example


forte/processors/allennlp_processors.py
```diff
@@ -48,8 +51,13 @@ def initialize(self, resources: Resources, configs: Config):
 
         if configs.tag_formalism not in MODEL2URL:
             raise ProcessorConfigError('Incorrect value for tag_formalism')
-        model_url = MODEL2URL[configs.tag_formalism]
-        self.predictor: Predictor = Predictor.from_path(model_url)
+        if configs.tag_formalism == 'stanford':
+            self.predictor = {
+                'stanford': Predictor.from_path(MODEL2URL['stanford'])}
+        if configs.tag_formalism == 'srl':
```
hunterhector, line 57: Please read the documentation for the meaning of `config` here: https://github.com/asyml/forte/blob/3a09d8d7db4a79e0b7f65aebf8ae0b1902dd1d91/forte/processors/allennlp_processors.py#L93

`tag_formalism` is only applicable to the `parsing` predictor. What you should check here is `processors` contains `srl`

forte/processors/allennlp_processors.py
```diff
@@ -153,3 +165,39 @@ def _create_dependencies(input_pack, tokens, result):
                                   parent=tokens[heads[i] - 1],
                                   child=token)
             relation.rel_type = deps[i]
+
+    @staticmethod
+    def _create_srl(input_pack, tokens, result):
```
hunterhector, line 170: Let's add typing to the variables and the return types. I know the existing functions don't have these. But we are enforcing it a bit more strict now.

It would be great if you also add the typing for those functions.

forte/processors/allennlp_processors.py
```diff
@@ -153,3 +165,39 @@ def _create_dependencies(input_pack, tokens, result):
...
+                    begin = i
+                    end = i
+                if item.startswith('I-'):
+                    end = i
+            return pred_span, arguments
+        verbs = result['verbs']
+
+        for _, verb_item in enumerate(verbs):
+            pred_span, arguments = parse_allennlp_srl_tags(verb_item['tags'])
+            if arguments == []:
```
hunterhector, line 194: There may be chances where we have the `Predicate`, but cannot find the arguments for it.  Note that this is still useful for downstream tasks.

hunterhector (commented):
Generally looks good. There are some small comments. And we are waiting for the actual test cases to merge this.

forte/processors/allennlp_processors.py
jieralice13: Got it. Will modify it

forte/processors/allennlp_processors.py
jieralice13: Sure, will do

forte/processors/allennlp_processors.py
jieralice13: So i just remove line 194 and 195? But it won't be saved in PredicateLink. Is it OK?

hunterhector: If there are no PredicateLink then yes, that's the expected output. We should not modify the output here.

forte/utils/utils_processor.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2019 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: 2021 I guess

forte/utils/utils_processor.py
```diff
@@ -0,0 +1,78 @@
...
+    return parsed_results
+
+
+def parse_allennlp_srl_tags(tags: str) -> \
+        Tuple[Span, List[Tuple[Span, str]]]:
+    r"""Parse the tag list of a specific verb output by
+    AllenNLP SRL processor.
+
+    Args:
+        tags (str): a str of semantic role lables.
```
hunterhector, line 55: labels

forte/utils/utils_processor.py
```diff
@@ -0,0 +1,78 @@
...
+                pred_span = Span(begin, end)
+            else:
+                arg_span = Span(begin, end)
+                arguments.append((arg_span, prev_argument))
+        prev_argument = argument
+        if item.startswith('B-'):
+            begin = i
+            end = i
+        if item.startswith('I-'):
+            end = i
```
hunterhector, line 77: There can be a small danger here, where the output tag from AllenNLP is buggy. We may have the chances where we haven't seen a `B-` but we saw an `I-`. Then there may be a danger where begin is `None`, or the previous `begin` value.

hunterhector (dismissed):

hunterhector (approved):

PR #369, opened 2021-01-29
update batcher implementation

Jiaqiang-Ruan:
### Description of changes
This function update the implementation of batcher. Yield tuple of (pack, instance, feature) to avoid using confusing update pool function.

This is internal implementation change. There shouldn't be any change in the user code.

### Test Conducted
Original test passed.


forte/data/batchers.py
```diff
@@ -84,32 +84,34 @@ def flush(self) -> Iterator[Dict]:
 
         """
         if self.current_batch:
-            yield self.current_batch
+            yield self.data_pack_pool, None, self.current_batch
```
hunterhector, line 87: I guess the return value type should change? Will be helpful to add what are these 3 in the `Returns:` doc too.

hunterhector: Also good to mention why it is always `None` (i.e. ignored here).

forte/processors/base/batch_processor.py
```diff
@@ -179,17 +174,27 @@ def predict(self, data_batch: Dict) -> Dict:
         """
         pass
 
-    def pack_all(self, output_dict: Dict):
+    def pack_all(self, packs: List[DataPack], output_dict: Dict):
```
hunterhector, line 177: I believe there are other classes that are using this function? Maybe need to add something there too? Otherwise the travis may break.

hunterhector (commented):
This PR will largely rely on whether CI can pass. I have a few small comments.

forte/data/batchers.py
Jiaqiang-Ruan: Got it.

forte/processors/base/batch_processor.py
Jiaqiang-Ruan: I passed the test on pipeline. I will search if there are any other place.

Jiaqiang-Ruan: It seems this function is only used in this class. We will see whether travis works.

hunterhector: If you use some IDE, you should be able to find the usage of it. Otherwise, you can search for the function. Or travis will tell you if it breaks.

Jiaqiang-Ruan: Yes. And this function should only used in this file.

hunterhector (approved):

PR #382, opened 2021-02-28
Predictor Patch

Jiaqiang-Ruan:
RJQ: overwrite the flush function in the predictor (missed in previous PR)

### Description of changes
Add one missed out overwrite function for predictor

User code do not need to change. 

### Test Conducted
Pipeline test done.


hunterhector:
The PR fails because of another repo. I will fix that part.

hunterhector (approved):
should be fine.

PR #396, opened 2021-03-25
Add processor input output type checking and record writing

### LINKED ISSUE #397: Add processor input output type checking and record writing
For forte processors, some processor would require a specific entry type as input. We don't have it checked anywhere with current forte implementation so far. We would like an error message poping up if expected type is not satisfied for the input datapack.

Add a function to enforce consistency checking for whole pipeline and each pipeline component. For the processor that does input type checking we add "expected_type" function to specify type required for this processor and for the processor that produce specific output type, we add "record" function to specify the output type for the next processor in the pipeline.

jennyzhang-petuum:
This PR fixes #397 

### Description of changes
For forte processors, some processor would require a specific entry type as input. We added "record" dictionary as an attribute to "Meta" of the datapack and for the processor that does input type checking we add "expected_type" function and for the processor that produce specific output type, we added "record" function. We also added function to enforce consistency checking for whole pipeline 
"record" would be added to meta of any pack. 
### Test Conducted
Added `RecordCheckPipelineTest` in pipeline_test.py


hunterhector:
Can you create the issue for tracking purpose?

forte/data/readers/base_reader.py
```diff
@@ -222,16 +222,27 @@ def iter(self, *args, **kwargs) -> Iterator[PackType]:
...
                     self._data_packs.append(pack)
                 yield pack
 
         self._cache_ready = True
 
+    def record(self, record_meta: Dict):
+        r"""Modify the datapack meta record field
+        Args:
+            record_meta: the field in the datapack for type record that need to fill in
+            for consistency checking.
```
hunterhector, line 241: I think you need to have an indention if the docstring for an argument is longer than a line.

hunterhector: Here is an example of wrong formatted docstring:
https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L290
https://asyml-forte.readthedocs.io/en/latest/code/data.html#forte.data.data_pack.DataPack.get_original_span

And here is an example of the correct format:
https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L414
https://asyml-forte.readthedocs.io/en/latest/code/data.html#forte.data.data_pack.DataPack.deserialize

forte/evaluation/base/base_evaluator.py
```diff
@@ -47,3 +47,10 @@ def get_result(self) -> Any:
         r"""The evaluator gather the results and the score can be obtained here.
         """
         raise NotImplementedError
+
+    def record(self, record_meta: Dict):
+        pass
+
+    @classmethod
+    def expected_type(cls) -> Dict:
```
hunterhector, line 55: Could you write the docstring here so the implementer of this class knows what to implement here?

forte/evaluation/base/base_evaluator.py
```diff
@@ -47,3 +47,10 @@ def get_result(self) -> Any:
         r"""The evaluator gather the results and the score can be obtained here.
         """
         raise NotImplementedError
+
+    def record(self, record_meta: Dict):
```
hunterhector, line 51: Can you explain what to do if the user need to implement this function

forte/pipeline.py
```diff
@@ -120,6 +120,14 @@ def __init__(self, resource: Optional[Resources] = None):
             self.resource = resource
 
         self.initialized: bool = False
+        self._check_type_consistency: bool = False
+
+    def enforce_consistency(self):
```
hunterhector, line 125: Not you can enforce the consistency but cannot disable it again. There are two options:
1) add a boolean as parameter here
2) use another function to disable it

forte/pipeline.py
```diff
@@ -120,6 +120,14 @@ def __init__(self, resource: Optional[Resources] = None):
...
+        r"""This function enforces a type consistency checking for
+        all the pipeline components.
```
hunterhector, line 127: Please explain what does type consistency check is. The users won't understand what is happening here. You can use an example to explain this.

forte/processors/spacy_processors.py
```diff
@@ -126,3 +126,6 @@ def _process(self, input_pack: DataPack):
 
         # Process sentence parses.
         self._process_parser(result.sents, input_pack)
+
+    def record(self, record_meta: Dict):
```
hunterhector, line 130: Use docstring to explain what is being recorded here.

forte/processors/base/base_processor.py
```diff
@@ -38,11 +39,34 @@ def __init__(self):
...
+            for expected_t in expected_types:
+                if expected_t not in input_pack._meta.record.keys():
+                    raise ExpectedRecordNotFound(
+                        f"The record type {expected_t} is not found in "
+                        f"meta of the provided pack.")
         # Set the component for recording purpose.
         input_pack.set_control_component(self.name)
         self._process(input_pack)
-
+        # add type record of output to meta
```
hunterhector, line 68: Use full sentences for comments if we can. Here: 
"Add type records of output to the `meta` of the `input_pack`."

forte/processors/base/base_processor.py
```diff
@@ -38,11 +39,34 @@ def __init__(self):
...
+        would be checked before running the processor if enforce_consistency()
+        was called for the pipeline.
+        """
+        return {}
+
     def process(self, input_pack: PackType):
+        if self._check_type_consistency:
+            expected_types = self.expected_type()
+            # check if expected types are in input pack
+            for expected_t in expected_types:
```
hunterhector, line 60: Here only the type is checked, but the attributes are not checked. Can you check for those as well?

forte/pipeline.py
```diff
@@ -176,6 +186,8 @@ def initialize_processors(self):
             try:
                 processor.assign_manager(self._proc_mgr)
                 processor.initialize(self.resource, config)
+                if self._check_type_consistency:
+                    processor.enforce_consistency()
```
hunterhector, line 190: how about the check for evaluator?

hunterhector (commented):
See comments inline.

forte/pipeline.py
jennyzhang-petuum: From my understanding, Evaluator inherets PipelineComponent and its also part of self.components here. Would it be also initialized and set _check_type_consistency here?

hunterhector: I see. The code wrote "processor" so I thought this is just for "processor", but here it actually mean "evaluator"

forte/pipeline.py
```diff
@@ -119,6 +119,28 @@ def __init__(self, resource: Optional[Resources] = None):
             self.resource = resource
 
         self.initialized: bool = False
+        self._check_type_consistency: bool = False
+
+    def enforce_consistency(self, enforce: bool = True):
+        r"""This function enforces a type consistency checking for
+        all the pipeline components. When this function is called
+        with enforce is ``True``, all the pipeline component would check
```
hunterhector, line 127: components

forte/pipeline.py
```diff
@@ -119,6 +119,28 @@ def __init__(self, resource: Optional[Resources] = None):
...
+        if the input datapack record matches with the expected type
+        if function ``expected_type`` is implemented for the processor.
+        For example, Allennlp processor requires entry type of
+        ``ft.onto.base_ontology.Sentence``, and Spacy processor would
```
hunterhector, line 131: Let's do not write the name of the processors, those requirements may change.

forte/pipeline.py
```diff
@@ -119,6 +119,28 @@ def __init__(self, resource: Optional[Resources] = None):
             self.resource = resource
 
         self.initialized: bool = False
+        self._check_type_consistency: bool = False
+
+    def enforce_consistency(self, enforce: bool = True):
+        r"""This function enforces a type consistency checking for
+        all the pipeline components. When this function is called
```
hunterhector, line 126: Please explain the behavior at the beginning in a concise way, then another paragraph for the detailed description:

This function determines whether the pipeline will enforce the content expectations specified in each pipeline component. Each component will check whether the input pack contains the expected data via checking the meta-data, and throws a `forte.common.exception.ExpectedEntryNotFound` if the check fails.

forte/pipeline_component.py
```diff
@@ -31,6 +31,14 @@ class PipelineComponent(Generic[PackType]):
     def __init__(self):
         self.resources: Resources = Resources()
         self.configs: Config = Config({}, {})
+        self._check_type_consistency = False
+
+    def enforce_consistency(self):
```
hunterhector, line 36: You should probably add a boolean input just like the `pipeline`: `    def enforce_consistency(self, enforce: bool = True):`

forte/processors/base/base_processor.py
```diff
@@ -36,7 +37,62 @@ def __init__(self):
         super().__init__()
         self.selector = DummySelector()
 
+    def record(self, record_meta: Dict):
+        r"""Method to add output type record of current processor
```
hunterhector, line 41: of the current processor

forte/processors/base/base_processor.py
```diff
@@ -36,7 +37,62 @@ def __init__(self):
...
+        the pipeline.
+        """
+        return {}
+
+    def check_record_and_writes_record(self, input_pack: PackType):
+        # pylint: disable=protected-access
+        r"""Method to check type consistency if
+        :meth:`~forte.pipeline.Pipeline.enforce_consistency` is enabled
+        for the pipeline and write record of the output type of the current
+         processor to the datapack. If the expected type or its attributes
```
hunterhector, line 64: weird indent here.

forte/processors/base/base_processor.py
```diff
@@ -36,7 +37,62 @@ def __init__(self):
...
+    @classmethod
+    def expected_type(cls) -> Dict[str, List[str]]:
+        r"""Method to add expected type for current processor input which
+        would be checked before running the processor if
+        :meth:`~forte.pipeline.Pipeline.enforce_consistency` was enabled for
+        the pipeline.
+        """
+        return {}
+
+    def check_record_and_writes_record(self, input_pack: PackType):
```
hunterhector, line 59: Like the previous comment, probably better split this function to two.

forte/processors/base/base_processor.py
```diff
@@ -36,7 +37,62 @@ def __init__(self):
...
+        to :attr:`forte.data.data_pack.Meta.record`
+
+        Args:
+            record_meta: the field in the datapack for type record that need to
+                fill in for consistency checking.
+        """
+        pass
+
+    @classmethod
+    def expected_type(cls) -> Dict[str, List[str]]:
```
hunterhector, line 51: Just for the name, it might not be simply `type`, attributes are also included.

hunterhector (commented):
I realize we don't have tests for the new code. Can you add the tests?

forte/utils/utils_processor.py
```diff
@@ -80,3 +83,38 @@ def parse_allennlp_srl_tags(tags: str) -> \
...
+        for expected_t in expectation:
+            if expected_t not in input_pack._meta.record.keys():
+                raise ExpectedRecordNotFound(
+                    f"The record type {expected_t} is not found in "
+                    f"meta of the prediction datapack.")
+            else:
+                expected_value = expectation.get(expected_t)
+                if expected_value is not None:
+                    for expected_t_v in expected_value:
+                        if expected_t_v not in input_pack._meta.record \
```
hunterhector, line 114: If we are checking for `contains` (i.e x in y), should we use `set()` instead of `list()`?

tests/forte/pipeline_with_record_check_test.py
```diff
@@ -0,0 +1,183 @@
...
+import unittest
+from typing import Dict, List
+
+from forte.data.base_pack import PackType
+from forte.data.data_pack import DataPack
+from forte.evaluation.base import Evaluator
+from forte.common import ProcessExecutionException
+from forte.pipeline import Pipeline
+
+from tests.forte.pipeline_test import SentenceReader, DummyPackProcessor
```
hunterhector, line 28: how about we put this test directly to the `pipeline_test`? Having some dependencies between tests may not be a good idea. We do have such imports before but I am considering removing them.

tests/forte/pipeline_with_record_check_test.py
```diff
@@ -0,0 +1,183 @@
...
+        nlp.enforce_consistency(enforce=True)
+        reader = DummySentenceReader()
+        nlp.set_reader(reader)
+        dummy = DummyPackProcessorTwo()
+        nlp.add(dummy)
+        nlp.initialize()
+        data_path = data_samples_root + "/random_texts/0.txt"
+        with self.assertRaises(ProcessExecutionException):
+            nlp.process(data_path)
+
```
hunterhector, line 151: How about we also test for `enforce=False` right after this? i.e.:

```suggestion
nlp.enforce_consistency(enforce=False)
nlp.process(data_path)
# can do some check here, or do nothing because if the exception is not thrown, it should be ok.
```

hunterhector (commented):
a few small comments.

hunterhector (approved):

PR #404, opened 2021-03-29
More config options for AllenNLPProcessor

DianLiI:
### Description of changes
Adding two types for config options in AllenNLPProcessor
1. Model URL's.
2. Use gpu and basic control of gpu assignment (for each model)

N/A

### Test Conducted
Tested against a pipeline using AllenNLPProcessor.


hunterhector:
We plan to put all wrappers to the other repository in the future:　　

https://github.com/asyml/forte-wrappers

Do you think we should still make the change in this repository?

Issue #401, opened 2021-03-26
Profiling time cost of each processor


hunterhector:
**Is your feature request related to a problem? Please describe.**
When running a pipeline with multiple processors, it is helpful to know the resource and time consumption of each processor. These can actually be measured by the framework. 

**Describe the solution you'd like**
1. Provide an option for the pipeline to start the profiling.
2. The system profiles the time of each processor.
3. The system can also profile other relevant resources if needed, such as CPU and GPU, but if this profiling is expensive, we should probably leave them to 3rd party profiling tools.

The main processing happens around here:
https://github.com/asyml/forte/blob/87f243aa0cdfe2914e1541e798b5fdb1a8650b26/forte/pipeline.py#L341

which is called here:
https://github.com/asyml/forte/blob/87f243aa0cdfe2914e1541e798b5fdb1a8650b26/forte/pipeline.py#L515

So it is possible to inject some timer around these lines.

**Describe alternatives you've considered**

**Additional context**

PR #421, opened 2021-04-07
Profiling time cost of each processor

mylibrar:
This PR fixes https://github.com/asyml/forte/issues/401 (Profiling time cost of each processor). 

### Description of changes
1. Provide an option for the pipeline to start the profiling. Add a boolean **Pipeline._enable_profiling** to the Pipeline class to enable time profiling. User can toggle the boolean to enable/disable profiling of pipeline with **Pipeline.set_profiling()**. To enable profiling, you should call the method before initializing the pipeline.
2. The system profiles the time of each processor. Add a float list **Pipeline._profiler** to aggregate time cost for each processors. Timers are injected around each call of **Pipeline._process_with_component()** and each time cost is added to the corresponding component index in **Pipeline._profiler**. 
3. **BaseReader** is modified to support time profiling of readers. Timers are injected around the yield statement in **BaseReader.iter()** to time the process of reading each data pack. Profiler info will be logged with the following format:
    - Reader: Name_Of_Reader, [time_cost (in seconds)]
    - Component [0]: Name of it, [time_cost (in seconds)] 
    - Component [1]: Name of it, [time_cost (in seconds)]
    - ...

May slow down the execution of relevant operations for readers and processors on the pipeline.

### Test Conducted
A unit test for time profiling is added to check logger info. The test specifies a dummy PackProcessor which can arbitrarily delay the process. The test will then check the logger info to verify that the corresponding delays are correctly logger.

### DIFF
--- a/forte/data/base_reader.py
+++ b/forte/data/base_reader.py
@@ -18,7 +18,7 @@ import logging
 import os
 from abc import abstractmethod, ABC
 from pathlib import Path
-from typing import Any, Iterator, Optional, Union, List
+from typing import Any, Iterator, Optional, Union, List, Dict, Set
 
 from forte.common.configuration import Config
 from forte.common.exception import ProcessExecutionException
@@ -209,6 +209,7 @@ class BaseReader(PipelineComponent[PackType], ABC):
                     yield pack
 
     def iter(self, *args, **kwargs) -> Iterator[PackType]:
+        # pylint: disable=protected-access
         r"""An iterator over the entire dataset, giving all Packs processed
         as list or Iterator depending on `lazy`, giving all the Packs read
         from the data source(s). If not reading from cache, should call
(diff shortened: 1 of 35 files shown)


forte/data/base_reader.py
```diff
@@ -208,6 +214,31 @@ def _lazy_iter(self, *args, **kwargs):
                     pack.add_all_remaining_entries()
                     yield pack
 
+    def set_profiling(self, enable_profiling: bool = True):
+        r""" Set profiling option.
+        Args:
+            enable_profiling: A boolean of whether to enable profiling
+                for the reader or not.
```
hunterhector, line 221: default value in the docstring.

forte/pipeline.py
```diff
@@ -121,6 +122,10 @@ def __init__(self, resource: Optional[Resources] = None):
         self.initialized: bool = False
         self._check_type_consistency: bool = False
 
+        # provide an option for the pipeline to start the profiling
```
hunterhector, line 125: guess we should document clearly that this is for time profiling.

tests/forte/processors/timeprofiling_test.py
```diff
@@ -0,0 +1,72 @@
...
+            self.pl.finish()
+
+        self.assertEqual(len(log.output), 1)
+        lines: List[str] = log.output[0].split('\n')
+        self.assertEqual(len(lines), 5)
+        self.assertEqual('INFO:forte.pipeline:Pipeline Time Profile', lines[0])
+        self.assertIn('- Reader: forte.data.readers.string_reader.StringReader, ', lines[1])
+        self.assertIn('- Component [0]: __main__.DummyPackProcessor, 0.90', lines[2])
+        self.assertIn('- Component [1]: __main__.DummyPackProcessor, 0.50', lines[3])
+        self.assertIn('- Component [2]: __main__.DummyPackProcessor, 1.20', lines[4])
```
hunterhector, line 69: I guess we cannot really assert the exact time to be the same as sleep time. There might be some unexpected delays. We can probably only assert that we have some non-zero time profiled for these processors.

hunterhector (commented):
only a few small comments

hunterhector (approved):

PR #437, opened 2021-04-22
Add record subclass type checking.

### LINKED ISSUE #436: Provide subclass type checking
When the downstream processor requires a type and and its attributes, if the upstream processor provide the subclass type of this type, it would also be accepted. For example, when the processor requires a `Token` type but the upstream provides `Word` type in the datapack, it would also be accepted if `Token` is the parent entry of `Word`.

Using ontology specification file as part of the resource to be passed in the pipeline in the constructor. Construct the entry tree that represent the parent-children relationship of these entry nodes. When doing the type checking, firstly acquire all the types from the input datapack and collect all the parent nodes of these types and add to the dictionary of types and attributes for comparsion. Check if all the expected types and their attributes could be found in the dictionary.

jennyzhang-petuum:
This PR fixes #436. 

### Description of changes
Using ontology specification file as part of the resource to be passed in the pipeline in the constructor. Construct the entry tree that represent the parent-children relationship of these entry nodes. When doing the type checking, firstly acquire all the types from the input datapack and collect all the parent nodes of these types and add to the dictionary of types and attributes for comparsion. Check if all the expected types and their attributes could be found in the dictionary.

### Test Conducted
Tests are added in the pipeline_test.py. It's checking for a ontology specification file that has imports like `forte/tests/forte/data/ontology/test_specs/example_ontology.json` to be used in the pipeline. In this file, `"ft.onto.example_ontology.Word"` is a subclass type of `"ft.onto.example_import_ontology.Token": {"pos", "lemma"}`. So when the downstream processor `DummyEvaluatorThree` and `DummyPackProcessorThree` expected types are set to be
`"ft.onto.example_import_ontology.Token": {"pos", "lemma"}`, and upstream processor `DummySentenceReaderTwo` writes `"ft.onto.example_ontology.Word"` into the record, the pipeline should run fine and no error exception should be thrown out.


forte/processors/base/base_processor.py
```diff
@@ -58,8 +58,35 @@ def expected_types_and_attributes(cls) -> Dict[str, Set[str]]:
...
+        would be collected from this tree and add to the returned record
+        dictionary for later comparison to enable subclass type checking.
+
+        Args:
+            input_pack: The input datapack.
+
+        Returns:
+            input_pack_record: The input pack record content combined with
+            all the parent types and attributes collected from
+            merged_entry_tree
```
hunterhector, line 81: quote the variable name "merged_entry_tree" with back tip.

forte/processors/base/base_processor.py
```diff
@@ -74,7 +101,8 @@ def check_record(self, input_pack: PackType):
         """
         if self._check_type_consistency:
             expectation = self.expected_types_and_attributes()
-            record_types_and_attributes_check(expectation, input_pack)
+            input_pack_record = self.collect_input_pack_record(input_pack)
+            record_types_and_attributes_check(expectation, input_pack_record)
```
hunterhector, line 105: I think you can simply do it in line.

```suggestion
            record_types_and_attributes_check(self.expected_types_and_attributes(), self.collect_input_pack_record(input_pack))
```

forte/processors/third_party/allennlp_processors.py
```diff
@@ -101,10 +101,12 @@ def default_configs(cls):
         """
         config = super().default_configs()
         config.update({
-            'processors': "tokenize,pos,depparse",
+            'processors': "tokenize, pos, depparse",
```
hunterhector, line 104: will adding the space before each step create problem?

forte/processors/third_party/allennlp_processors.py
```diff
@@ -101,10 +101,12 @@ def default_configs(cls):
...
             'tag_formalism': "stanford",
             'overwrite_entries': False,
-            'allow_parallel_entries': True
+            'allow_parallel_entries': True,
+            'onto_file': 'forte/ontology_specs/allennlp.json'
```
hunterhector, line 108: let's don't pass ontology to each single processor

forte/processors/third_party/allennlp_processors.py
```diff
@@ -192,6 +195,8 @@ def expected_types_and_attributes(cls) -> Dict[str, Set[str]]:
         :meth:`~forte.pipeline.Pipeline.enforce_consistency` was enabled for
         the pipeline.
         """
-        expectation_dict: Dict[str, Set[str]] = dict()
-        expectation_dict["ft.onto.base_ontology.Sentence"] = set()
+        expectation_dict: Dict[str, Set[str]] = {
+             "ft.onto.base_ontology.Sentence": set()
+        }
+
```
hunterhector, line 201: I think you can directly return this inline.
```suggestion
return {
             "ft.onto.base_ontology.Sentence": set()
        }
```

forte/utils/utils_processor.py
```diff
@@ -86,15 +85,17 @@ def parse_allennlp_srl_tags(tags: str) -> \
...
     match with input_pack.record. If not, an error of
     :class:`~forte.common.exception.ExpectedRecordNotFound` will be raised.
 
     Args:
         expectation: Dictionary of types and their attributes required for
             the current processor/evaluator.
-        input_pack: The input pack.
+        input_pack_record: The input pack record content combined with
+            all the parent types and attributes collected from
+            merged_entry_tree.
```
hunterhector, line 98: Quote the variable name.

forte/evaluation/base/base_evaluator.py
```diff
@@ -93,6 +94,32 @@ def ref_pack_record(self, record_meta: Dict[str, Set[str]]):
...
+        Args:
+            input_pack: The input datapack.
+
+        Returns:
+            input_pack_record: The input pack record content combined with
+            all the parent types and attributes collected from
+            merged_entry_tree
+
+        """
+        input_pack_record = input_pack._meta.record.copy()
```
hunterhector, line 117: Is this function used twice or more? We should create a small util function and reference it, don't copy the code.

hunterhector (commented):
Actually, let's don't use spacy and allennlp ontology for our test purpose? Because they are going to be moved to the 3rd party repo.

For other comments, see them inline.

forte/data/ontology/code_generation_objects.py
```diff
@@ -691,8 +691,8 @@ def print_traverse(self):
         traverse(self.root, path)
 
     def collect_parents(self, node_dict: Dict[str, Set[str]]):
-        r"""Collect all the parent nodes for all the nodes in the node_dict
-        and add the types and attributes of these parent nodes to node_dict.
+        r"""Collect all the parent nodes for all the nodes in the `node_dict`
+        and add the types and attributes of these parent nodes to `ode_dict`.
```
hunterhector, line 695:
```suggestion
        and add the types and attributes of these parent nodes to `node_dict`.
```

forte/pipeline.py
```diff
@@ -122,6 +125,7 @@ def __init__(self, resource: Optional[Resources] = None):
             self.resource = Resources()
         else:
             self.resource = resource
+        self.resource.update(onto_specs=ontology_file)
```
hunterhector, line 160: So we are planning to have the ontology content in the pipeline for other purposes too. I think it is better to load the content of the `ontology_file` because the file path may not be accessible in all scenario (for example, in a distributed machine setting).

forte/data/ontology/ontology_code_generator.py
```diff
@@ -578,9 +583,49 @@ def parse_ontology_spec(
...
+            spec_dict = json.load(f)
+
+        # Assume no import key in the ontology file
+        if SchemaKeywords.imports in spec_dict.keys():
+            raise OntologySpecValidationError(
+                "the input ontology specification file should not have import"
+                "as key, i.e. if there is an existing ontology file that need"
+                "to be imported, you should run generate_ontology create with"
+                "-merged_path specified and use this generated file as ontology"
+                "file")
```
hunterhector, line 617: Maybe hard for users to understanding this message.
```
raise OntologySpecValidationError(
    "The system cannot build entry trees from an ontology file that imports"
    " other ontologies. But we find `import` keywords in the provided ontology "
    f" {ontology_path}" 
    " Please provide a merged ontology file with all imports resolved"
    " by using the command `generate_ontlogy create --merged_path`"
)
```

forte/data/ontology/code_generation_objects.py
```diff
@@ -634,3 +634,110 @@ def get(self, module_name: str) -> ModuleWriter:
...
+
+
+class EntryTreeNode:
+    def __init__(self, name: str):
+        self.children: List[EntryTreeNode] = []
+        self.parent: Optional[EntryTreeNode] = None
+        self.name: str = name
+        self.attributes: Set[str] = set()
+
+    def attr_string(self):
```
hunterhector, line 646: fyi, you can implement python's internal `__repr__` to simplify print.

forte/data/ontology/code_generation_objects.py
```diff
@@ -634,3 +634,110 @@ def get(self, module_name: str) -> ModuleWriter:
...
+        input_node_dict = node_dict.copy()
+        for node_name in input_node_dict.keys():
+            found_node = search(self.root, search_node_name=node_name)
+            if found_node is not None:
+                while found_node.parent.name != 'root':
+                    node_dict[found_node.parent.name] = found_node.\
+                        parent.attributes
+                    found_node = found_node.parent
+
+    def get_all_children_below(self, node_name: str):
```
hunterhector, line 711: where is this function used?

hunterhector (commented):
some small comments.

forte/data/ontology/code_generation_objects.py
jennyzhang-petuum: It's not used anywhere. Just in case it's needed later.

hunterhector (approved):
LGTM

Issue #453, opened 2021-05-20
Removing forte-wrapper processors from the Forte repo.


jzpang:
This issue is the same as https://github.com/asyml/forte-wrappers/issues/16. One is the remove in Forte, the other is the add in forte-wrapper repo.

The following is copied from https://github.com/asyml/forte-wrappers/issues/16.
Is your feature request related to a problem? Please describe.
Once the processors in this repository has been set up and tested, we should start removing the processors copy in the original repo.

The thrid_party and ir in https://github.com/asyml/forte/tree/master/forte/processors, that can be found in this repo. Should be removed.

Describe the solution you'd like

Remove the processors and test cases in Forte repo.
Fix examples referencing those in Forte repo, and use the forte-wrapper version
Describe alternatives you've considered

Additional context

PR #454, opened 2021-05-20
Remove processors in forte-wrappers from Forte

jzpang:
This PR partially fixes https://github.com/asyml/forte/issues/453. 

### Description of changes
Removed some third party processors to forte-wrapper repo, and corresponding test cases. Imports were also changed. 

### Test Conducted


jzpang:
NLTK prcessor was not removed in current version, the path is: 
forte/processors/third_party/nltk_processors.py

Due to the effect on the following files:
tests/forte/processors/base/data_augment_replacement_processor_test.py
tests/forte/processors/data_augment/algorithms/eda_processors_test.py
tests/forte/processors/nltk_processors_test.py
tests/forte/processors/pretrained_encoder_processors_test.py
tests/forte/processors/writer_test.py

Also need to change the following import in future:
forte/processors/third_party/__init__.py
examples/data_augmentation/data_select/data_select_and_augment_example.py
examples/biobert_ner/ner_predict_example.py
examples/chatbot/chatbot_example.py
examples/pipelines/process_dataset_example.py
examples/pipelines/process_string_example.py
examples/serialization/serialize_example.py


ElasticSearchIndexProcessor was not removed in current version, the path is:
forte/processors/ir/elastic_search_index_processor.py

Due to the effect on the following files:
forte/processors/data_augment/selector_index_processor.py


TextGenerationProcessor was not removed in current version, the path is:
forte/processors/third_party/text_generation_processor.py

Due to the effect on the following files(will cause test case failure):
forte/tests/examples/gpt2_test.py

Also need to change the following files in future:
forte/examples/gpt2/multipack_pipeline_gpt2.py
forte/examples/gpt2/sample_multipack_pipeline_gpt.yml

hunterhector (commented):
Thanks for removing the code and also update the examples. 

One thing is this causes two test coverage decrease:
1. https://app.codecov.io/gh/asyml/forte/compare/454/changes#D1L327
Here we probably have a test case the throws a `ProcessorConfigError`. Can we add a small test case back to cover this part?

2. https://app.codecov.io/gh/asyml/forte/compare/454/changes#D4L46
This part is because we shouldn't have used any allennlp results in the main repo now. So we should check whether we can remove these code. There are two functions about allennlp in this file, can you check if we can remove both of them?

Btw, thanks for listing out all the processors that we plan to remove, can you open some issues to track them? For each of them we can decide a strategy to remove them, and each of them can be a separate issue.

jzpang:
> Thanks for removing the code and also update the examples.
> 
> One thing is this causes two test coverage decrease:
> 
> 1. https://app.codecov.io/gh/asyml/forte/compare/454/changes#D1L327
>    Here we probably have a test case the throws a `ProcessorConfigError`. Can we add a small test case back to cover this part?

I have checked the previous test cases, these lines were covered by Allennlp and spacy processor. For both of them, they have a config check in initialize function, for example, https://github.com/asyml/forte/blob/master/forte/processors/third_party/allennlp_processors.py#L53.

I checked several processors that have configs in current forte repo, and they don't have similar checks. So even if we wrote similar test cases, the same error could not be raised. So, if we want to cover these lines, we need to add similar checks in the current forte processors, which means to change the current forte processors. Do we want to do this at the current step?

> 2. https://app.codecov.io/gh/asyml/forte/compare/454/changes#D4L46
>    This part is because we shouldn't have used any allennlp results in the main repo now. So we should check whether we can remove these code. There are two functions about allennlp in this file, can you check if we can remove both of them?
> 
Yes I will removed them.

> Btw, thanks for listing out all the processors that we plan to remove, can you open some issues to track them? For each of them we can decide a strategy to remove them, and each of them can be a separate issue.

hunterhector:
> I checked several processors that have configs in current forte repo, and they don't have similar checks. So even if we wrote similar test cases, the same error could not be raised. So, if we want to cover these lines, we need to add similar checks in the current forte processors, which means to change the current forte processors. Do we want to do this at the current step?

So we now learn that it is a bad practice to test framework functions using specific implementations. In many of our other test cases, we create "dummy processors" that help us check the behavior, example like below:

https://github.com/asyml/forte/blob/master/tests/forte/pipeline_test.py#L255

jzpang:
> > I checked several processors that have configs in current forte repo, and they don't have similar checks. So even if we wrote similar test cases, the same error could not be raised. So, if we want to cover these lines, we need to add similar checks in the current forte processors, which means to change the current forte processors. Do we want to do this at the current step?
> 
> So we now learn that it is a bad practice to test framework functions using specific implementations. In many of our other test cases, we create "dummy processors" that help us check the behavior, example like below:
> 
> https://github.com/asyml/forte/blob/master/tests/forte/pipeline_test.py#L255

Thanks for pointing this file. I added a new test for config check, it now can cover those lines.

hunterhector (approved):

PR #459, opened 2021-05-24
add attribute_name checking and new attribute for Sentence

### LINKED ISSUE #458: Add attribute_name checking and new attribute `classification` for `Sentence` in base_ontology
For the current `_check_attr_type(self, key, value)` method in `Entry` class, key was not checked if exists in the current 
class before adding new attributes with the key specified for that class. 
When predicting/annotating classification result for `Sentense` entry type, we don't have a field to add the result to. 

Add attribute_name checking to `_check_attr_type(self, key, value)` method in `Entry` class.
Add new attribute `classification` for `Sentence` in base_ontology.

jennyzhang-petuum:
This PR fixes #458  . 

### Description of changes
add attribute_name checking and new attribute for Sentence
Added for the `Entry` class to check if the key exists before entering the respective attributes to the dict.
### Test Conducted
No new test cases added


forte/data/ontology/core.py
```diff
@@ -221,6 +221,11 @@ def _check_attr_type(self, key, value):
         """
         if key not in default_entry_fields:
             hints = get_type_hints(self.__class__)
+            if key not in hints.keys():
+                raise KeyError(
```
hunterhector, line 225: I actually suggest converting this to a warning. I realize that a type hint may not necessarily present. Could you also change the line below `TypeError` to a warning as well? Here, both the warnings can be "based on type annotation, blah blah."

There will be runtime exceptions being thrown when the insertion is uncessful. Btw, can you add test cases for these, to show that warnings and exceptions are thrown as expected?

hunterhector (approved):
Thanks!

hunterhector (approved):

PR #472, opened 2021-06-09
Creates ftx package for additional packages

hunterhector:
This PR introduces a new package naming convention `ftx` to include additional ontology classes. `ftx` stands for `forte ontology extra`


hunterhector:
solved together with https://github.com/asyml/forte/pull/471

PR #476, opened 2021-06-14
Add RemoteProcessor Service

mylibrar:
This PR fixes #475. 

### Description of changes
- Add Forte service in **`pipeline.py`**:
    - A Forte pipeline can be serialized into an intermediate representation that is dumped to a YAML file by calling `Pipeline().save()`.
    - Each pipeline has a FastAPI app that can be used to open an end point for remote service by calling `Pipeline().serve()`.
    - Forte can directly load an intermediate representation from file path to initialize a pipeline and create a remote service by calling `forte.serve()`.
- Add a `RemoteProcessor`:
    - This will use the service created by Forte pipeline on a specified end point.
    - It wraps up the input `DataPack` to a POST request and update the input according to the response from service.
    - The service pipeline must have a `RawDataDeserializeReader`.
### Possible influences of this PR.
For `pylint` checking, `Pipeline` class has reached the maximum number of public methods (20/20)

### Test Conducted
A unit test `remote_processor_test.py` is provided to verify the intermediate representation and RemoteProcessor.

PR #477, opened 2021-06-18
Add service validation

mylibrar:
This PR is an update to #476. 

### Description of changes
- pipeline.py
    - User can assign a name to pipeline service by calling `Pipeline().serve(name="name")` or `forte.serve(pl_path="pl_path", name="name")` 
    - Add a new page `\records` in FastAPI app which will return a collection of records from all components in the pipeline.
- remote_processor.py
    - User can now configure `RemoteProcessor` to validate the remote service by adding `do_validation`, `expected_name` and `expected_records` field.
```python
pipeline.add(
    RemoteProcessor,
    config={
        "do_validation": True,
        "expected_name": service_name,
        "expected_records": json.dumps(
            {"Token": ["1", "2"], "Document": ["2"]}
        ),
    },
)
```
- remote_processor_test.py
    - Add test case for service name and records checking. 

### Test Conducted
`remote_processor_test.py` is updated to test validation.

### DIFF
--- a/forte/pipeline.py
+++ b/forte/pipeline.py
@@ -469,7 +469,7 @@ class Pipeline(Generic[PackType]):
 
         # Check record types and attributes of each pipeline component
         if self._do_init_type_check:
-            current_records = {}
+            current_records: Dict[str, Set[str]] = {}
             self._reader.record(current_records)
             for component in self.components:
                 record_types_and_attributes_check(
--- a/forte/pipeline.py
+++ b/forte/pipeline.py
@@ -57,6 +57,7 @@ from forte.process_manager import ProcessManager, ProcessJobStatus
 from forte.processors.base import BaseProcessor
 from forte.processors.base.batch_processor import BaseBatchProcessor
 from forte.utils import create_class_with_kwargs
+from forte.utils.utils_processor import record_types_and_attributes_check
 
 logger = logging.getLogger(__name__)
 
@@ -115,6 +116,7 @@ class Pipeline(Generic[PackType]):
         resource: Optional[Resources] = None,
         ontology_file: Optional[str] = None,
         enforce_consistency: bool = False,
+        do_init_type_check: bool = False,
     ):
         r"""
 
@@ -146,6 +148,12 @@ class Pipeline(Generic[PackType]):
                 `enforce_consistency=True`, processor A would check if this
                 type exists in the record of the output of the
                 previous pipeline component.
+            do_init_type_check: Determine whether to check records types and
+                attributes during pipeline initialization. Default to `False`.
+                If this boolean is set to `True`, each component in the
+                pipeline will be validated by comparing its
+                ``expected_types_and_attributes`` with the accumulated
+                ``records`` from all the downstream components.
         """
         self._reader: BaseReader
         self._reader_config: Optional[Config] = None
@@ -193,6 +201,9 @@ class Pipeline(Generic[PackType]):
         self._profiler: List[float] = []
 
         self._check_type_consistency = enforce_consistency
+
+        # Indicate whether do type checking during pipeline initialization
+        self._do_init_type_check: bool = do_init_type_check
 
     def enforce_consistency(self, enforce: bool = True):
         r"""This function determines whether the pipeline will check
@@ -338,6 +349,13 @@ class Pipeline(Generic[PackType]):
                     component.record(records)
             return {"status": "OK", "records": records}
 
+        @app.get("/expectation")
+        def get_expectation():
+            expectation: Dict[str, Set[str]] = {}
+            if len(self.components) > 0:
+                expectation = self.components[0].expected_types_and_attributes()
+            return {"status": "OK", "expectation": expectation}
+
         @app.post("/process")
         def run_pipeline(body: RequestBody):
             args = json.loads(body.args)
@@ -448,6 +466,16 @@ class Pipeline(Generic[PackType]):
         if self._enable_profiling:
             self.reader.set_profiling(True)
             self._profiler = [0.0] * len(self.components)
+
+        # Check record types and attributes of each pipeline component
+        if self._do_init_type_check:
+            current_records = {}
+            self._reader.record(current_records)
+            for component in self.components:
+                record_types_and_attributes_check(
+                    component.expected_types_and_attributes(), current_records
+                )
+                component.record(current_records)
 
         return self
 
--- a/forte/pipeline.py
+++ b/forte/pipeline.py
@@ -289,17 +289,21 @@ class Pipeline(Generic[PackType]):
         with open(path, "w") as f:
             yaml.safe_dump(self._dump_to_config(), f)
 
-    @property
-    def _remote_service_app(self):
+    def _remote_service_app(self, name: str = ""):
         r"""Return a FastAPI app that can be used to serve the pipeline.
         Currently it only supports the `process` function, but it can be
         extended by adding new interfaces that wrap up any Pipeline method.
         Refer to https://fastapi.tiangolo.com for more info.
 
+        Args:
+            name: Assign a name to the pipeline service for validation.
+                Default to `''`.
+
         Returns:
             FastAPI: A FastAPI app for remote service.
         """
         app = FastAPI()
+        records: str = ""
 
         class RequestBody(BaseModel):
             args: str = "[]"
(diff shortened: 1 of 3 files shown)


tests/forte/processors/remote_processor_test.py
```diff
@@ -111,12 +119,23 @@ def test_remote_processor(self, input_output_pair):
...
+        remote_processor.set_test_mode(
+            serve_pl._remote_service_app(name=service_name)
+        )
 
         # Build test pipeline
         test_pl: Pipeline[DataPack] = Pipeline[DataPack]()
         test_pl.set_reader(StringReader())
-        test_pl.add(remote_processor)
+        test_pl.add(
+            remote_processor,
```
hunterhector, line 130: I just realize here the remote service can only function correctly if we use a reader that can take a data pack (RawDataDeserializeReader in this case). If our remote processor is configured so that the reader is something else like `StringReader`, then it may interpret the input wrong. This is something we could check during validation. Because we are building a `DataPack` pipeline, then it should check whether the remote service supports `DataPack`. We can do the following:

1. During `serve()`, we can have a configuration `input_format="string"`, which means the input format defaults to `string` (or other default values you find appropriate). For the remote service to support "DataPack", we need to set the `input_format` to be `DataPack` (in the future we may support `MultiPack` too).  For the time being, we could support these two values. However, we do not have a way for the `Reader` to check this input. This could be a potential feature that's useful, we might do it someday.
2. During `add(remote_processor)`, we should check whether the output from the previous step matches the remote processor configuration. Unfortunately, in Python, we cannot reliably find the output type of the previous PipelineComponent, hence we simply add this check as part of the validation configuration.

We should also group the validation settings into the same group, like the following:
```
config={               
   "validation": {
       "do_validation": True,                
      "input_format": "DataPack",
      "expected_name": service_name,                
       "expected_records": {
            "Token": ["1", "2"], 
            "Document": ["2"]}  
       },
   }
```

tests/forte/processors/remote_processor_test.py
```diff
@@ -111,12 +119,23 @@ def test_remote_processor(self, input_output_pair):
...
+            config={
+                "do_validation": True,
+                "expected_name": service_name,
+                "expected_records": json.dumps(
```
hunterhector, line 134: why there is a `json.dumps` here? this is a user-oriented interface, we should avoid these extra operations to be done by the caller.

forte/processors/misc/remote_processor.py
```diff
@@ -67,14 +68,40 @@ def initialize(self, resources: Resources, configs: Config):
                 f"{response.status_code}: "
                 "Remote service not started or invalid endpoint configs."
             )
+        service_name: str = response.json()["name"]
+
+        if self.configs.do_validation:
```
hunterhector, line 73: After the discussion with Jenny, it seems it is a better idea to implement this validation at the processor level. We can keep this validate service name part here since this is special.

The config name can also be changed to something more meaningful, instead of `do_validation`, it can be `do_init_type_check`

forte/processors/misc/remote_processor.py
```diff
@@ -67,14 +68,40 @@ def initialize(self, resources: Resources, configs: Config):
...
+            # Validate service name
+            if service_name != self.configs.expected_name:
+                raise ProcessorConfigError(
+                    "Validation fail: Name Mismatch. "
```
hunterhector, line 77: The error message is unspecific. Here is an great article about this (and in general how to write good APIs): https://blog.keras.io/category/essays.html

forte/processors/misc/remote_processor.py
```diff
@@ -67,14 +68,40 @@ def initialize(self, resources: Resources, configs: Config):
...
+                    f"'{service_name}' != '{self.configs.expected_name}'"
+                )
+
+            # Validate the output records
+            response = self._requests.get(f"{self.configs.url}/records")
+            if response.status_code != 200 or response.json()["status"] != "OK":
+                raise ProcessorConfigError(
+                    f"{response.status_code}: "
+                    "Fail to fetch records from remote service."
```
hunterhector, line 86: Similar as above, let's write good error messages.

forte/pipeline.py
```diff
@@ -1084,7 +1116,9 @@ def serve(pl_config_path: str, host: str = "localhost", port: int = 8008):
             pipeline.
         host: Port number of pipeline service.
         port: Host name of pipeline service.
+        name: Assign a name to the pipeline service for validation.
+            Default to `''`.
     """
     pipeline: Pipeline = Pipeline()
     pipeline.init_from_config_path(pl_config_path)
-    pipeline.serve(host=host, port=port)
+    pipeline.serve(host=host, port=port, name=name)
```
hunterhector, line 1124: Probably `service_name` instead of `name`

forte/pipeline.py
```diff
@@ -1084,7 +1116,9 @@ def serve(pl_config_path: str, host: str = "localhost", port: int = 8008):
             pipeline.
         host: Port number of pipeline service.
         port: Host name of pipeline service.
+        name: Assign a name to the pipeline service for validation.
```
hunterhector, line 1119: Explain how this name will be used.

forte/pipeline.py
```diff
@@ -308,30 +312,53 @@ class RequestBody(BaseModel):
...
+            name: Assign a name to the pipeline service for validation.
+                Default to `''`.
         """
         self.initialize()
         uvicorn.run(
-            self._remote_service_app, host=host, port=port, log_level="info"
+            self._remote_service_app(name=name),
+            host=host,
+            port=port,
+            log_level="info",
```
hunterhector, line 361: I observed `uvicorn` duplicate every error messages, maybe the logging setup is not correct?

```
INFO:     Started server process [14115]
INFO:uvicorn.error:Started server process [14115]
INFO:     Waiting for application startup.
INFO:uvicorn.error:Waiting for application startup.
INFO:     Application startup complete.
INFO:uvicorn.error:Application startup complete.
INFO:     Uvicorn running on http://localhost:8008 (Press CTRL+C to quit)
INFO:uvicorn.error:Uvicorn running on http://localhost:8008 (Press CTRL+C to quit)
```

hunterhector (commented):
See comments inline.

forte/processors/misc/remote_processor.py
```diff
@@ -56,16 +57,73 @@ class RemoteProcessor(PackProcessor):
...
+        if self._validation.do_init_type_check:
+            # Validate service name and input format
+            if service_name != self._validation.expected_name:
+                raise ProcessorConfigError(
+                    "Validation fail: The expected service name "
+                    f"('{self._validation.expected_name}') does not match "
+                    "the actual name returned by remote service "
+                    f"('{service_name}'). Consider updating the configs of "
+                    "RemoteProcessor so that 'validation.expected_name' "
+                    f"equals to '{service_name}'."
```
hunterhector, line 88: The user may actuall calls the wrong service, can ask them to double check.

tests/forte/processors/remote_processor_test.py
```diff
@@ -111,12 +120,29 @@ def test_remote_processor(self, input_output_pair):
...
+            remote_processor,
+            config={
+                "validation": {
+                    "do_init_type_check": True,
+                    "input_format": input_format,
+                    "expected_name": service_name,
+                    "expected_records": {
+                        "Token": {"1", "2"},
+                        "Document": {"2"},
+                    },
```
hunterhector, line 142: Did this work now? I thought we cannot have unknown names.

forte/processors/misc/remote_processor.py
```diff
@@ -56,16 +57,73 @@ class RemoteProcessor(PackProcessor):
...
+            if response.status_code != 200 or response.json()["status"] != "OK":
+                raise ProcessorConfigError(
+                    f"{response.status_code} {response.reason}: "
+                    "Fail to fetch records from remote service. Please make "
+                    f"sure that the remote service at {self.configs.url} is "
+                    "a valid pipeline service that is up and running."
+                )
+            self._records = response.json()["records"]
+
+    def check_record(self, input_pack):
```
hunterhector, line 111: We can put `check_record` to base processor

forte/processors/misc/remote_processor.py
```diff
@@ -56,16 +57,73 @@ class RemoteProcessor(PackProcessor):
...
+        """
+        super().check_record(input_pack)
+        if (
+            self._validation.do_init_type_check
+            and self._validation.expected_records is not None
+        ):
+            # Validate the output records
+            record_types_and_attributes_check(
+                self._validation.expected_records.todict(), self._records
             )
```
hunterhector, line 127: `do_init_type_check` should be moved to the base too (along with the configurations)

forte/pipeline.py
```diff
@@ -401,6 +467,21 @@ def initialize(self) -> "Pipeline":
...
             self._profiler = [0.0] * len(self.components)
 
+        # Check record types and attributes of each pipeline component
+        if self._do_init_type_check:
+            current_records: Dict[str, Set[str]] = {}
+            self._reader.record(current_records)
+            for component in self.components:
+                if not (
+                    hasattr(component, "expected_types_and_attributes")
+                    and hasattr(component, "record")
```
hunterhector, line 477: Can we do these two `if` conditions separately? That means, run the two statements for each different condition.

hunterhector (dismissed):
now looks good.

hunterhector (commented):
only a small comment left

hunterhector (approved):

Issue #479, opened 2021-06-22
Add BIO tagging merge function


jzpang:
**Is your feature request related to a problem? Please describe.**
We need a function to do BIO tagging entity merge.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #480, opened 2021-06-22
add a bio tagging merge function

jzpang:
This PR fixes https://github.com/asyml/forte/issues/479. 

### Description of changes
add a bio tagging merge function.

### Test Conducted

### DIFF
--- a/forte/utils/tagging_scheme.py
+++ b/forte/utils/tagging_scheme.py
@@ -15,57 +15,59 @@
 Utility functions related to tagging scheme.
 """
 
-import warnings
-from typing import Optional, List, Union, Tuple, Any
+import logging
+from typing import Optional, List, Union, Tuple
 
 
 def bio_merge(tags: List[str], types: List[Union[str, None]],
-              start: Optional[List[int]]=None, end: Optional[List[int]]=None) \
-        -> Tuple[List[Any], List[Any], List[Any]]:
+              index: Optional[List[Tuple[int, int]]]=None) \
+        -> Tuple[List[Union[str, None]],
+                 List[Tuple[Union[int, None], Union[int, None]]]]:
     r""" This function merged BIO-schemed augmented tagging scheme results and
-    return entity mentions information.
+    return chunks information.
 
     For example, BIO NER tags could be merged by passing
     tags = ['B', 'O', 'B', 'I']
     types = ['PER', '', 'LOC', 'LOC']
-    start = [0, 11, 20, 24]
-    end = [1, 19, 22, 26]
+    index = [(0, 1), (11, 19), (20, 22), (24, 26)]
 
-    After merging BIO tags, the result will be returned as
+    After merging BIO tags, the results will be returned as
     result_types = ['PER', 'LOC']
-    result_start = [0, 20]
-    result_end = [1, 26]
+    result_index = [(0, 1), (20, 26)]
+
+    The function handles 'I' with no leading 'B' tag. If we encounter
+    "I" while its type is different from the previous type, we will consider
+    this "I" as a "B" and start a new record here.
 
     The function can also handle tags with no types, for example, in some word
     segmentation tasks. In this case the input `types` should be set as a list
     of None, and the returned result_type will be a list of None.
 
-
     Args:
         tags: list of bio tags, contains "B", "I", "O" labels.
         types: list of entity type, could be PER, LOC in NER task.
-        start: list of start index for each input tag. default is None.
-        end: list of end index for each input tag. default is None.
+        index: list of (start, end) index for each input tag. default is None.
 
     Returns:
         result_types: list of merged entity type.
-        result_start: list of start index for the merged entities.
-        result_start: list of end index for the merged entities.
+        result_index: list of (start, end) index for the merged entities.
     """
-    prev_type = None
-    prev_tag = None
-    prev_start = None
-    prev_end = None
-    new_entity = False
-    is_indexed = True
+    prev_type: Optional[str] = None
+    prev_tag: Optional[str] = None
+    prev_start: Optional[int] = None
+    prev_end: Optional[int] = None
+    new_entity: bool = False
+    is_indexed: bool = True
 
     # No start or end information is provided, do not process index information
-    if start is None or end is None:
+    if index is None:
         is_indexed = False
         start = []
         end = []
-        warnings.warn('start and end indexes for the tags was not provided '
+        logging.warning('start and end indexes for the tags was not provided '
                       'and will be returned as `None`')
+    else:  # get start and end index
+        start, end = zip(*index)
 
     # input check
     if len(tags) != len(types) or (is_indexed and (len(start) != len(tags) or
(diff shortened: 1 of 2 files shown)


forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+
+import warnings
+from typing import Optional, List, Union, Tuple, Any
+
+
+def bio_merge(tags: List[str], types: List[Union[str, None]],
+              start: Optional[List[int]]=None, end: Optional[List[int]]=None) \
+        -> Tuple[List[Any], List[Any], List[Any]]:
+    r""" This function merged BIO-schemed augmented tagging scheme results and
+    return entity mentions information.
```
hunterhector, line 26: May not be `entity mention`, BIO can be applied to other domains

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Utility functions related to tagging scheme.
+"""
+
+import warnings
+from typing import Optional, List, Union, Tuple, Any
+
```
hunterhector, line 20: Fix the format with black https://github.com/psf/black#usage

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+def bio_merge(tags: List[str], types: List[Union[str, None]],
+              start: Optional[List[int]]=None, end: Optional[List[int]]=None) \
+        -> Tuple[List[Any], List[Any], List[Any]]:
+    r""" This function merged BIO-schemed augmented tagging scheme results and
+    return entity mentions information.
+
+    For example, BIO NER tags could be merged by passing
+    tags = ['B', 'O', 'B', 'I']
+    types = ['PER', '', 'LOC', 'LOC']
```
hunterhector, line 30: why don't we pass the tags and types together, i.e. ['B-PER', 'O']

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+    start = [0, 11, 20, 24]
+    end = [1, 19, 22, 26]
```
hunterhector, line 32: It might be also easier for users to add `start` and `end` together, like `[(0, 1), (11, 19)]`, you can unzip them inside.

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+
+    After merging BIO tags, the result will be returned as
```
hunterhector, line 34: `results`

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+    result_types = ['PER', 'LOC']
+    result_start = [0, 20]
+    result_end = [1, 26]
```
hunterhector, line 37: Return the results as tuples too. `[(0, 1), (20, 26)]`

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+        types: list of entity type, could be PER, LOC in NER task.
+        start: list of start index for each input tag. default is None.
+        end: list of end index for each input tag. default is None.
+
+    Returns:
+        result_types: list of merged entity type.
+        result_start: list of start index for the merged entities.
+        result_start: list of end index for the merged entities.
+    """
+    prev_type = None
```
hunterhector, line 55: Need type annotation for variables

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Utility functions related to tagging scheme.
+"""
+
+import warnings
```
hunterhector, line 18: Any preference to use `warnings` instead of `logging.warn`

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+from typing import Optional, List, Union, Tuple, Any
+
+
+def bio_merge(tags: List[str], types: List[Union[str, None]],
+              start: Optional[List[int]]=None, end: Optional[List[int]]=None) \
+        -> Tuple[List[Any], List[Any], List[Any]]:
```
hunterhector, line 24: The rerturn type `Any` is not very helpful

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,115 @@
...
+    After merging BIO tags, the result will be returned as
+    result_types = ['PER', 'LOC']
+    result_start = [0, 20]
+    result_end = [1, 26]
+
+    The function can also handle tags with no types, for example, in some word
+    segmentation tasks. In this case the input `types` should be set as a list
+    of None, and the returned result_type will be a list of None.
+
+
```
hunterhector, line 43: Looks like this function handles incorrect tag sequences, can you document how will they be handled?

forte/utils/tagging_scheme.py
jzpang: Because some tasks, like ws, also used bio tagging but no NE with it. To make the function more general, it was split here.

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+import logging
+from typing import Optional, List, Union, Tuple
+
+
+def bio_merge(
+    tags: List[str],
+    types: List[Union[str, None]],
+    index: Optional[List[Tuple[Union[int, None], Union[int, None]]]] = None,
```
hunterhector, line 25: I think you can use `Optional[int]` in place of `Union[int, None]`

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,134 @@
...
+    result_types = ['PER', 'LOC']
+    result_index = [(0, 1), (20, 26)]
+
+    The function handles 'I' with no leading 'B' tag. If we encounter
+    "I" while its type is different from the previous type, we will consider
+    this "I" as a "B" and start a new record here.
+
+    The function can also handle tags with no types, for example, in some word
+    segmentation tasks. In this case the input `types` should be set as a list
+    of None, and the returned result_type will be a list of None.
```
hunterhector, line 47: Let's not return a list of `None`, instead we can return a list or a None, i.e. `Optional[List[str]]`

hunterhector, line 47: `result_types` not `result_type`

forte/utils/tagging_scheme.py
hunterhector, line 25: `indices` not `index`

forte/utils/tagging_scheme.py
```diff
@@ -0,0 +1,134 @@
...
+import logging
+from typing import Optional, List, Union, Tuple
+
+
+def bio_merge(
+    tags: List[str],
+    types: List[Union[str, None]],
+    index: Optional[List[Tuple[Union[int, None], Union[int, None]]]] = None,
+) -> Tuple[
+    List[Union[str, None]], List[Tuple[Union[int, None], Union[int, None]]]
```
hunterhector, line 27: The second return element: `result_indices`, should not contain `None`. It should be `List[Tuple[int, int]]`

hunterhector (commented):
A couple of type suggestions. Maintaining the type correctness here actually improve the code quality here. Now mypy would not allow you to make `result_start` to be `List[int]` because it is not aware that it can not be `None`. This can be done easily with an `assert` or an exception: https://adamj.eu/tech/2021/05/17/python-type-hints-how-to-narrow-types-with-isinstance-assert-literal/

hunterhector (approved):
looks good!

PR #499, opened 2021-08-10
Add pipeline states to ir

### LINKED ISSUE #498: Add pipeline states to intermediate representation
[Pipeline._dump_to_config()](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L276) contains the configs for selectors and omits the pipeline states like `_check_type_consistency`, `_do_init_type_check`, `_enable_profiling` while the expected behaviors are:
- The configs of the processor should not contains information about the selectors, unless additional provided by the user explicitly. The Selectors should be stored separately.
- Other pipeline states should be included in the pipeline.

-  Most of the states passed from `__init__` should be included in the serialized form.
- The `selectors` should be included in the serialized form.

**Additional context**

mylibrar:
This PR fixes #498 . 

### Description of changes
- BaseProcessor
    - Remove the `"selector"` key in `default_configs()`.
- Pipeline
    - Add a new entry `"PIPELINE_STATES"` in IR to serialized the current state of pipeline.
        - `_initialized`
        - `_enable_profiling`
        - `_check_type_consistency`
        - `_do_init_type_check`
        - `resource.onto_specs_dict`
        - `resource.merged_entry_tree`

### Test Conducted
`remote_processor_test.py` is updated to check whether a pipeline can correctly update its states based on IR.


forte/pipeline.py
```diff
@@ -299,6 +321,33 @@ def _dump_to_config(self):
                     "configs": config.todict(),
                 }
             )
+
+        # Serialize current states of pipeline
+        configs.append(
```
hunterhector, line 326: So this design mixes the component types and pipeline states at the same level. While this implementation works, it may not be easy to maintain, especially when we add something else in the future. My suggested configuration could be (basically separated the two into two dicts):
{
    “components”: {
        “type”: “xxx.NLTKSegmenter”
         ...
    }, # We store the components here
    “states”: {
        “_initialized”: True
        ...
    } # We store the states here
}

hunterhector: As you also notice, this would affect some of our examples, so we can change them here:
https://github.com/asyml/forte-wrappers/blob/ac2d3715a4b86db44c025ef9155a63a1b979cf28/examples/gpt2/sample_multipack_pipeline_gpt.yml

Now it is good to start adding a version to the serialized pipeline, so that we know how to handle changes in the future

forte/pipeline.py
```diff
@@ -281,24 +319,63 @@ def _dump_to_config(self):
...
                     ),
                     "configs": config.todict(),
+                    "selector": {
+                        "type": ".".join(
+                            [selector.__module__, type(selector).__name__]
+                        ),
+                        # TODO: This presumes that class attributes' names are
+                        # the same as the paramaters' names passed to
+                        # selector's constructor, which may not be always true.
+                        "kwargs": selector.__dict__ or None,
```
hunterhector, line 353: Maybe replace `None` with `{}` since you will use`{}` in place of `None` any ways?

forte/pipeline.py
```diff
@@ -281,24 +319,63 @@ def _dump_to_config(self):
...
                 {
                     "type": ".".join(
                         [component.__module__, type(component).__name__]
                     ),
                     "configs": config.todict(),
+                    "selector": {
+                        "type": ".".join(
+                            [selector.__module__, type(selector).__name__]
+                        ),
+                        # TODO: This presumes that class attributes' names are
```
hunterhector, line 350: One alternative is we don't store `kwargs`, but instead assign attributes using `setattr`. But we might run into error during `__init__` (for example, if the selector has a `__init__` time kwargs check).

Actually, a more important problem is that the `selectors` are currently having arbitrary init parameters. Some of them may not be serializable.

hunterhector (commented):
Overall looks good to me. There are a few comments to fix.

But for the problem on selector serialization, we may need to have a better solution. The `selectors` are currently very flexible but hard to serialize, an alternative is to restrict its interface (for example, use HParam/Config and remove all __init__ kwargs). This may require another PR. Let's fix the rest of the comments and keep this PR on hold.

forte/pipeline.py
```diff
@@ -219,6 +219,9 @@ def __init__(
         # Indicate whether do type checking during pipeline initialization
         self._do_init_type_check: bool = do_init_type_check
 
+        # The version of intermediate representation format
+        self.FORTE_IR_VERSION: str = "0.0.1"
```
hunterhector, line 223: Maybe it is better to put this string in this file "https://github.com/asyml/forte/blob/master/forte/version.py", and then refer here.

hunterhector:
Probably also see if we can fix the pipeline format here?

https://github.com/asyml/forte-wrappers/blob/ac2d3715a4b86db44c025ef9155a63a1b979cf28/examples/gpt2/sample_multipack_pipeline_gpt.yml

forte/data/selector.py
```diff
@@ -39,7 +39,7 @@
 
 class Selector(Generic[InputPackType, OutputPackType]):
     def __init__(self, **kwargs):
-        pass
+        self._stored_kwargs = kwargs
```
hunterhector, line 42: Looks like now all the `Selector` implementation need to add a call to `__init__` and pass in the values. I think this would create confusion. I think there are two ways:

1. Use `locals` (looks hacky)
https://stackoverflow.com/questions/3136915/passing-all-arguments-of-a-function-to-another-function/3137022#3137022
We can call `locals` in the first line of the base `__init__`, which should probably save only the function parameters

2. Add a `todict` and `fromdict` function in the `Selector` base instead, and implement the functions in the following way:
   a. `todict`, put all values in `dict` to JSON
   b. `fromdict`, load all values from this.

We will allow the users to implement `todict` and `fromdict` in their own way. If they don't, we will use the default implementation.

forte/pipeline.py
```diff
@@ -281,24 +323,89 @@ def _dump_to_config(self):
         Returns:
             dict: A dictionary storing IR.
         """
-        configs: List[Dict] = []
-        configs.append(
+
+        def get_type(instance) -> str:
```
hunterhector, line 327: We have some functions like this here:

https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L35

mylibrar:
Blocked by #503

forte/data/selector.py
```diff
@@ -66,12 +68,32 @@ def select(self, pack: MultiPack) -> Iterator[DataPack]:
...
+            }
+        )
     """
 
-    def __init__(self, select_name: str):
-        super().__init__()
-        assert select_name is not None
-        self.select_name: str = select_name
+    def __init__(self, *args, **kwargs):
+        assert (len(args) == 0) ^ (len(kwargs) == 0)
```
hunterhector, line 85: hmm, I think raising valueerror is better than assert, and we can provide clear error message on this.

forte/data/selector.py
```diff
@@ -66,12 +68,32 @@ def select(self, pack: MultiPack) -> Iterator[DataPack]:
...
+        if args:
+            configs = {"select_name": args[0]}
+        else:
+            assert ("configs" in kwargs) or ("select_name" in kwargs)
```
hunterhector, line 89: similarly, no `assert`, use clear exceptions

forte/data/selector.py
```diff
@@ -37,9 +39,9 @@
 ]
 
 
-class Selector(Generic[InputPackType, OutputPackType]):
-    def __init__(self, **kwargs):
-        pass
+class Selector(Generic[InputPackType, OutputPackType], Configurable):
+    def __init__(self, configs: Optional[Union[Config, Dict[str, Any]]] = None):
```
hunterhector, line 43: This is not the correct way to use `Configurable`. What we should do is leave the empty `__init__`, and the configurable will handle the configs in `initialize` function.

The only thing we need to do is to call the `initialize` function in `pipeline.initialize` @mylibrar maybe you can add that?

forte/data/selector.py
```diff
@@ -66,12 +68,32 @@ def select(self, pack: MultiPack) -> Iterator[DataPack]:
...
+                "select_name": "foo"
+            }
+        )
     """
 
-    def __init__(self, select_name: str):
-        super().__init__()
-        assert select_name is not None
-        self.select_name: str = select_name
+    def __init__(self, *args, **kwargs):
```
hunterhector, line 84: To preserve backward compatibility is easy here, we simply need to load `self._select_name` from the args, and then in the `initialize` function, we overwrite `self._select_name`.

forte/data/selector.py
zhanyuanucb: agree

forte/data/selector.py
zhanyuanucb: agree!

forte/data/selector.py
zhanyuanucb: Could you clarify this more?
I didn't see an `initialize` method in `Configurable` and seems like right now `Configurable` is using its method `make_configs` to set up configurations.
I thought `Selector.initialize` was only relevant to some states of a selector, which is different than its configs?

hunterhector (approved):
I think these all look good now.

PR #516, opened 2021-09-02
Datasets reader for Classification Tasks

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/525. 

### Description of changes
1. Implemented `ClassificationDatasetReader` and one utility function for generating indices for subtext such as title, description while concatenating them. 
2. Implemented two examples to use `ClassificationDatasetReader` and a text classifier to make predictions 

It helps users to initialize dataset reader for classification tasks more easily.

### Test Conducted
mypy, black, pylint, flake8, pytest coverage


hunterhector:
Can you also fix the following PR meta content:
1. The title of this PR should be a good description
2. Change the `[issue_link]` to the real issue link

.gitignore
```diff
@@ -1,5 +1,9 @@
 # Created by https://www.gitignore.io/api/python
 
+## test file
+test.py
```
hunterhector, line 4: probably no need to add this in gitignore

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+from forte.data.base_reader import PackReader
+from ft.onto.base_ontology import *
+
+
+__all__ = ["ClassificationDatasetReader"]
+
+
+class ClassificationDatasetReader(PackReader):
+    r"""
+    A generic dataset reader class for classification datasets.
```
hunterhector, line 31: Let's explain in details the scope of this reader, how to use it, and provide an example in this class docstring.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.index2class = None
+        self.label2index = None
+
+    def set_up(self):
+        assert self.configs.index2class is not None, (
+            "User must set index2class to enable"
+            " the dataset reader encode labels correctly"
```
hunterhector, line 43: I feel this sentence may have grammar error.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+
+class ClassificationDatasetReader(PackReader):
+    r"""
+    A generic dataset reader class for classification datasets.
+
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.index2class = None
```
hunterhector, line 37: Use `_index2class` or `__index2class`

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+    def initialize(self, resources: Resources, configs: Config):
+        super().initialize(resources, configs)
+        self.set_up()
+
+    def _collect(self, csv_file: str) -> Iterator[Tuple[int, str]]:
+        with open(csv_file, encoding="utf-8") as f:
+            data = csv.reader(f, delimiter=",", quoting=csv.QUOTE_ALL)
+            next(data)
```
hunterhector, line 61: Do we always need to skip the first line?

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+    def _parse_pack(self, line_info: Tuple[int, str]) -> Iterator[DataPack]:
+        line_id, line = line_info
+        # content label title
+        pack = DataPack()
+
+        # subtext fields must follow the ontology names
+        data_fields = self.configs.data_fields
+        assert (
+            "label" in data_fields
+        ), "There must be label data field in reader config."
```
hunterhector, line 77: 1. This check should be done at `initialize`
2. Don't use `assert`, let's use `Exceptions`. Here we can use `ProcessorConfigError` defined in `forte.common`

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+        assert len(self.configs.data_fields) == len(line), (
+            "Data fields provided in config "
+            "is not aligned with the actual line info from dataset."
+            + str((data_fields, line))
```
hunterhector, line 81: I think this error message will not be clear, `data_fields` and `line` will be print together and hard to read in console.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+        with open(csv_file, encoding="utf-8") as f:
+            data = csv.reader(f, delimiter=",", quoting=csv.QUOTE_ALL)
+            next(data)
+            for line_id, line in enumerate(data):
+                yield line_id, line
+
+    def _cache_key_function(self, line_info: Tuple[int, str]) -> str:
+        return str(line_info[0])
+
+    def _parse_pack(self, line_info: Tuple[int, str]) -> Iterator[DataPack]:
```
hunterhector, line 68: The type annotation should be wrong, it should be something like `Tuple[int, List[str]]` (depends on what CSV reader would output).

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+            len(self.configs.subtext_fields) > 0
+        ), "There must be at least one subtext to reader to select from"
+
+        df_dict = OrderedDict()
+        for df, value in zip(data_fields, line):
+            df_dict[
+                df
+            ] = value  # in general, value can be subtext or other types of data
+
+        # it determines the order of concatenation
```
hunterhector, line 93: Such information should be known to the user, not just as inline comment.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+        subtext_fields = self.configs.subtext_fields
+        # get text and subtext indices
+        assert set(self.configs.subtext_fields).issubset(
+            set(self.configs.data_fields)
+        ), "subtext fields must be a subset of data fields"
```
hunterhector, line 98: Again, should be done in `initialize` and should not use assert

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+
+        text, subtext_indices = generate_text_n_subtext_indices(
+            subtext_fields, df_dict
+        )
+        pack.set_text(text, replace_func=self.text_replace_operation)
```
hunterhector, line 103: use `self.set_text` instead of this, so you don't need to call `text_replace_operation` yourself.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def generate_text_n_subtext_indices(subtext_fields, data_fields_dict):
+    """
+    Retrieve subtext from data fields and concatenate them into text.
+    Also, we generate the indices for these subtext accordingly.
+
+    :param subtext_fields: subtext names
+    :param data_fields_dict: a dictionary with subtext names as key and subtext string as value.
+    :return: a tuple of aggregated text and indices for subtext.
```
hunterhector, line 165: :param is not our sphinx doc standard.

hunterhector (commented):
Some comments, and you probably want to mark the PR as `work in progress` to avoid sending notifications to everyone.

hunterhector:
Remember to change the commit message `This PR fixes [issue link]` to include the link

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,227 @@
...
+            entry_class(pack, start_idx, end_idx)
+        # for now, we use document to store concatenated text and set the class here
+        doc = Document(pack, 0, subtext_indices[subtext_fields[-1]][1])
+        doc.document_class = [self.configs.index2class[class_id]]
+
+        pack.pack_name = line_id
+        yield pack
+
+    @classmethod
+    def default_configs(cls):
```
mylibrar, line 182: Maybe it's better to reorganize the inline comments below into a doctoring. You can refer to https://github.com/asyml/forte/blob/9597db37e21fb998f7e235bbbd66b93f1924e64b/forte/processors/base/writers.py#L126

examples/classification/bank_customer_intent.py
```diff
@@ -0,0 +1,138 @@
...
+    "index2class": index2class,
+    "subtext_fields": [
+        "ft.onto.ag_news.Description"
+    ],  # select subtexts to concatenate into text
+    "digit_label": False,  # specify whether label in dataset is digit
+    "text_label": True,  # either digit label or text label
+    "one_based_index_label": False,  # if it's digit label,
+    # whether it's one-based so that reader can adjust it
+}
+reader_config = Config(
```
mylibrar, line 111: You don't need to merge the `default_configs()` here. This step is taken care of by the pipeline, so all you need to do is `pl.set_reader(ClassificationDatasetReader(), config=this_reader_config)`. Same case for the `ZeroShotClassifier` below.

examples/classification/amazon_review_sentiment.py
```diff
@@ -0,0 +1,50 @@
...
+
+pl = Pipeline()
+
+# initialize labels
+class_names = ["negative", "positive"]
+index2class = dict(enumerate(class_names))
+
+# initialize reader config
+this_read_config = {"index2class": index2class}
+reader_config = Config(
```
mylibrar, line 22: Same here https://github.com/asyml/forte/pull/516/files#r705857866

forte/data/readers/classification_reader.py
feipenghe: Also mentioned it in default configs https://github.com/feipenghe/forte/blob/feaa1f3c6afcba2417e108caab29bdeb34388a72/forte/data/readers/classification_reader.py#L189

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def generate_text_n_subtext_indices(subtext_fields, data_fields_dict):
+    """
+    Retrieve subtexts from data fields and concatenate them into text.
+    Also, we generate the indices for these subtexts accordingly.
+
+    Args:
+        subtext_fields: a list of ontology that needs to be concatenated into a input string
+        data_fields_dict: a dictionary with subtext names as key and subtext string as value.
```
hunterhector, line 227: return value missing in docstring

hunterhector: The argument explanation is also quite hard to follow

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+
+    The user must specify the index2label mapping, data fields and subtext fields
+    and other parameters in configs based on their tasks in order to initialize
+    the dataset reader properly.
+    Index2label mapping is a dictionary with indices as keys and string labels as values,
+    it helps encoding and decoding string labels.
+    Data fields is a list of ontology paths representing data fields in the dataset.
+    They must have the same length and align with each other.
+    Subtext fields is also a list of ontology paths that represent data fields that
+     will be concatenated into input strings in the same order as the list.
```
hunterhector, line 44: I think for first time users, it is very hard for them to understand what data fields and subtext fields are by reading the documentation alone, can you try to explain with simple terms, more details, and with immediate examples?

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+                "Data fields provided in config "
+                "is not aligned with the actual line info from dataset.\n"
+                "Data fields length: "
+                + str(len(self.configs.data_fields))
+                + "\n",
+                "Line length: " + str(len(line)),
+            )
+
+        df_dict = OrderedDict()
+        for df, value in zip(data_fields, line):
```
hunterhector, line 135: I think you can create a dict simply by doing `dict(zip(data_fields, line))`

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+        # content label title
+        pack = DataPack()
+
+        # subtext fields must follow the ontology names
+        data_fields = self.configs.data_fields
+
+        if len(self.configs.data_fields) != len(line):
+            raise ProcessorConfigError(
+                "Data fields provided in config "
+                "is not aligned with the actual line info from dataset.\n"
```
hunterhector, line 127: so what is "actual line info", please use more user friendly terms. I feel like line info means the column format in the dataset.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+                "Data fields length: "
+                + str(len(self.configs.data_fields))
```
hunterhector, line 129: let's use fstring style instead of + concatenation

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+    They must have the same length and align with each other.
+    Subtext fields is also a list of ontology paths that represent data fields that
+     will be concatenated into input strings in the same order as the list.
+    User can select an arbitrary sequence of subtexts to concatenate
+     as long as there is at least one subtext in the sequence.
+     The number and the order of subtexts can be customized depending on the use cases.
+
+
+
+    The first line of dataset usually specifies column names of data fields,
```
hunterhector, line 51: explanation of dataset format should go first

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+    The user must specify the index2label mapping, data fields and subtext fields
+    and other parameters in configs based on their tasks in order to initialize
+    the dataset reader properly.
+    Index2label mapping is a dictionary with indices as keys and string labels as values,
+    it helps encoding and decoding string labels.
+    Data fields is a list of ontology paths representing data fields in the dataset.
+    They must have the same length and align with each other.
+    Subtext fields is also a list of ontology paths that represent data fields that
+     will be concatenated into input strings in the same order as the list.
+    User can select an arbitrary sequence of subtexts to concatenate
```
hunterhector, line 45: subtext is a term hard to understand

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+            - data_fields: these fields provides one-to-one correspondence
+                between given dataset fields and ontology paths.
+            - index2class: mapping from indices to classes
+            - subtext_fields: ordered subtext fields that the user want to
+                concatenate into an input text in the same order
+            - digit_label:  specify whether label in dataset is digit
+            - text_label: specify whether label in dataset is text
+            - one_based_index_label: if dataset provides digit label,
+                check if it's one-based index, if so set it to True.
+            - skip_first_line: many datasets' first line are columns names,
```
hunterhector, line 195: a more generic implementation here would be "skip_k_starting_lines" so that it can be an integer that can be configured, -1 could mean not skip

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+                set this config to True if it's the case.
+        """
+        config: dict = super().default_configs()
+        config.update(
```
hunterhector, line 199: we don't do `config.update` anymore, should directly return the new set of configs.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+                    "ft.onto.base_ontology.Title",
+                    "ftx.onto.ag_news.Description",
+                ],
+                "index2class": None,
+                "subtext_fields": [
+                    "ft.onto.base_ontology.Title",
+                    "ftx.onto.ag_news.Description",
+                ],
+                "digit_label": True,
+                "text_label": False,
```
hunterhector, line 212: First, I didn't find the usage of `text_label` config, secondly, this and the `digit_label` can be merged into one variable.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+    https://github.com/asyml/forte/examples/classification
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._index2class = None
+        self._class2index = None
+
+    def set_up(self):
+        assert self.configs.index2class is not None, (
```
hunterhector, line 71: in setup you should also check other configuration values, like `data_fields` and so on

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+            "User must set _index2class to enable"
```
hunterhector, line 72: `_index2class` is a wrong name, and you should mention this is set in config

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def _parse_pack(
+        self, line_info: Tuple[int, List[str]]
+    ) -> Iterator[DataPack]:
+        line_id, line = line_info
+        # content label title
+        pack = DataPack()
+
+        # subtext fields must follow the ontology names
+        data_fields = self.configs.data_fields
```
hunterhector, line 122: this assignment seems unnecessary

hunterhector (commented):
We should put more efforts on improving the docstring

forte/data/readers/classification_reader.py
feipenghe: Those checks require comparing with the input datapack so I checked under `_parse_pack` method.
https://github.com/feipenghe/forte/blob/96e14de31ddeac4d4f5201ac567d3b9076223579/forte/data/readers/classification_reader.py#L115

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+__all__ = ["ClassificationDatasetReader"]
+
+
+class ClassificationDatasetReader(PackReader):
+    r"""class:`ClassificationDatasetReader` is designed to read text
+    classification dataset that contains input text and digit/text labels.
+
+    User must specify `forte_data_fields` which is a list of forte ontologies
```
hunterhector, line 31: Try this:

This reader is designed to read table-like classification datasets, there are a couple of values that need to be provided via configuration, including `forte_data_fields`, `input_ontologies`, and `index2class`.

`forte_data_fields` is a list representing the column headers of the dataset.  Each element in the `forte_data_fields` list is either a label that indicates the document class or a Forte entry type that will be used to store the content. Apparently, this list should follow the column order of the target dataset. 

For example:  in the amazon polarity dataset (https://huggingface.co/datasets/amazon_polarity), the columns are [label, title, content]. We can configure `forte_data_fields`  to be ['label', 'ft.onto.base_ontology.Title',
    'ft.onto.ag_news.Description']. `label` is a special keyword to specify the label/document class column, while the latter two ontology types will be used by the reader to store the text from the `title` and `content` column respectively.

Users could also provide `input_ontologies` via configuration, to indicate which of the above-specified fields will be kept. For example, if `title` is not needed, then you can specify this value as [`ftx.onto.ag_news.Description `]. Then only the text from the description part will be stored.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+    The number of input ontologies and the order of concatenation
+    can be customized in the list. For example, if we only want titles and
+    descriptions in our input text, we specify "input_ontologies":
+    ['ft.onto.base_ontology.Title', 'ftx.onto.ag_news.Description']" in the
+    configuration. We can also only include descriptions by specifying
+    "input_ontologies":['ftx.onto.ag_news.Description']" which is very
+    flexible in customizing input text.
+
+
+    User must also specify `index2class` which is the mapping from zero-based
```
hunterhector, line 53: why zero-based, what if my dataset is not zero based

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+    User must specify `forte_data_fields` which is a list of forte ontologies
+    or labels and it should be one-to-one correspondence with the original data
+    fields. This configuration helps generating labels and initializing data
+    fields within corresponding ontologies wrappers.
+    For example, for amazon polarity dataset,
+    https://huggingface.co/datasets/amazon_polarity, the original data fields
+    are [label, title, content]. We want to identify them as either class
+    label or specific ontologies by specifying
+    "forte_data_fields: ['label', 'ft.onto.base_ontology.Title',
+    'ft.onto.ag_news.Description']" in the configuration.
```
hunterhector, line 40: you use `ft` here but `ftx` later

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+
+    User must also specify `input_ontologies` that will be concatenate as
```
hunterhector, line 42: 1. the name `input_ontologies` is not intuitive, this is the field for text, so maybe something like `text_fields`
2. why this value is mandatory, will it be easier if we use all the fields from `forte_data_fields` by default?

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+    input text in a list.
```
hunterhector, line 43: what do you mean by `concatenate` text in a list, what list

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        yield pack
+
+    @classmethod
+    def default_configs(cls):
+        r"""This defines a basic configuration structure for classification dataset reader.
+
+        Here:
+            - forte_data_fields: these fields provides one-to-one
+            correspondence between given original dataset data fields and
+            labels/ontologies. For data fields without usage, user can
```
hunterhector, line 190:
```suggestion
            labels/ontologies. For columns that are not needed, users can
```

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+    'ft.onto.ag_news.Description']" in the configuration.
+
+    User must also specify `input_ontologies` that will be concatenate as
+    input text in a list.
+    The number of input ontologies and the order of concatenation
+    can be customized in the list. For example, if we only want titles and
+    descriptions in our input text, we specify "input_ontologies":
+    ['ft.onto.base_ontology.Title', 'ftx.onto.ag_news.Description']" in the
+    configuration. We can also only include descriptions by specifying
+    "input_ontologies":['ftx.onto.ag_news.Description']" which is very
```
hunterhector, line 49: The `default_config` uses ft.onto.base_ontology.Document, now we introduce this `ag_news` ontology, which is confusing.

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+                "ft.onto.base_ontology.Document",
+            ],
+            "digit_label": True,
+            "one_based_index_label": True,
+            "skip_k_starting_lines": 1,
+        }
+
+
+def generate_text_n_input_ontology_indices(
+    input_ontologies, forte_data_fields_dict
```
hunterhector, line 223: typing of arguments

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+):
```
hunterhector, line 224: return type of arguments

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Retrieve ontologies from data fields and concatenate them into text.
+    Also, we generate the indices for these ontologies accordingly.
+
+    Args:
+        input_ontologies: a list of ontology that needs to be concatenated into a input string.
+        forte_data_fields_dict: a dictionary with ontology names as key and
+            ontology string as value.
```
hunterhector, line 232: return value not documented

examples/classification/amazon_review_sentiment.py
```diff
@@ -0,0 +1,37 @@
...
+csv_path = "path_to_dataset/amazon_review_polarity_csv/test.csv"
+pl = Pipeline()
+
+# initialize labels
+class_names = ["negative", "positive"]
+index2class = dict(enumerate(class_names))
+pl.set_reader(ClassificationDatasetReader(),
+              config={"index2class": index2class})
+pl.add(NLTKSentenceSegmenter())
+pl.add(NLTKWordTokenizer())
```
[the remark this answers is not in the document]
feipenghe: I was following https://github.com/asyml/forte/blob/master/examples/pipelines/process_dataset_example.py and thought tokenization is usually part of the data pre-processing.

forte/data/readers/classification_reader.py
feipenghe: By default, we can use all the fields from `forte_data_fields` except for labels. Since both order and selected data fields are important for building correct input text, I didn't implement the default `forte_data_fields`. But I will add it now.

forte/data/readers/classification_reader.py
feipenghe: The list is `input_ontologies`. And text in the Forte entries in `input_ontologies` will be concatenated.

forte/data/readers/classification_reader.py
feipenghe: Previously, I referred to an `ag_news` example which uses 'ftx.onto.ag_news.Description', and later I changed to `ft.onto.base_ontology.Document` from 'ftx.onto.ag_news.Description'. The minor issue is that the concatenated text is also stored in `ft.onto.base_ontology.Document` and it results in two `Document` for each DataPack, one for one input column name and the other for concatenated text. Maybe I should implement a superclass or a child class of `Document` to differentiate the two data?

forte/data/readers/classification_reader.py
feipenghe: Keys are only used internally and values are `class_names` from dataset. The dataset class names can be any digit/text that is stored in values.

forte/data/readers/classification_reader.py
feipenghe: added return type.

examples/classification/bank_customer_intent.py
```diff
@@ -0,0 +1,125 @@
...
+    "why_verify_identity",
+    "wrong_amount_of_cash_received",
+    "wrong_exchange_rate_for_cash_withdrawal",
+]
+index2class = dict(enumerate(class_names))
+
+# initialize reader config
+this_reader_config = {
+    "forte_data_fields": [
+        "ftx.onto.base.Document",
```
hunterhector, line 99: what is this entry type? I didn't find it in our repo

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,258 @@
...
+    `forte_data_fields` is a list representing the column headers of the
+    dataset. Each element in the `forte_data_fields` list is either a label
+    that indicates the document class or a Forte entry type that will be used
+    to store the content. Apparently, the list should follow the column order
+    of the dataset.
+
+    For example, for amazon polarity dataset,
+    (https://huggingface.co/datasets/amazon_polarity), the column names
+    are [label, title, content]. We can configure `forte_data_fields`to be
+    ['label', 'ft.onto.base_ontology.Title', 'ft.onto.base.Document'].
```
hunterhector, line 41:
```suggestion
    ['label', 'ft.onto.base_ontology.Title', 'ft.onto.base_ontology.Document'].
```

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # it determines the order of concatenation
+        text_fields = self.configs.text_fields
+
+        # get text and ontology indices
+        text, input_ontology_indices = generate_text_n_input_ontology_indices(
+            text_fields, df_dict
+        )
+        pack.set_text(text)
+        # self.set_text(pack, text)
```
hunterhector, line 159: Should use `self.set_text`, since `pack.set_text` will not call `replace_func` automatically, in fact, in the future we could even make `pack.set_text` to be private and always force people to use `self.set_text`

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,258 @@
...
+        pack.pack_name = line_id
+        yield pack
+
+    @classmethod
+    def default_configs(cls):
+        r"""This defines a basic configuration structure for classification dataset reader.
+
+        Here:
+            - forte_data_fields: these fields provides one-to-one
+            correspondence between given original dataset column names and
```
hunterhector, line 201: the indent here should be wrong, could you try to build the sphinx doc and double check the format

forte/data/readers/classification_reader.py
hunterhector: need return at the function signature too

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,258 @@
...
+    Retrieve ontologies from data fields and concatenate them into text.
+    Also, we generate the indices for these ontologies accordingly.
+
+    Args:
+        text_fields: a list of ontology that needs to be concatenated into a input string.
+        forte_data_fields_dict: a dictionary with ontology names as keys and
+            ontology strings as values.
+
+    Returns:
+        str, dict: a concatenated text and dictionary that keys are forte data
```
hunterhector, line 244: Tuple[str, dict[str, Tuple[int, int]]]

forte/data/readers/classification_reader.py
hunterhector: If you want to have `Document` to represent the whole thing, `Title` for the title, and something for the body, then maybe you should create an ontology like `Body` for body. Note that `body` should not be a child class nor super class of `Document`.

forte/data/readers/classification_reader.py
feipenghe: Is it expecting some errors if it's wrong? I changed the format to double spaces instead of column starting from `Here` column. And use `sphinx-build -W -b html -d _build/doctrees . _build/html` before and after changing the format doesn't make any difference.

hunterhector: no unfortunately there is no error (an error would make this easier). basically, we would want the documentation to look good. Based on some of my experience, if the word `correspond` is not indented correctly (now it is indented at the same level of the `-` symbol), it will break the list structure. Can you maybe post a screenshot of the documentation you generated here so that I can take a look?

hunterhector: Here is an example of the documentation that's rendered terribly: https://asyml-forte.readthedocs.io/en/latest/code/data.html#forte.data.data_pack.DataPack.get_original_span, you can see from the [source](https://asyml-forte.readthedocs.io/en/latest/_modules/forte/data/data_pack.html#DataPack.get_original_span), the docstring looks fine, but the indent is wrong, which generated messy documentation.

feipenghe: I found other reader docs generated but not the `classification_reader.py`. Is there any common reason for it?

hunterhector: oh, i see, I always forget about this. We need to edit one file to include this: https://github.com/asyml/forte/blob/master/docs/code/data.rst

examples/classification/bank_customer_intent.py
```diff
@@ -0,0 +1,125 @@
...
+    "why_verify_identity",
+    "wrong_amount_of_cash_received",
+    "wrong_exchange_rate_for_cash_withdrawal",
+]
+index2class = dict(enumerate(class_names))
+
+# initialize reader config
+this_reader_config = {
+    "forte_data_fields": [
+        "ft.onto.base.Document",
```
hunterhector, line 99: This type name is still wrong. Did you try run this example? Be careful about these.

examples/classification/bank_customer_intent.py
```diff
@@ -0,0 +1,125 @@
...
+        "label",
+    ],
+    "index2class": index2class,
+    "text_fields": [
+        "ft.onto.base.Document"
```
hunterhector, line 104: Be careful of the type, and please run the example and make sure it works

forte/data/readers/classification_reader.py
```diff
@@ -0,0 +1,260 @@
...
+    `forte_data_fields` is a list representing the column headers of the
+    dataset. Each element in the `forte_data_fields` list is either a label
+    that indicates the body class or a Forte entry type that will be used
+    to store the content. Apparently, the list should follow the column order
+    of the dataset.
+
+    For example, for amazon polarity dataset,
+    (https://huggingface.co/datasets/amazon_polarity), the column names
+    are [label, title, content]. We can configure `forte_data_fields`to be
+    ['label', 'ft.onto.base_ontology.Title', 'ft.onto.base.Body'].
```
hunterhector, line 41: This "ft.onto.base.Body" is also wrong, please be careful of what you wrote, too much wrong typing here and there.

forte/data/readers/classification_reader.py
hunterhector: where's the return type

examples/classification/bank_customer_intent.py
hunterhector: Please resolve this

forte/data/readers/classification_reader.py
hunterhector: Please resolve this

forte/data/readers/classification_reader.py
hunterhector: Please resolve this

hunterhector (commented):
Can you check the type names you write and resolve all comments

forte/data/readers/classification_reader.py
feipenghe: I render it `classification_reader.html` through online html viewer. It looks like this. I am not sure how to render it locally like https://asyml-forte.readthedocs.io/en/latest/code/data.html 

image

examples/classification/bank_customer_intent.py
feipenghe: Fixed.

examples/classification/bank_customer_intent.py
feipenghe: Fixed.

forte/data/readers/classification_reader.py
feipenghe: Fixed.

feipenghe:
> Can you check the type names and resolve all pending comments

For the typing of input parameters of `generate_text_n_input_ontology_indices()`. I think I fixed in version [1/27's commits](https://github.com/asyml/forte/blob/335a2c6e9ae10562aac2d7b47b80ce16be98ce8a/forte/data/readers/classification_reader.py) but the one shown here is from a much earlier commit.

forte/data/readers/classification_reader.py
hunterhector: So you have the HTML files ready? What do you mean by rendering locally? You can simply view HTML pages the with browser.

forte/data/readers/classification_reader.py
hunterhector: Return type is needed for the function annotation, don't only the docstring.

def generate_text_n_input_ontology_indices(
    text_fields: List[str], forte_data_fields_dict: Dict[str, str]
) --> <return type here>:

hunterhector:
> generate_text_n_input_ontology_indices

I don't see the return type for the function annotation (not the docstring)

forte/data/readers/classification_reader.py
feipenghe: This is the new rendering result that I used the method I mentioned in the meeting.
image

feipenghe:
> > generate_text_n_input_ontology_indices
> 
> I don't see the return type for the function annotation (not the docstring)

Fixed it.

examples/classification/amazon_review_sentiment.py
```diff
@@ -0,0 +1,35 @@
+import sys
```
hunterhector, line 1: copyright header

hunterhector (approved):

PR #526, opened 2021-09-12
Implementing Word Splitting Data Augmentation Algorithm

Pushkar-Bhuse:
### Description of changes
This PR implements a Word Splitting Data Augmentation Algorithm in Forte. This algorithms splits n random words at into two (The position at which the word is split is randomly chosen) where n = alpha * input length.

This PR adds an additional data augmentation algorithm to Forte. It also helps understand ways to add additional annotations from the same processor.

### Test Conducted
The tests check if the generated text is equal to the expected one. Also, it checks if the words that were split are not present in the generated text.

### DIFF
--- a/forte/processors/data_augment/algorithms/word_splitting_processor.py
+++ b/forte/processors/data_augment/algorithms/word_splitting_processor.py
@@ -1,0 +1,124 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Data augmentation processor for the Random Word Splitting operation.
+Randomly choose n words (With length greater than 1) and split it at a random position.
+Do this n times, where n = alpha * input length.
+Example: Original Text -> "I will be there soon." , Augmented Text -> "I w ill be there so on."
+"""
+
+from math import ceil
+import random
+from typing import List, Iterable
+
+from forte.common.configuration import Config
+from forte.common.resources import Resources
+from forte.data.data_pack import DataPack
+from forte.data.multi_pack import MultiPack
+from forte.data.ontology import Annotation
+from forte.processors.data_augment import ReplacementDataAugmentProcessor
+from forte.utils.utils import get_class
+
+__all__ = ["RandomWordSplitDataAugmentProcessor"]
+
+
+class RandomWordSplitDataAugmentProcessor(ReplacementDataAugmentProcessor):
+    r"""
+    This class creates a processor to perform Random Word Splitting.
+    It randomly chooses n words in a sentence and splits each word at a random position
+    where n = alpha * input length.
+    """
+
+    def initialize(self, resources: Resources, configs: Config):
+        super().initialize(resources, configs)
+
+    def _augment(self, input_pack: MultiPack, aug_pack_names: List[str]):
+        augment_entry = get_class(self.configs["augment_entry"])
+
+        for pack_name in aug_pack_names:
+            data_pack: DataPack = input_pack.get_pack(pack_name)
+            annotations: List[List(Annotation, int)] = []
+            test: List[List(Annotation, int)] = []
+            endings = []
+            annos: Iterable[Annotation] = data_pack.get(augment_entry)
+            for idx, anno in enumerate(annos):
+                test.append([anno.text, idx])
+                annotations.append([anno, idx])
+                endings.append(anno.end)
+            if len(annotations) > 0:
+                annotation_to_split = random.sample(
+                    [
+                        anno
+                        for anno in annotations
+                        if (anno[0].end - anno[0].begin) > 1
+                    ],
+                    ceil(self.configs["alpha"] * len(annotations)),
+                )
+                annotation_to_split = sorted(
+                    annotation_to_split, key=lambda x: x[1], reverse=True
+                )
+                for i in range(len(annotation_to_split)):
+                    src_anno, src_idx = annotation_to_split[i]
+                    splitting_position = random.randrange(
+                        1, (src_anno.end - src_anno.begin)
+                    )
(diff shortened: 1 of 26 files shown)


forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Data augmentation processor for the Random Word Splitting operation.
+Randomly choose n words (With length greater than 1) and split it at a random position.
+Do this n times, where n = alpha * input length.
+"""
```
hunterhector, line 19: Quick example at the module docstring would be helpful

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+from forte.data.data_pack import DataPack
+from forte.data.multi_pack import MultiPack
+from forte.data.ontology import Annotation
+from forte.processors.data_augment import ReplacementDataAugmentProcessor
+from forte.utils.utils import get_class
+
+__all__ = ["RandomWordSplitDataAugmentProcessor"]
+
+
+class RandomWordSplitDataAugmentProcessor(ReplacementDataAugmentProcessor):
```
hunterhector, line 36: Need a class docstring to specify what it does

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+    def initialize(self, resources: Resources, configs: Config):
+        super().initialize(resources, configs)
+
+    def _augment(self, input_pack: MultiPack, aug_pack_names: List[str]):
+        augment_entry = get_class(self.configs["augment_entry"])
+
+        for pack_name in aug_pack_names:
+            data_pack: DataPack = input_pack.get_pack(pack_name)
+            annotations: List[Annotation] = []
+            pos = []
```
hunterhector, line 46: Maybe rename this to `endings`? Just easier for reading.

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+            annos: Iterable[Annotation] = data_pack.get(augment_entry)
+            for anno in annos:
+                if len(anno.text) > 1:
```
hunterhector, line 49: So `anno.end - anno.begin > 1` should be slightly better than `anno.text` since `anno.text` will do an additional string slicing

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+                    annotations.append((anno, anno.end))
+                    pos.append(anno.end)
+            if len(annotations) > 0:
+                for _ in range(ceil(self.configs["alpha"] * len(annotations))):
+                    annotation_to_split = random.choice(annotations)
```
hunterhector, line 54: I think we can use `random.sample` to take multiple ones instead of using `for+choice`

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+                    src_anno = annotation_to_split[0]
+                    insert_pos = annotation_to_split[1]
```
hunterhector, line 56:
```suggestion
                    src_anno, insert_pos = annotation_to_split
```

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+                    splitting_position = random.randrange(1, len(src_anno.text))
```
hunterhector, line 57: Again, use `src_anno.end - src_anno.begin` instead `len(src_anno.text)`

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+                    word_split = [
+                        src_anno.text[:splitting_position],
+                        src_anno.text[splitting_position:],
+                    ]
+                    if insert_pos != pos[-1]:
+                        word_split[1] = word_split[1] + " "
```
hunterhector, line 63: hmm, so the second word may have an additional space after it? This may make the `Annotation` not accurate.

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,101 @@
...
+
+                    first_position = insert_pos
+                    second_position = insert_pos + 1
+
+                    self._delete(src_anno)
+                    self._insert(word_split[0], data_pack, first_position)
+                    self._insert(word_split[1], data_pack, second_position)
```
hunterhector, line 70: I think `delete+insert+insert` may not give us the original annotations, can they? 

If necessary maybe we can modify the base class a bit to accommodate something like this.

tests/forte/processors/data_augment/algorithms/word_splitting_processor_test.py
```diff
@@ -0,0 +1,106 @@
...
+        self.nlp.add(
+            component=WhiteSpaceTokenizer(), selector=AllPackSelector()
+        )
+
+    @data(
+        (
+            [
+                "Mary and Samantha arrived at the bus station early "
+                "but waited until noon for the bus ."
+            ],
```
hunterhector, line 55: Could we try an two-sentence data. and then use `PeriodSentenceSplitter`, see if the resulting `Sentence` is aligned as expected?

hunterhector (commented):
See comments in line.

Further, I am not certain that `delete` + 2 `insert` would work, but your test case seems fine to me. Do you know what happens under the hood? Why this case would work?

forte/processors/data_augment/algorithms/word_splitting_processor.py
wanglec: I also have this question. In addition, should we place the space in front of it or after it, and what about the first insertion word?

hunterhector:
Hmm, so a merge has caused many files that are in the master branch being in this PR, which makes it slightly hard to review the code.

How about you re-create a clean branch from the current master, add your updates there (should probably be only a few files). And then we re-open a PR?

Pushkar-Bhuse:
I made the new PR.

hunterhector:
this PR is fixed in https://github.com/asyml/forte/pull/532

PR #532, opened 2021-09-18
Updated Branch : Implementing Word Splitting Data Augmentation

Pushkar-Bhuse:
### Description of changes
This PR implements a Word Splitting Data Augmentation Algorithm in Forte. This algorithms splits n random words at into two (The position at which the word is split is randomly chosen) where n = alpha * input length.

This PR adds an additional data augmentation algorithm to Forte. It also helps understand ways to add additional annotations from the same processor.

### Test Conducted
The tests check if the generated text is equal to the expected one. Also, it checks if the words that were split are not present in the generated text.


tests/forte/processors/data_augment/algorithms/word_splitting_processor_test.py
```diff
@@ -0,0 +1,114 @@
...
+        self.nlp.add(component=RandomWordSplitDataAugmentProcessor())
+        self.nlp.initialize()
+
+        for idx, m_pack in enumerate(self.nlp.process_dataset(texts)):
+            aug_pack = m_pack.get_pack("augmented_input_src")
+
+            self.assertEqual(aug_pack.text, expected_outputs[idx])
+
+            for j, token in enumerate(aug_pack.get(Token)):
+                self.assertEqual(token.text, expected_tokens[idx][j])
```
hunterhector, line 107: Can we add some test cases for other Annotation?

For example, let's label "Mary and Samantha" and "the bus station" as "EntityMention", and check if they can be recovered correctly?

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,122 @@
...
+                        word_split[1] = " " + word_split[1]
+
+                    self._insert(word_split[1], data_pack, second_position)
+                    self._delete(src_anno)
+                    self._insert(word_split[0], data_pack, first_position)
+
+    @classmethod
+    def default_configs(cls):
+        """
+        Returns:
```
hunterhector, line 94: I just realize this is not included in the documentation, can you add it here: https://github.com/asyml/forte/blob/master/docs/code/data_aug.rst

Btw, can you also help add the typo op to the doc too: https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/typo_replacement_op.py

Also, can you help check if the generated docstring is correct for this function? I think the indent and new lines can affect the outcome.

To check this (after you added them in the rst file above), perform `cd docs` and then `make html`, and then browse to the page to see if the outcome is expected.

I think this should be a good example: https://asyml-forte.readthedocs.io/en/latest/code/data.html#forte.data.batchers.ProcessingBatcher.default_configs

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,122 @@
...
+                "augment_entry": "ft.onto.base_ontology.Token",
+                "other_entry_policy": {
+                    "type": "",
+                    "kwargs": {
+                        "ft.onto.base_ontology.EntityMention": "auto_align",
+                        "ft.onto.base_ontology.Document": "auto_align",
+                        "ft.onto.base_ontology.Sentence": "auto_align",
+                    },
+                },
+                "data_aug_op": "forte.processors.data_augment.algorithms.",
```
hunterhector, line 114: This `data_aug_op` does not point to a correct class name.

forte/processors/misc/simple_processors.py
```diff
@@ -58,3 +59,30 @@ def _process(self, input_pack: DataPack):
 
         if start < len(input_pack.text):
             input_pack.add_entry(Token(input_pack, start, len(input_pack.text)))
+
+
+class EntityMentionInserter(PackProcessor):
```
hunterhector, line 64: Are you using this class for test case? Please don't implement it in the main src, implement this class in the test case folders.

forte/processors/data_augment/algorithms/word_splitting_processor.py
hunterhector: Some parameters, like this `data_aug_op` is probably not even used in your class. Consider remove it?

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -0,0 +1,126 @@
...
+            - alpha: 0 <= alpha <= 1. indicates the percent of the words
+            in a sentence that are changed.
+        Config Values:
+            - `alpha` (float): The probability of splitting in [0, 1](Default = 0.1).
+        """
+        config = super().default_configs()
+        config.update(
+            {
+                "augment_entry": "ft.onto.base_ontology.Token",
+                "other_entry_policy": {
```
hunterhector, line 111: Is the "type" key under this "other_entry_policy" used?

Btw, I think the docstring should also explain what other keys are, especially what the "other_entry_policy" does and how to use it.

And I have a few suggestions for the default values:

1. `ft.onto.base_ontology.EntityMention` does not seem to be important and be the default. "ft.onto.base_ontology.Document" and "Sentence" might be fine.

Pushkar-Bhuse: I had gotten a configuration error on removing "type"

hunterhector: Understood.

hunterhector: Just update the other docstrings and I think it is fine. 

I think "type" might be an obsolete config but not fully erased from the base implementation. Would you mind find out what it does and see what we should do with it?

hunterhector (approved):

Issue #435, opened 2021-04-20
Handle unknown attributes and unknown types


hunterhector:
This issue is blocked by https://github.com/asyml/forte/issues/405.

**Is your feature request related to a problem? Please describe.**
Forte ontology defines the data types that can be used in the system,  sometimes the ontology evolves by adding new classes or new attributes. Some existing pipelines will break (during deserialization) since they encountered unknown types and attributes. 

Example:
The old ontology may define an `EntityMention`, with attribute `ner_type`. At some point we updated the definition, adding a new attribute, say `score` to `EntityMention`, now the `EntityMention` class changes from

```
class EntityMention:
   ner_type = ""
```
to 

```
class EntityMention:
   ner_type = ""
   score = 0
```

Loading data with the new types or attributes will break the deserialization. 

**Describe the solution you'd like**
Forte can be robust to these changes when a boolean flag is set to `on`, in such cases, it should ignore unknown entry types. 

On the other direction, when we load old data to a new pipeline, we can assign default `None` values to unknown entry attributes. 

**Describe alternatives you've considered**

**Additional context**

J007X:
After some research related to the current version changes it looks like current introduction of version to our pre-defined classes are still not yet able to handle the dynamic nature of the ontology type changes. Maybe in the future some design changes in ontology that each type defined will somehow keep some kind of version or strict type definition

hunterhector:
> After some research related to the current version changes it looks like current introduction of version to our pre-defined classes are still not yet able to handle the dynamic nature of the ontology type changes. Maybe in the future some design changes in ontology that each type defined will somehow keep some kind of version or strict type definition

Yes. Actually, the purpose of the current version system is not intended to solve these issues. We can consider revisiting this issue after the "data_efficiency" project, where we will introduce new data format and serialization mechanism

J007X:
After reviewing code and take into consideration of previous discussions,  as the current jsonpickle does not have a usable (or very consistent) exception for us to utilize, and adding strict types are consider some (bigger and) new enhancement not only to this issue (as it might also solve #434). We can close this for now until new feature design can be settled (for example in the new data format and serialization mech mentioned above).

hunterhector:
If we simply close this one, it is a bit hard to track. We probably should keep this open (I've already added a `won't fix` tag). Otherwise we need a different issue.

J007X:
Thanks Hector for adding the tag. I'm creating a new feature request for this one and related 434 about "Allowing more strict type "enforcement"

PR #560, opened 2021-11-07
Add exceptions on unknown types during deserialization

VincentYaoMBZUAI:
This PR is a follow-up for https://github.com/asyml/forte/issues/435. 

### Description of changes
Due to both the current serialization and deserialization functions being based on outsourcing packages and not self-designed, so there's not a suitable method to totally solve this issue. In addition, I added an AttributeError exception logging to "_deserialization" function of BasePack class. This might help move a small step forward for the AttributeError traceback.

There would be a piece of error information when the new types or attributes break the deserialization to help AttributeError traceback

### Test Conducted
We can use "stave_reader_test.py" as well to test this modification without using "line 54: sys.path.append(onto_path)" to make sure the newly created path is not in the python path.


forte/data/base_pack.py
```diff
@@ -193,10 +195,43 @@ def _deserialize(
...
+                except AttributeError as e:
+                    logging.exception(
+                        "There was an unknown type or attribute during"
+                        "deserialization and broke the process."
+                        "It might be caused by adding"
+                        "new classes or new attributes to the ontology."
+                        "Details are: %s",
+                        e,
+                    )
+                    # raise AttributeError(
```
hunterhector, line 229: I think we should raise the exception here and "fail fast"

forte/data/base_pack.py
```diff
@@ -193,10 +194,27 @@ def _deserialize(
...
             with _open(data_source, mode="rb") as f:  # type: ignore
-                pack = pickle.load(f)
+                try:
+                    pack = pickle.load(f)
+                except AttributeError as e:
+                    raise AttributeError(
+                        "There was an unknown type or attribute during"
+                        "deserialization and broke the process."
+                        "It might be caused by adding"
+                        "new classes or new attributes to the ontology."
```
hunterhector, line 216: Can we have a test case to see if this exception will be raised?

Btw, here we have a similar problem: https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L134, and we are able to identify which entry class cause the problem, can we do something similar?

hunterhector:
Any updates on this? Currently, I don't think the added exception will be raised.

forte/data/base_pack.py
hunterhector: Test case?

hunterhector (commented):
Missing test cases

hunterhector:
Still no test cases for this PR.

Issue #561, opened 2021-11-10
Add DataPackBoxer in Caster


mylibrar:
**Is your feature request related to a problem? Please describe.**
Forte already has a [MultiPackBoxer](https://github.com/asyml/forte/blob/257f5d4798c632a67e63a4d549c81f2b3b87b714/forte/data/caster.py#L49) that can cast `DataPack` into `MultiPack`. We may want to add a new caster that can perform the opposite conversion, i.e., from `MultiPack` to `DataPack`.

**Describe the solution you'd like**
Add a new class called `DataPackBoxer` in [caster.py](https://github.com/asyml/forte/blob/master/forte/data/caster.py#L49) and implement the `cast()` method to cast `MultiPack ` into `DataPack`.

**Describe alternatives you've considered**
We may introduce a configuration parameter called `"pack_name"` to extract the target DataPack from input MultiPack.

**Additional context**
A corresponding unit test for the new `DataPackBoxer` class is also desired.

PR #564, opened 2021-11-28
MultiPack to SinglePack boxer

VincentYaoMBZUAI:
This PR fixes https://github.com/asyml/forte/issues/561

### Description of changes
This PR creates a new class "DataPackBoxer" to cast a DataPack from a MultiPack, where the DataPack is the only content of the original MultiPack, indexed by the attribute`pack_name`. It would be able to auto-box the multi-pack into a data-pack by simple 'getting pack' and returning an iterator that produces the boxed data-pack.

Forte already has a MultiPackBoxer that can cast DataPack into MultiPack. The PR will be able to add a new caster that can perform the opposite conversion, casting MultiPack to DataPack.

### Test Conducted
A test case "test_datapack_boxer()" is designed and added in "data_type_infer_test.py", which can be used to test the caster by checking the output_pack_type is DataPack while input_pack_type is MultiPack.


hunterhector:
the PR message need to be updated as the following:

`This PR fixes [https://github.com/asyml/forte/issues/561]. `

`This PR fixes https://github.com/asyml/forte/issues/561. `

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
     @staticmethod
     def output_pack_type():
         return MultiPack
+
+
+class DataPackBoxer(Caster[MultiPack, DataPack]):
+    """
+    This class creates a DataPack from a MultiPack, this DataPack is the only
+    contain of the original MultiPack, indexed by the :attr:`pack_name`.
```
hunterhector, line 86: `:attr:pack_name` is a bit unclear. We should mention this comes from the configurtions.

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
     @staticmethod
     def output_pack_type():
         return MultiPack
+
+
+class DataPackBoxer(Caster[MultiPack, DataPack]):
+    """
+    This class creates a DataPack from a MultiPack, this DataPack is the only
```
hunterhector, line 85: Check the grammar too.

`this DataPack is the only contain` is a bit confusing

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
...
+    contain of the original MultiPack, indexed by the :attr:`pack_name`.
+    """
+
+    def cast(self, pack: MultiPack) -> DataPack:
+        """
+        Auto-box the multi-pack into a data-pack by simple getting pack.
```
hunterhector, line 91: So we don't refer to them as `MultiPack` or `DataPack`, sometimes `Data Pack`. Let's keep this consistent.

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
...
+
+        Args:
+            pack: The data pack to be boxed
```
hunterhector, line 94: The description is wrong, this is the MultiPack to take DataPack from.

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
...
+
+        Returns: An iterator that produces the boxed data-pack.
```
hunterhector, line 96: The return value is not an iterator

forte/data/caster.py
```diff
@@ -78,3 +78,43 @@ def input_pack_type():
...
+
+        """
+        if pack.pack_name:
+            pack_name = (
+                pack.pack_name.strip("_multi")
+                if "_multi" in pack.pack_name
+                else pack.pack_name
+            )
+        else:
+            pack_name = None
```
hunterhector, line 106: Doesn't quite make sense here, if `pack_name` is None, then what's the behavior of `get_pack`?

There are a couple of ways that I consider how this boxer would work:
1. allow using `pack_index` to take a pack, and we can make the default value to be 0
2. check whether `pack_index` and `pack_name` both exist and try to pick the preferred one
3. throw an exception during the initialize stage to check if we have valid `pack_name` and `pack_index`

tests/forte/data/datapack_type_infer_test.py
```diff
@@ -37,3 +36,11 @@ def test_multipack_boxer(self, component):
...
         self.assertTrue(caster.input_pack_type() is DataPack)
         self.assertTrue(caster.output_pack_type() is MultiPack)
+
+    @data(
+        DataPackBoxer,
+    )
+    def test_datapack_boxer(self, component):
+        caster = component()
+        self.assertTrue(caster.input_pack_type() is MultiPack)
+        self.assertTrue(caster.output_pack_type() is DataPack)
```
hunterhector, line 46: The test case doesn't contain the new class `DataPackBoxer`

hunterhector:
The correct way to write the PR message is as follow:

`This PR fixes https://github.com/asyml/forte/issues/561`

You need to remove the square brackets

forte/data/caster.py
```diff
@@ -78,3 +81,46 @@ def input_pack_type():
     @staticmethod
     def output_pack_type():
         return MultiPack
+
+
+class DataPackBoxer(Caster[MultiPack, DataPack]):
+    """
+    This class creates a DataPack from a MultiPack, this DataPack is the only
+    content of the original MultiPack.
```
hunterhector, line 89: Not sure if `this DataPack is the only content` is clear. It seems to imply the original MultiPack only contains one data pack.  Also, we don't "create" a DataPack here. In fact, what a caster does is to pass on the pack, here, instead of passing the multipack, it pass on a single data pack within this pack.

forte/data/caster.py
```diff
@@ -78,3 +81,46 @@ def input_pack_type():
...
+        Auto-box the MultiPack into a DataPack by using pack_index to take the unique pack.
+
+        Args:
+            pack: The MultiPack to be boxed.
+
+        Returns: A DataPack boxed from the MultiPack.
+
+        """
+
+        p = pack.get_pack_at(0)
```
hunterhector, line 103: why are we getting index 0? We should ask for which index to get here.

forte/data/caster.py
```diff
@@ -78,3 +81,46 @@ def input_pack_type():
...
+
+        if p.pack_name:
+            self.configs.pack_name = (
+                p.pack_name.strip("_multi")
+                if "_multi" in p.pack_name
+                else p.pack_name
+            )
+        else:
+            self.configs.pack_name = None
```
hunterhector, line 112: What is this part doing here? Looks like they change the values in `self.configs`. But why do we need to change the config? They will not affect anything useful to the users.

What we need to do here, is to return the pack from the multipack that matches the `pack_name` or `pack_index`.

data_samples/ontonotes/00_1/abc_0039.gold_conll
```diff
@@ -0,0 +1,80 @@
+#begin document (bn/abc/00/abc_0039); part 000
```
hunterhector, line 1: I think we can reuse the current dataset without adding a new file

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
     @staticmethod
     def output_pack_type():
         return MultiPack
+
+
+class DataPackBoxer(Caster[MultiPack, DataPack]):
+    """
+    This pass on a single DataPack within the MultiPack.
```
hunterhector, line 88: `passes on`

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
     @staticmethod
     def output_pack_type():
         return MultiPack
+
+
+class DataPackBoxer(Caster[MultiPack, DataPack]):
```
hunterhector, line 86: More intuitively, this might be called "MultiPackUnboxer"

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
...
+    """
+    This pass on a single DataPack within the MultiPack.
+    """
+
+    def cast(self, pack: MultiPack, pack_index: int = 0) -> DataPack:
```
hunterhector, line 91: The `cast` function signature should be consistent, we should not add this additional `pack_index` attribute in this function.

As we've discussed before, we should move this to the configs.

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
...
+        Auto-box the MultiPack into a DataPack by using pack_index to take the unique pack.
+
+        Args:
+            pack: The MultiPack to be boxed.
+
+        Returns: A DataPack boxed from the MultiPack.
+
+        """
+
+        if pack_index in range(pack.num_pack):
```
hunterhector, line 102: to check if the index is in range, I think we can simply check for less?

`if pack_index < range(pack.num_pack)`

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
...
+            p = pack.get_pack_at(pack_index)
+            return p
+        else:
+            raise KeyError(
+                f"pack_index: {pack_index} is not in this multi-pack."
+            )
+
+    @classmethod
+    def default_configs(cls):
+        return {"pack_name": "default_multi"}
```
hunterhector, line 112: why this `pack_name` is not used?

forte/data/caster.py
```diff
@@ -78,3 +81,40 @@ def input_pack_type():
...
+
+        Returns: A DataPack boxed from the MultiPack.
+
+        """
+
+        if pack_index in range(pack.num_pack):
+            p = pack.get_pack_at(pack_index)
+            return p
+        else:
+            raise KeyError(
```
hunterhector, line 106: IndexError is more appropriate than KeyError

tests/forte/datapack_boxer_test.py
```diff
@@ -0,0 +1,79 @@
...
+        nlp_2.add(DataPackBoxer())
+        nlp_2.initialize()
+
+        dataset_path = data_samples_root + "/ontonotes/00_1"
+        expected_pack_name_multi = "bn/abc/00/abc_0039_multi"
+        expected_pack_name = "bn/abc/00/abc_0039"
+
+        # check that the MultiPack is yielded
+        pack_1 = nlp_1.process(dataset_path)
+        self.assertEqual(pack_1.pack_name, expected_pack_name_multi)
```
hunterhector, line 71: this checks the name, can we also check the instance class type with isinstance?

tests/forte/pipeline_test.py
```diff
@@ -734,6 +734,36 @@ def test_pipeline_multipack_reader(self):
         # check that all packs are yielded
         self.assertEqual(num_packs, reader.count)
 
+    # def test_multi_pack_to_data_pack_boxer(self):
```
hunterhector, line 737: if this part is not needed, can you remove them?

data_samples/ontonotes/00_1/abc_0039.gold_conll
hunterhector: we should remove this file if it is no longer needed

hunterhector (commented):
overall looks fine to me, I think we just need to remove the redundant file and fix the CI to merge this.

hunterhector:
btw, besides removing the redundant `ontonotes/00_1`, we should also update the branch to make it up to date.

hunterhector (approved):

Issue #566, opened 2021-12-07
Add metric ontology


zhanyuanucb:
**Is your feature request related to a problem? Please describe.**
In some tasks like training, tuning, or validation, we need components to report metrics so that we can gauge their performance.

**Describe the solution you'd like**
Define metric ontology so that components can report metrics in output DataPack(s). More specifically, define:
- `SingleMetric` represents metrics containing a single number. For example, accuracy, loss...
- `NdMetric` represents metrics that are N-d arrays. For example, cross-validation scores (1-d array), confusion matrix (2-d array)...

**Describe alternatives you've considered**

**Additional context**
The current ontology generator couldn't add `ndarray` attribute to an ontology. Need to create another issue for this feature.

hunterhector:
Let's start assigning owners to the issues since there is another team that working to pick issues from our repo.

PR #568, opened 2021-12-08
Add metric ontology

zhanyuanucb:
This PR fixes https://github.com/asyml/forte/issues/566

### Description of changes
Added `forte/ontology_specs/metric.json`

Due to the `ndarray` attribute in `NdMetric`, the metric ontology spec isn't ready to be auto-generated by the current ontology generator yet.

### Test Conducted
Unit tests are to be added


hunterhector:
Could you do the following to make sure the PR is up to standard?

1. Change the PR name to the content instead of a number
2. Make the branch up to date to the branch
3. Make sure the CI passes.

hunterhector:
This PR is fixing the wrong issue link

forte/ontology_specs/metric.json
```diff
@@ -0,0 +1,39 @@
+{ 
+    "name": "metric", 
+    "definitions": [ 
+      {
+        "entry_name": "ft.onto.metric.BaseMetric",
```
hunterhector, line 5: I believe we can simply use `Metric`, just to make things simple

forte/ontology_specs/metric.json
```diff
@@ -0,0 +1,39 @@
...
+        "description": "A base metric entity, all metric entities should inherit from it.", 
+        "attributes": [ 
+          { 
+              "name": "metric_name", 
+              "type": "str" 
+          }
+        ]
+      },
+      { 
+        "entry_name": "ft.onto.metric.SingleMetric",
```
hunterhector, line 16: We will define some metrics as examples in this file later (such as accuracy), how about we add them now?

hunterhector (commented):
We normally also add the generated class to Forte repo so users don't have to re-generate these.

forte/ontology_specs/metric.json
zhanyuanucb: I thought by assigning proper `metric_name` (e.g. `accuracy` or `loss_value`) metrics like accuracy or loss value could be covered by SingleMetric.

hunterhector: I see. Having a `metric_name` is not a bad idea, maybe we should do that.

zhanyuanucb: The base `Metric` class already had this `metric_name`

hunterhector: i see, now i understand

hunterhector (approved):

Issue #547, opened 2021-10-25
JSON Schema for serialized data packs


hunterhector:
**Is your feature request related to a problem? Please describe.**
JSON Schema can be used to check whether a data pack serialization is correct, which can help the system to locate problems early and precisely. However, currently, there is no JSON Schema defined in this system.

**Describe the solution you'd like**
Create the JSON Schema for the data packs. It'd be nice if we can create a process to create the schema, here is one proposal:

1. To create a data pack with representative entries stored and serialize that to JSON (this process can be stored for future use)
2. May use a tool to convert the JSON to a schema
3. Edit the resulting schema to generalize it if needed.

**Describe alternatives you've considered**

**Additional context**
The JSON Schema may need to be updated after https://github.com/asyml/forte/issues/405.

PR #574, opened 2021-12-20
JSON Schema for serialized data packs #547

batermj:
This PR fixes https://github.com/asyml/forte/issues/547.

### Description of changes
Add version in the version file and one specific version member variable to base pack.

With the version member variable set in the base pack class, version system were set up.

### Test Conducted
Serialization and Deserialization.


hunterhector:
The correct way to write the PR message is the following:

`This PR fixes https://github.com/asyml/forte/issues/547. `

This way Github will automatically link the PR and the issue

forte/version.py
```diff
@@ -19,3 +19,4 @@
 VERSION_SHORT = "{0}.{1}".format(_MAJOR, _MINOR)
 VERSION = "{0}.{1}.{2}".format(_MAJOR, _MINOR, _REVISION)
 FORTE_IR_VERSION = "0.0.1"
+PACK_VERSION = [0, 0, 1]
```
hunterhector, line 22: We should probably also use the `Major` `Minor` `Revision` style for pack version too.

forte/data/base_pack.py
```diff
@@ -97,6 +98,7 @@ def __init__(self, pack_name: Optional[str] = None):
         super().__init__()
         self.links: List[LinkType] = []
         self.groups: List[GroupType] = []
+        self.pack_version: List[int] = PACK_VERSION
```
hunterhector, line 101: Since a version spec should only contain 3 fields, we shouldn't use `List` but should store these explicitly, probably as `NamedTuple`?

hunterhector (changes requested):
There are no test cases added for this PR. Although it is mentioned that a serialization test is conducted, there are no new tests used to test the `pack_version`. 

By having the version, we should expect Forte to check it during the serialization stage, and there are a couple of tests we should perform:
1. After serialisation, check the string to make sure the version is there (and consistency with version.py)
2. Create a file with a different version, and throw an error with a proper message.
3. After loading the correct pack, we could check whether the version is correct.

There could be other test cases that could help out.

hunterhector:
I just realize this PR is submitted to solve #547, but what it actually does is adding the versions, how is this different from `https://github.com/asyml/forte/pull/581`?

hunterhector:
close since no updates

Issue #567, opened 2021-12-08
Ontology generator allows ndarray attribute


zhanyuanucb:
**Is your feature request related to a problem? Please describe.**
Users cannot add `ndarray` attributes to their custom ontology through the current ontology generator. Given that Forte is able to take `ndarray` in [`Entry`](https://github.com/asyml/forte/blob/master/forte/data/ontology/core.py#L182) and [`Query`](https://github.com/asyml/forte/blob/master/forte/data/ontology/top.py#L523), it will be helpful to allow the ontology generator to take in `ndarray` attribute.

**Describe the solution you'd like (WIP)**
Files needed to be modified:
- `ontology_code_const.py`: add `ndarray` to `SUPPORTED_PRIMITIVES`
- `ontology_code_generator`

**Describe alternatives you've considered**

**Additional context**

PR #575, opened 2021-12-20
WIP: generator accepts NdArray attribute

zhanyuanucb:
This PR fixes https://github.com/asyml/forte/issues/567. 

### Description of changes

- In ontology_code_const.py:

  - Add “NdArray” to COMPOSITE

  - Add “shape” and “dtype” to SchemaKeyWords

- In ontology_code_generator.py:
  - modify parse_property to handle att_type == “NdArray”
  - define parse_ndarray to parse ndarray attributes:
    - ndarray_value_type
  - return a NdArrayProperty

- In code_generation_objects.py
  - add NdArrayProperty
 
- In forte.data.ontology.core.py
  - add FNdArray that wraps a numpy ndarray and store related properties like dtype and shape

- validation_schema.json
  - Add schema for NdArray attribute

### Test Conducted
- `tests/forte/data/ontology/ontology_code_generator_test.py`
  - test generated code for sample Entry containing NdArray attribute
- `tests/forte/data/ontology/ndarray_attribute_test.py`:
  - Test serialization
  - Test setting `array` property


forte/data/ontology/ontology_code_generator.py
```diff
@@ -1078,6 +1079,38 @@ def parse_entry(
...
+        att_name: str,
+        att_type: str,
+        desc: str,
+    ):
+        if SchemaKeywords.ndarray_value_type not in schema:
+            raise TypeNotDeclaredException(
+                f"Item type for the attribute {att_name} of the entry "
+                f"[{entry_name.class_name}] not declared. This attribute is "
+                f"a composite type: {att_type}, it should have a "
+                f"{SchemaKeywords.ndarray_value_type}."
```
hunterhector, line 1096: This exception is not used properly. `TypeNotDeclaredException` is used when the ontology spec uses a certain type, but we don't know what type it is (hence not declared).

For example, the ontology `Sentence` uses the type `SpecialToken`, but we don't know what `SpecialToken` is.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -1078,6 +1079,38 @@ def parse_entry(
...
+            )
+        value_type = schema[SchemaKeywords.ndarray_value_type]
+        # TODO: add validaton
+        manager.add_object_to_import(value_type)
+
+        self_ref = entry_name.class_name == value_type
```
hunterhector, line 1102: why are you using a `self_ref` here? do you know what `self_ref` does?

forte/data/ontology/code_generation_objects.py
```diff
@@ -382,6 +382,45 @@ def to_field_value(self):
         return self.name
 
 
+class NdArrayProperty(Property):
```
hunterhector, line 385: The current implementation is basically a copy of DictProperty, which isn't what we want. What we need to do here is to design an `NDarray` spec language, and then translate that to code. Here, you used `value_type`, what's its purpose? In a `DictProperty`, we use `key_type` and `value_type` because that's what a `Dict` looks like (it has key and value so we ask for key type and value_type).

In an `NDarray`, there shouldn't be values for users to provide, we should only accept types that are available for a numpy array (e.g. int64, float, str etc). If the `value_type` is for the `numpy.ndarray`, then we also do not want the schema to provide it, we should hard-code this internally (perhaps use a variable internally `ndarry_class` for this).

forte/ontology_specs/metric.json
```diff
@@ -0,0 +1,48 @@
...
+        "entry_name": "ft.onto.metric.NdMetric", 
+        "parent_entry": "ft.onto.metric.BaseMetric", 
+        "description": "A N-dimensional metric entity, used to present a metric of N-d array (e.g. confusion matrix).", 
+        "attributes": [ 
+          { 
+              "name": "value", 
+              "type": "NdArray",
+              "value_type": "numpy.ndarray"
+          },
+          {
```
hunterhector, line 36: attributes like `shape` and `dtype` should be part of `NdArray` property, they shouldn't be separate attributes to handle. And `"value_type": "numpy.ndarray"` looks redundant to me, we surely cannot support any arbitrary value type here so we either hard-code it, or only provide a few implementations for user to choose from, we should not use arbitrary class.

forte/data/ontology/core.py
```diff
@@ -598,6 +598,63 @@ def __iter__(self) -> Iterator[KeyType]:
         yield from self.__data
 
 
+class FNdArray:
```
hunterhector, line 601: Let's add more docstring to this.

forte/data/ontology/core.py
```diff
@@ -598,6 +598,63 @@ def __iter__(self) -> Iterator[KeyType]:
...
+    @property
+    def dtype(self):
+        return self._dtype
+
+    @property
+    def shape(self):
+        return self._shape
+
+    @property
+    def array(self):
```
hunterhector, line 625: is it better to call this func `data`?

forte/data/ontology/ontology_code_generator.py
```diff
@@ -1078,6 +1080,38 @@ def parse_entry(
...
+    def parse_ndarray(
+        self,
+        manager: ImportManager,
+        schema: Dict,
+        att_name: str,
+        desc: str,
+    ):
+        ndarray_dtype = None
+        if SchemaKeywords.ndarray_dtype in schema:
+            # TODO: validate dtype
```
hunterhector, line 1092: what's the missing validation here? is it to check if the value is acceptable? This would be nice, but we do have json schema that can validate this.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -1078,6 +1080,38 @@ def parse_entry(
...
+            ndarray_dtype = schema[SchemaKeywords.ndarray_dtype]
+
+        ndarray_shape = None
+        if SchemaKeywords.ndarray_shape in schema:
+            # TODO: validate shape
+            ndarray_shape = schema[SchemaKeywords.ndarray_shape]
+
+        # TODO: Throw warning if either dtype or shape is missing
```
hunterhector, line 1100: plan to deal with this todo? I found that `todo` will simply be forgotten in the future. We either fix them here or create issues to solve them in the future.

forte/data/ontology/validation_schema.json
```diff
@@ -95,6 +95,14 @@
                 "value_type": {
                   "description": "Item type for the case of Dice attributes",
                   "type": "string"
+                },
+                "ndarray_dtype": {
```
hunterhector, line 99: do we have some tests to validate a JSON spec with this schema, hopefully, both success and failure cases?

hunterhector (commented):
The PR looks very good, only a few small comments.

forte/data/ontology/validation_schema.json
zhanyuanucb: Added a test against invalid shape.
Do you think we can rely on numpy to validate dtype?

forte/data/ontology/ontology_code_generator.py
zhanyuanucb: done.

forte/data/ontology/ontology_code_generator.py
zhanyuanucb: Now I don't think we need to validate dtype here since numpy will raise a TypeError if the dtype is not acceptable.

forte/data/ontology/core.py
zhanyuanucb: Sounds good

forte/data/ontology/core.py
zhanyuanucb: Agree

forte/data/ontology/code_generation_objects.py
```diff
@@ -382,6 +383,45 @@ def to_field_value(self):
...
```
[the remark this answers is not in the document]
zhanyuanucb: Agree

forte/data/ontology/core.py
```diff
@@ -601,8 +601,8 @@ def __iter__(self) -> Iterator[KeyType]:
 class FNdArray:
     """
     FNdArray is a wrapper of a NumPy array that stores shape and data type
-    of the array when they are specified. Only when both shape and data type
-    are provided, will FNdArray initializes an empty array accordingly.
+    of the array if they are specified. Only when both shape and data type
+    are provided, will FNdArray initialize an empty array accordingly.
```
hunterhector, line 605: So is `empty` accurate here, better refer to the numpy documentation to ensure users expect the actual content of the array correctly: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html

forte/data/ontology/validation_schema.json
hunterhector: it'd be great if schema can check the possible dtype values. Since this is done without actually running the code, while numpy validation happens late at run time.

tests/forte/data/ontology/test_specs/test_invalid_ndarray_shape.json
```diff
@@ -0,0 +1,18 @@
...
+      { 
+        "entry_name": "ft.onto.test_ndarray.BadShapeNdEntry1", 
+        "parent_entry": "forte.data.ontology.top.Generics", 
+        "description": "A N-dimensional entity having invalid shape.", 
+        "attributes": [ 
+          { 
+              "name": "value", 
+              "type": "NdArray",
+              "ndarray_dtype": "int",
+              "ndarray_shape": "[2, 2]"
```
hunterhector, line 13: I am curious how is the shape invalid

zhanyuanucb: an array of integers is expected here. so it should be a `[2, 2]` instead of a string `"[2, 2]"`; likewise, an integer like `2` is also invalid here.

hunterhector: oh, i see, this is a string, insteresting

forte/data/ontology/core.py
zhanyuanucb: agree

forte/data/ontology/validation_schema.json
zhanyuanucb: sounds good.
I've added more tests against dtype and shape

hunterhector: the validation looks awesome, and thanks for the tests

hunterhector (approved):

Issue #497, opened 2021-08-04
Allow generate_ontology to generate namespace packages


mylibrar:
**Is your feature request related to a problem? Please describe.**
The autogenerated packages of ontology are regular packages, which means they all have `__init__.py` file under package directory. We want to support namespace package generation maybe by adding a flag in `generate_ontology` to control whether to create `__init__.py`.

**Describe the solution you'd like**
Maybe we can modify https://github.com/asyml/forte/blob/1e3ba68f7272a99ccde87ee87e3450793e585383/forte/data/ontology/ontology_code_generator.py#L432 to allow namespace package generation and add a flag to control it.

**Describe alternatives you've considered**

**Additional context**

PR #577, opened 2021-12-26
Allow generate_ontology to generate namespace packages

VincentYaoMBZUAI:
This PR fixes https://github.com/asyml/forte/issues/497. 

### Description of changes
Adding a flag "with_init" in generate_ontology to control whether to create "__init__.py". The flag "with_init" is also added in "code_generation_objects.ModuleWriter.write" and "code_generation_objects.ModuleWriter.make_module_dirs()" to control from original place that whether the "__init__.py" is going to be generated.

If the flag "with_init" is set False, the `__init__.py` is not to be generated automatically.

### Test Conducted
Test cases "test_with_init" and "test_without_init" are both designed in "ontology_code_negerator_test.py".


hunterhector:
this PR contains commits from a different PR (the data pack boxer). We shouldn't do it this way. Our convention is to open up a clean branch for each commit, and change only the commits related to this PR

forte/data/ontology/code_generation_objects.py
```diff
@@ -622,6 +626,8 @@ def make_module_dirs(
               placed
             include_init: True if `__init__.py` is to be generated in existing
               packages in which `__init__.py` does not already exists
+            with_init: True if `__init__.py` is to be generated automatically;
+                False, the `__init__.py` is not to be generated automatically
```
hunterhector, line 630: Notice that `namespace packaging` does not mean we don't generate `__init__` at all. Notice that there is an `include_init` attribute only, which will decide whether `__init__` will be generated. Here, we actually only want to not generate the `__init__` at the top level, like documented here: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages

For example, the ontology definition could be `ft.onto.something`, the folder structure would be `ft/onto/something`, we don't generate an `init` at `ft` level, but still include them in `onto` level. I guess it would require some understanding to the namespace packaging mechanism to do this correctly

hunterhector: Further, the variable should not be named as `with_init`, we should name it like `use_name_space_packaging` to make the purpose clear.

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -175,6 +175,130 @@ def test_include_and_exclude_init(self):
...
+                temp_dir,
+                is_dry_run=False,
+                include_init=True,
+                with_init=False,
+            )
+            gen_files = sorted(utils.get_generated_files_in_dir(folder_path))
+
+            # Assert the generated python files
+            exp_file_path = [
+                "ft/__init__",
```
hunterhector, line 247: `__init__` file is still here? I thought we want to not generate them if the variable is False

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -175,6 +175,130 @@ def test_include_and_exclude_init(self):
...
+                temp_dir,
+                is_dry_run=False,
+                include_init=True,
+                with_init=True,
+            )
+            gen_files = sorted(utils.get_generated_files_in_dir(folder_path))
+
+            # Assert the generated python files
+            exp_file_path = [
+                "ft/__init__",
```
hunterhector, line 196: and why are we having `__int__` path, should they be `__init__.py`?

forte/data/base_pack.py
```diff
@@ -32,6 +32,7 @@
     Iterable,
 )
 
+
```
hunterhector, line 35: let's not add unnecessary changes to this PR.

forte/data/caster.py
```diff
@@ -37,14 +37,6 @@ class Caster(
     def cast(self, pack: InputPackType) -> OutputPackType:
         raise NotImplementedError
 
-    @staticmethod
-    def input_pack_type():
-        raise NotImplementedError
-
-    @staticmethod
-    def output_pack_type():
-        raise NotImplementedError
```
hunterhector, line 46: let's not make changes to irrelevant files in this PR.

forte/data/ontology/code_generation_objects.py
```diff
@@ -622,6 +626,10 @@ def make_module_dirs(
               placed
             include_init: True if `__init__.py` is to be generated in existing
               packages in which `__init__.py` does not already exists
+            use_name_space_packaging: True if namespace packaging is to be used,
+                and `__init__.py` is to be generated at the top level;
+                False, namespace packaging is to be used,
```
hunterhector, line 631: docstring might be wrong, what I read here is not matter True of False, namespace packaging is to be used.

Just to clarify, when we use name_space_packaging, we don't generate at the top level

forte/data/ontology/code_generation_objects.py
```diff
@@ -638,12 +646,21 @@ def make_module_dirs(
                 Path(os.path.join(temp_path, AUTO_GEN_FILENAME)).touch()
 
             # Create init file
-            if not dest_path_exists or include_init:
-                init_file_path = os.path.join(temp_path, "__init__.py")
-                with open(init_file_path, "w", encoding="utf-8") as init_file:
-                    init_file.write(f"# {AUTO_GEN_SIGNATURE}\n")
-
-    def write(self, tempdir: str, destination: str, include_init: bool):
+            if use_name_space_packaging:
```
hunterhector, line 649: how does `use_name_space_packaging` control the top level `__init__` here?

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -178,6 +178,130 @@ def test_include_and_exclude_init(self):
...
+                "ft/onto/example_import_ontology",
+            ]
+            exp_files = sorted(
+                [
+                    f"{os.path.join(folder_path, file)}.py"
+                    for file in exp_file_path
+                ]
+            )
+
+            # Now, corrupt two of the init files
```
hunterhector, line 210: what does "corrupt" mean?

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -178,6 +178,130 @@ def test_include_and_exclude_init(self):
...
+            corrupted_path = os.path.join(folder_path, "ft/__init__.py")
+            corrupted_path_onto = os.path.join(
+                folder_path, "ft/onto/__init__.py"
+            )
+            exp_files = [
+                file
+                for file in exp_files
+                if file != corrupted_path and file != corrupted_path_onto
+            ]
```
hunterhector, line 219: Why not just write the expected files list explicitly here? This variable is changed 3 times with difficult-to-follow logic. Now I don't understand what's the content of this `exp_files`

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -178,6 +178,130 @@ def test_include_and_exclude_init(self):
...
+            exp_files = sorted(
+                [
+                    f"{os.path.join(folder_path, file)}.py"
+                    for file in exp_file_path
+                ]
+            )
+            self.assertEqual(gen_files, exp_files)
+
+            # Now, corrupt one of the init files
+            corrupted_path = os.path.join(folder_path, "ft/__init__.py")
```
hunterhector, line 285: Again, why this "corrupt file"? The behavior of "use_name_space_packaging" should simply be it will control whether a top level init file is generated. I don't understand the corruption logic here.

tests/forte/data/ontology/ontology_code_generator_test.py
hunterhector: Could you explain why when `use_name_space_packaging` is true, we still have `ft/__init__.py`? And please write write `ft/__init__.py`. Do not append ".py" later, this adds complexity in understanding the logic

mylibrar (commented):
- Instead of adding a boolean flag `use_name_space_packaging`, maybe it's better to set an integer argument `namespace_depth` to allow customized number of levels of namespace packaging. We want to disable the generation of `__init__.py` for all the directory levels above `namespace_depth`. For example, if we have an ontology `level1.levle2.level3.something` and `namespace_depth=2`, then we remove `__init__.py` under `level1` and `level1/level2` while keeping `__init__.py` under `level1/level2/level3`. When `namespace_depth<=0`, we just disable namespace packaging and include `__init__.py` in all directory levels.
- This new feature should be reflected in `generate-ontology` CLI: https://github.com/asyml/forte/blob/master/forte/command_line/generate_ontology/__main__.py

hunterhector:
updates?

forte/data/ontology/code_generation_objects.py
```diff
@@ -684,12 +690,21 @@ def make_module_dirs(
                 Path(os.path.join(temp_path, AUTO_GEN_FILENAME)).touch()
 
             # Create init file
-            if not dest_path_exists or include_init:
-                init_file_path = os.path.join(temp_path, "__init__.py")
-                with open(init_file_path, "w", encoding="utf-8") as init_file:
-                    init_file.write(f"# {AUTO_GEN_SIGNATURE}\n")
-
-    def write(self, tempdir: str, destination: str, include_init: bool):
+            if not use_name_space_packaging:
```
mylibrar, line 693: Looks like this is also going to disable the generation of `__init__.py` below the top level. Can we verify that `__init__.py` is only removed at the top level directory?

forte/data/ontology/code_generation_objects.py
```diff
@@ -684,12 +690,29 @@ def make_module_dirs(
...
+                if not dest_path_exists or include_init:
+                    init_file_path = os.path.join(temp_path, "__init__.py")
+                    with open(
+                        init_file_path, "w", encoding="utf-8"
+                    ) as init_file:
+                        init_file.write(f"# {AUTO_GEN_SIGNATURE}\n")
+
+            elif temp_path == os.path.join(tempdir, self.pkg_dir):
```
hunterhector, line 701: It seems this fix is used to fix a comment abut, to control that we don't generate `__init__.py` at the top level. But there are two issues:

1. The major issue, this solution doesn't seem to work. It tries to generate init at the `self.pkg_dir` level, but this is not the top-level if I understand it correctly, it is probably the last level. Thus this fix is trying to generate `__init__.py` **only** inside the last level. I think since your test cases only contain 2-layer folders, this cannot be spotted. To do this, we need to feed in the folder level during the function calls. 
2. The minor one, duplicate code: from 695 to 699 and 703 to 707, we simply copy past the code, we should conduct all logical checks first, and only have this code once.

mylibrar (dismissed):

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -178,6 +178,50 @@ def test_include_and_exclude_init(self):
             exp_files = [file for file in exp_files if file != corrupted_path]
             self.assertEqual(gen_files, exp_files)
 
+    def test_namespace_depth(self):
```
hunterhector, line 181: maybe we could use `@data` to try different namespace depth and different expected output. Currently we only tested `depth=1` right?

forte/command_line/generate_ontology/__main__.py
```diff
@@ -194,6 +194,18 @@ def main():
         help="If True, will not enforce prefix check.",
     )
 
+    create_parser.add_argument(
```
hunterhector, line 197: Can we try to incorporate this new feature into the documentation here? https://github.com/asyml/forte/blob/master/docs/toc/ontology_generation.md

forte/command_line/generate_ontology/__main__.py
```diff
@@ -194,6 +194,18 @@ def main():
...
+        "-n",
+        "--namespace_depth",
+        type=int,
+        required=False,
+        default=-1,
+        help="set an integer argument namespace_depth"
+        "to allow customized number of levels of namespace packaging."
+        "The generation of __init__.py for all the"
+        "directory levels above namespace_depth will be disabled.",
```
hunterhector, line 206: we should describe what is the default value, and what -1 means in the help section.

forte/command_line/generate_ontology/__main__.py
hunterhector: probably still no updates to the readme?

tests/forte/data/ontology/ontology_code_generator_test.py
```diff
@@ -77,105 +77,52 @@ def assert_generation_equal(self, file_a, file_b):
...
+                is_dry_run=False,
+                include_init=False,
+                namespace_depth=namespace_depth,
             )
             gen_files = sorted(utils.get_generated_files_in_dir(folder_path))
 
-            # Assert the generated python files after removing the corrupted
-            # file which should not have been regenerated
-            exp_files = [file for file in exp_files if file != corrupted_path]
             self.assertEqual(gen_files, exp_files)
```
hunterhector, line 126: Is it the case that the `exp_files` is the same no matter when `include_init = True` or `include_init = False`?  I think their behavior should be different. Do you need to remove the temporary files between different generation steps?

VincentYaoMBZUAI: Hi Hector,
1. the readme is updated, please kindly check __main__.py line 202 and 206-209
2. May I ask the priority of include_init and namespace_depth, as they are controlling the same behaviour at some generation steps. Currently, namespace_depth has a higher priority.

hunterhector: 1. I didn't see the readme got updated. I mean literally this file: https://github.com/asyml/forte/blob/master/docs/toc/ontology_generation.md, you can see here we list the options for the generation tool: https://github.com/asyml/forte/blob/master/docs/toc/ontology_generation.md#ontology-generation-steps, now since we update the tool, this part needs to be updated.
2. I see, so `include_init` might not be needed. So to clarify this, we have two options: a) remove the `include_init` b) explain in docstring on which one have priority, and what happen if we use one of them but not the other.   I would say removing `include_init` might be easier?

hunterhector (approved):

PR #585, opened 2022-01-08
Add audio support to DataPack

### LINKED ISSUE #582: Add support for audio data to DataPack
We want DataPack to support multi media data representation including text and audio. This allows DataPack to handle audio processing tasks (e.g., Audio Classification, Audio Separation and Segmentation) and can also drive several multi media use cases (e.g., automatic speech recognition (ASR)).

Support audio payload in DataPack as a numpy array with sample_rate, span_unit, seq_length, and other related info added as meta data.

Currently DataPack use [`DataPack._text`](https://github.com/asyml/forte/blob/818564c48d33bc5acf6f46a16a987bb88e1cf990/forte/data/data_pack.py#L157) to store text data, so maybe we can use `DataPack._audio` to store audio payload. Correspondingly, we might need to add some basic operations on the audio data, such as [`DataPack.audio`](https://github.com/asyml/forte/blob/818564c48d33bc5acf6f46a16a987bb88e1cf990/forte/data/data_pack.py#L241) and [`DataPack.set_audio()`](https://github.com/asyml/forte/blob/818564c48d33bc5acf6f46a16a987bb88e1cf990/forte/data/data_pack.py#L340).

The type of audio data can be a numpy array, which is a commonly used data structure to store raw waveform. Python libraries like [Librosa](https://librosa.org/doc/latest/index.html) and [soundfile](https://pysoundfile.readthedocs.io/en/latest/) support loading audio files from different formats (wav, flac, mp3, etc.) into a numpy array with its sample rate. We can add a `AudioReader` to wrap this loading operation.

We also need to store some metadata along with the waveform array. For example, `sample rate` is the key info that determines the unit of timestamps. Right now we have a [`span_unit`](https://github.com/asyml/forte/blob/818564c48d33bc5acf6f46a16a987bb88e1cf990/forte/data/data_pack.py#L89) for text, so maybe we should add a similar field for audio to handle different units (e.g., sample, segment, frame). Another example is `channel`, which specifies the number of channels of the audio (stereo or mono), though this can also be inferred from the shape of`DataPack._audio`. Other optional metadata includes `bit-depth`, which is an indicator of the resolution of audio measurement. All these information can be store in the `Meta` class [here](https://github.com/asyml/forte/blob/818564c48d33bc5acf6f46a16a987bb88e1cf990/forte/data/data_pack.py#L62).

mylibrar:
This PR fixes #582.

### Description of changes
- Add new dependency: `soundfile`
  - `soundfile>=0.10.3` is inserted to `setup.py` and `docs/requirements.txt`
  - CI workflow `main.yml` is updated with `sudo apt-get install -y libsndfile1-dev` as required by [soundfile for Linux](https://pysoundfile.readthedocs.io/en/latest/#installation).
- Update `DataPack` with audio support
  - Add `_audio` for payload and `sample_rate` for metadata
  - Corresponding operations are also added
- Add `AudioReader` to load audio files
  - User can provide a directory with a file extension, and the reader will automatically walk through the directory and load all the files with the specified suffix.

### Test Conducted
A unit test for `AudioReader` is added. It builds and runs an audio processing pipeline for automatic speech recognition (ASR) in order to jointly test `AudioReader` and `DataPack`.


forte/data/readers/audio_reader.py
```diff
@@ -0,0 +1,74 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The reader that reads audio files into Datapacks.
+"""
+import os
+from typing import Any, Iterator
+import soundfile
```
hunterhector, line 19: so this reader would depend on the `soundfile` dependency, which means it needs to be in our core requirement. So what happens would be if we do `pip install forte` and `from forte.data.readers.misc_reader import xxx`, this would fail (since this is in the `__init__.py`)

We need to think of a better way to place this reader. Any suggestions

forte/data/data_pack.py
```diff
@@ -365,6 +378,18 @@ def set_text(
             self.__orig_text_len,
         ) = data_utils_io.modify_text_and_track_ops(text, span_ops)
 
+    def set_audio(self, audio: np.ndarray, sample_rate: Optional[int]):
+        r"""Set the audio payload and sample rate of the :class:`DataPack`
+        object.
+
+        Args:
+            audio: A numpy array storing the audio waveform.
+            sample_rate: An integer specifying the sample rate.
```
hunterhector, line 387: do you think `sample_rate` should be required here? It can be optional in init

hunterhector (commented):
Thanks for the PR, looks like it works OK.

Some other small suggestions:
1. We are at a point where we have quite a few data samples in the `data_samples` folder. Could you add a `README.md` in that folder? This time let's add a description of what the `audio_reader_test` folder contains.
2. Once we add this PR we would need to start documenting the feature, so from the start, let's consider adding a new markdown file in the `docs` folder, start to serve as a tutorial for the audio project. Then we can make a link from https://github.com/asyml/forte/wiki, and from the root `README.md`
3.

forte/data/readers/audio_reader.py
```diff
@@ -30,6 +29,21 @@
 class AudioReader(PackReader):
     r""":class:`AudioReader` is designed to read in audio files."""
 
+    def initialize(self, resources, configs):
+        # pylint: disable=attribute-defined-outside-init
+        super().initialize(resources, configs)
+        try:
+            import soundfile  # pylint: disable=import-outside-toplevel
```
hunterhector, line 36: I think we can install at the class scope (line 31), then we don't need to store `self._soundfile` to access it inside the class.

setup.py
```diff
@@ -50,7 +50,7 @@
         # transformers 4.10.0 will break the translation model we used here
         "augment": ["transformers>=3.1, <=4.9.2", "nltk"],
         "stave": ["stave>=0.0.1.dev12"],
-        "audio": ["soundfile>=0.10.3"],
+        "audio_ext": ["soundfile>=0.10.3"],
```
hunterhector, line 53: let's add to the main README section on what these extra install mean

forte/data/readers/audio_reader.py
mylibrar: Seems like it will trigger an error when building docs: https://github.com/asyml/forte/runs/4783761361?check_suite_focus=true#step:6:37

README.md
```diff
@@ -48,6 +48,15 @@ cd forte-wrappers
 pip install src/spacy
 ```
 
+Some components or modules in forte may require some [extra requirements](https://github.com/asyml/forte/blob/master/setup.py#L45):
```
hunterhector, line 51: write down the command for `-e` install?

hunterhector (approved):

PR #586, opened 2022-01-12
add base_store and data_store class skeleton

### LINKED ISSUE #571: An interface class for the new efficient data structure
For the new efficient data structure, we need a new interface for all the operations, including add, get, delete, and attribute operations. For each API, we need to define its input, output and a description of functions.

**Depends on**

The interface requires  a class definition with abstract methods, including add_entry_raw, set_attr, get_attr, delete_entry, get, get_data, get_single.

**Additional context**

wanglec:
This PR fixes #571.

### Description of changes
Add the skeleton class for base_tank, which is also called data_tuple class previously. In this data structure, entries are stored in separate sorted lists as tuples according to their types for better performance. We will keep adding more details to the skeleton and implementing functions.

This class will be used by data_pack. It may potentially speed up data pack processing and improve efficiency.

### Test Conducted


forte/data/base_store.py
```diff
@@ -0,0 +1,160 @@
...
+    r"""The base class which will be used by :class:
+    `~forte.data.data_store.DataStore`."""
+
+    # pylint: disable=too-many-public-methods
+    def __init__(self):
+        r"""
+        This is a base class for the efficient underlying data structure. A current
+        implementation of `BaseStore` is `DataStore`.
+
+        A `BaseStore` contains a piece of natural language text and a
```
hunterhector, line 38: we probably don't need to store the `text`

forte/data/data_store.py
```diff
@@ -0,0 +1,231 @@
...
+                This class records the order of the array and assigns type id,
+                which is the index of each sortedlist, to each entry type.
+
+                Each entry list in the SortedList has the format of
+                [entry_type, tid, begin, end, attr_1, attr_2, ..., attr_n]
+                The first four fields are compulsory for every entry type.
+                Each entry type has a fixed field of attributes.
+                E.g. Document SortedList has lists of structure:
+                [entry_type, tid, begin, end, document_class,sentiment,classifications]
+        #
```
hunterhector, line 27: this # looks a bit funny in the docstring, is it required?

forte/data/data_store.py
```diff
@@ -0,0 +1,231 @@
...
+import uuid
+from forte.data.ontology.core import EntryType
+from forte.data.base_store import BaseStore
+
+
+class DataStore(BaseStore):
+    def __init__(self):
+        """An implementation from the dataframe-like data tuple object.
+
+                This class stores entries by types in different lists. It
```
hunterhector, line 12: Is the indention here done on purpose?

forte/data/data_store.py
```diff
@@ -0,0 +1,231 @@
...
+                uses an array to store these SortedLists with the structure:
+                array([Document SortedList, Sentence SortedList, ...]).
+                Different types of annotations, like sentence, tokens and documents,
+                are stored in different SortedLists.
```
hunterhector, line 16: This small part of documentation could be written a bit better

forte/data/data_store.py
```diff
@@ -0,0 +1,231 @@
...
+            "Token": ["pos", "ud_xpos", "lemma", "chunk", "ner", "sense",
+                    "is_root", "ud_features", "ud_misc"],
+            "Document": ["document_class", "sentiment", "classifications"],
+            "Sentence": ["speaker", "part_id", "sentiment", "classification",
+                        "classifications"],
+        }
+        TODO: implement get_type_attributes() (Issue #570)
+        """
+        # self.type_attributes: dict = get_type_attributes()
+        self.type_attributes: dict = {}
```
hunterhector, line 55: these attributes should probably all be private

hunterhector (commented):
we need a small test case to demo how this works, and in that test case, you can populate the values like `type_attributes`, `elements`, `entry_dict` with demo data.

forte/data/data_store.py
wanglec: I used black to format it and somehow it just became like this. I will remove the indentation.

hunterhector: To confirm if the documentation is correct, need to build the doc locally to see if the output is good: https://github.com/asyml/forte/blob/master/.github/workflows/main.yml#L137

hunterhector: sounds good!

tests/forte/data/data_store_test.py
```diff
@@ -0,0 +1,206 @@
...
+        # The order is [Document, Sentence]. Initialize 2 entries in each list.
+        # Document entries have tid 1234, 3456.
+        # Sentence entries have tid 9999, 1234567.
+        # The type id for Document is 0, Sentence is 1.
+        self.data_store.elements = [
+            SortedList(
+                [
+                    ["Document", 1234, 0, 5, None, "Postive", None],
+                    [
+                        "Document",
```
hunterhector, line 50: What's the purpose of storing "Document" value here?

tests/forte/data/data_store_test.py
```diff
@@ -0,0 +1,206 @@
...
+                        3456,
+                        10,
+                        25,
+                        "Doc class A",
+                        "Negative",
+                        "Class B",
+                    ],
+                ],
+                key=lambda x: (x[2], x[3]),
```
hunterhector, line 59: it'd be better if you simply put `begin` and `end` these top sorting keys at the first two positions, then you don't need to use a sorting lambda function here.

tests/forte/data/data_store_test.py
```diff
@@ -0,0 +1,206 @@
...
+                70,
+                None,
+                None,
+                "Negative",
+                "Class C",
+                "Class D",
+            ],
+        }
+
+    def test_add_annotation_raw(self):
```
hunterhector, line 114: Simply implement one of the functions and test it here, you cannot test NontImpelemtned functions

tests/forte/data/data_store_test.py
qinzzz: > What's the purpose of storing this string "Document" here?

When user use only tid to refer to an entry, we want to know the type, so we also store it.

tests/forte/data/data_store_test.py
qinzzz: > it'd be better if you simply put `begin` and `end` these top sorting keys at the first two positions, then you don't need to use a sorting lambda function here.

Make sense!

tests/forte/data/data_store_test.py
hunterhector: I see, note (as in a previous comment), this need to be a full qualified class name

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        r"""Get the next entry of the same type as the tid entry.
+
+        Args:
+            tid (int): Unique id of the entry.
+
+        Returns:
+            The next entry of the same type as the tid entry.
+
+        """
+        # We retrieve the entry list from entry_dict using tid. We get its
```
hunterhector, line 217: The `next_entry` and `prev_entry` APIs should be designed with efficient access. When we use the `next_entry` API, we should assume we have the information of this entry, especially, its exact location in the data store. So we shouldn't redo biset here, but use the location information to directly get next.

Example usage of next_entry:
1. next_entry(exact_locatoin_value)
2. Sentence.next() # The class version of the Sentence can have a `next` function that directly calls the next_entry, note that no bisect should be needede.

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        # entry type, type id, begin and end indexes. Then, we find the
+        # `entry_type` sortedlist using type id. We bisect the sortedlist
+        # to find the entry list. We then remove the entry list from both
+        # entry_dict and the `entry_type` sortedlist.
+
+        raise NotImplementedError
+
+    def get_entry(self, tid: int) -> List:
+        r"""Look up the entry_dict with key `tid`.
```
hunterhector, line 177: Here you should as well get the location information of this entry (for example, which sortedlist, and its index in the sortedlist). So that this info can be reuse later, in functions like `next_entry`

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        Returns:
+            The value of attr_name for the entry with tid.
+        """
+        # We retrieve the entry list from entry_dict using tid. We get its
+        # entry type. We then locate the index of the attribute using the entry
+        # type, field dictionary and attr_name, and get the attribute.
+
+        raise NotImplementedError
+
+    def delete_entry(self, tid: int):
```
hunterhector, line 159: there should be a lower-level function of this function, which takes the exact location of the entry and remove it. Using `tid` means you will have to search it first (bisect). This won't be efficient if we already have the bisect results. So there should also be a bisect function:

`delete_entry(tid)` will use `_delete_entry_by_loc(list_id, index_id)` and `bisect`

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        [entry_type, tid, begin, end, document_class,sentiment,classifications]
+
+        Args:
+            pack_name (Optional[str], optional): A name for this data store.
+        """
+        super().__init__()
+        self._text = ""
+
+        """
+        The _type_attributes is a private dictionay that provides entry types
```
hunterhector, line 34: when writing documentation, use this ` to wrap around the variable names, so you should do `_type_attributes` instead of _type_attributes

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        and their corresponding attribues.
+        The keys are all valid ontology types as strings, including all the
+        types defined in ft.onto and ftx.onto.
+        The values are all the valid attributes for this type, also defined in
+        ft.onto and ftx.onto.
+        This structure is supposed to be generated by another function
+        get_type_attributes() and ready to use in DataStore.
+
+        Example:
```
hunterhector, line 43: use `code-block` in sphinx to deal with the long section of code example

forte/data/data_store.py
```diff
@@ -0,0 +1,239 @@
...
+        _type_attributes = {
+            "ft.onto.base_ontology.Token": ["pos", "ud_xpos", "lemma", "chunk", "ner", "sense",
+                    "is_root", "ud_features", "ud_misc"],
+            "ft.onto.base_ontology.Document": ["document_class", "sentiment", "classifications"],
+            "ft.onto.base_ontology.Sentence": ["speaker", "part_id", "sentiment", "classification",
+                        "classifications"],
+        }
+        TODO: implement get_type_attributes() (Issue #570)
```
hunterhector, line 51: don't put todo in docstring, put in comment. Or simply not using todo, but link to issue to here.

forte/data/base_store.py
```diff
@@ -0,0 +1,160 @@
...
+from forte.data.ontology.core import EntryType
+
+__all__ = ["BaseStore"]
+
+
+class BaseStore:
+    r"""The base class which will be used by :class:
+    `~forte.data.data_store.DataStore`."""
+
+    # pylint: disable=too-many-public-methods
```
hunterhector, line 32: are we having `too-many-public-methods` in this class? if not, don't disable this

forte/data/base_store.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self):
+        r"""
+        This is a base class for the efficient underlying data structure. A
+        current implementation of `BaseStore` is `DataStore`.
+
+        A `BaseStore` contains a collection of NLP entries (annotations, links,
```
hunterhector, line 38: They are not necessarily NLP entries. BaseStore simply manage the data for Forte entries.

forte/data/data_store.py
```diff
@@ -0,0 +1,261 @@
...
+            sentiment, classifications]
+
+        Args:
+            pack_name (Optional[str], optional): A name for this data store.
+        """
+        super().__init__()
+        self._text = ""
+
+        """
+        The `_type_attributes` is a private dictionay that provides entry types
```
hunterhector, line 36:
```suggestion
        The `_type_attributes` is a private dictionary that provides entry types
```

forte/data/data_store.py
```diff
@@ -0,0 +1,261 @@
...
+        r"""It removes the entry list from the sortedlist of its type.
+
+        Args:
+            type_id (int): The index of the sortedlist in `self.elements`.
+            index_id (int): Index of the entry in the sortedlist.
+        """
+        # We then remove the entry list from `entry_type` sortedlist.
+        raise NotImplementedError
+
+    def get_entry(self, tid: int):
```
hunterhector, line 230: missing return type

forte/data/data_store.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def _new_entry(self, type: str, begin, end):
+        """
+        generate a new entry with default fields. called by add_entry_raw()
+        to create a new entry with `type`, `begin`, and `end`.
+
+        Args:
+            type (str): type of this entry
+        Returns:
+            A list representing a new entry
```
hunterhector, line 129: it might be worthy to also return the location information of the new entry

forte/data/data_store.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+
+        tid: int = uuid.uuid4().int
+        entry = [begin, end, tid, type]
```
hunterhector, line 133: Notice here you assume `begin` and `end`, this only works for the `Annotation` type, not for others such as `Link`. Thus this function currently can not create all types of entries. There are two solutions: 1. change this to _new_annotation but then you need functions to create other types 2. make this function generic and can handle all types of entries.

forte/data/data_store.py
```diff
@@ -0,0 +1,263 @@
+from typing import Union, Type, List, Iterator, Tuple
+
+import uuid
+from forte.data.ontology.core import EntryType
+from forte.data.base_store import BaseStore
+
+
```
hunterhector, line 7: missing __all__ here

hunterhector (commented):
nothing special left, just one thing

hunterhector (approved):

Issue #588, opened 2022-01-13
Doc for ndarray attribute


zhanyuanucb:
**Is your feature request related to a problem? Please describe.**
The recent merge of https://github.com/asyml/forte/pull/575 added support to ndarray attribute, which needs to be covered in the ontology generation tutorial.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #589, opened 2022-01-14
Ndarray doc

zhanyuanucb:
This PR fixes https://github.com/asyml/forte/issues/588. 

### Description of changes
Add doc for ndarray attribute in `docs/ontology_generation.md `

### Test Conducted


hunterhector (commented):
I think we need some examples to show how to define the nd array, what happens with them (like they will create numpy arrays).  This new type is complex and in a tutorial like this, only saying we "support" them won't be enough.

hunterhector (approved):

PR #592, opened 2022-01-14
Add ontologies for audio data

### LINKED ISSUE #591: Add ontologies for audio data
DataPack supports audio payload but we don't have any ontologies created for audio data. For text data, we have a rich set of ontologies like Document, Sentence, Word, etc. Maybe we should consider adding the audio counterparts into our ontology system.

We can follow the same structure as the Annotation class, which inherits from Entry and derives multiple span based ontologies like Sentence, Subword, ...
The proposed solution is to add a new entry called AudioAnnotation to the [top](https://github.com/asyml/forte/blob/master/forte/data/ontology/top.py). Then we can use it  to derive many useful ontologies (e.g., Recording and AudioUtterance) and add them to the [base ontology](https://github.com/asyml/forte/blob/master/forte/ontology_specs/base_ontology.json).

mylibrar:
This PR fixes #591 . 

### Description of changes
New top level entry: AudioAnnotation
- It's identical to the implementation of [Annotation](https://github.com/asyml/forte/blob/45b9551a5c959ca676f12adea781f2ac2117e9ba/forte/data/ontology/top.py#L52) except that it replaces `text` property with `audio`.
- DataPack is updated accordingly to support add/update/query/delete operations on audio annotations.

New ontologies in base_ontology: Recording, AudioAnnotation
- Update the base_ontology.json and regenerate base_ontology.py

### Test Conducted
A unit test `audio_annotation_test` is added to test all the new ontologies added.


forte/data/data_pack.py
```diff
@@ -353,6 +389,23 @@ def get_span_text(self, begin: int, end: int) -> str:
         """
         return self._text[begin:end]
 
+    def get_span_audio(self, begin: int, end: int) -> str:
+        r"""Get the audio in the data pack contained in the span.
```
hunterhector, line 393: worth mentioning what the `begin` means in terms of the unit.

forte/data/data_pack.py
```diff
@@ -866,6 +929,12 @@ def get_data(
                     "currently not supported."
                 )
 
+            if audio_annotation_types:
+                raise NotImplementedError(
+                    "Querying audio annotation types based on ranges is "
+                    "currently not supported."
+                )
```
hunterhector, line 936: This function fetches data that are covered by an `Annotation` type, that's why Audio annotations cannot be obtained from it, since they are not in the same payload. But it seems like we can extend this function to make it accept `Annotation` or `AudioAnnotation`, and retrieve corresponding data respectively. Could consider making a new issue about this as the next step.

forte/data/ontology/top.py
```diff
@@ -545,5 +546,133 @@ def update_results(self, pid_to_score: Dict[str, float]):
...
+                the provided entry type. Default `True`.
+
+        Yields:
+            Each `Entry` found using this method.
+
+        """
+        yield from self.pack.get(entry_type, self, components, include_sub_type)
+
+
 SinglePackEntries = (Link, Group, Annotation, Generics)
```
hunterhector, line 677: I guess `AudioAnnotation` belongs to this list too?

tests/forte/data/audio_annotation_test.py
```diff
@@ -0,0 +1,125 @@
...
+                    self.assertTrue(array_equal(
+                        audio_utter.audio,
+                        pack.audio[configs["begin"]:configs["end"]]
+                    ))
+
+            # Check `DataPack.delete_entry(AudioAnnotation)`
+            for audio_annotation in list(pack.get(AudioAnnotation)):
+                pack.delete_entry(audio_annotation)
+            self.assertEqual(len(list(pack.all_audio_annotations)), 0)
+
```
hunterhector, line 122: Feel like we should also add tests related to the serialization part to make sure the serialized data looks as expected

tests/forte/data/audio_annotation_test.py
```diff
@@ -0,0 +1,125 @@
...
+class RecordingProcessor(PackProcessor):
+    """
+    A processor to add a Recording ontology to the whole audio data.
+    """
+    def _process(self, input_pack: DataPack):
+        Recording(
+            pack=input_pack, begin=0, end=len(input_pack.audio)
+        )
+
```
hunterhector, line 38: Let's add an example using the audio reader, the audio annotations and so on in the examples folder, to serve as tutorial materials.

tests/forte/data/audio_annotation_test.py
hunterhector: Can refer to the codecov diff to notice the main untested lines: https://app.codecov.io/gh/asyml/forte/compare/592/diff

forte/data/data_pack.py
```diff
@@ -1426,8 +1504,9 @@ def in_span(self, inner_entry: Union[int, Entry], span: Span) -> bool:
         r"""Check whether the ``inner entry`` is within the given ``span``. The
         criterion are as followed:
 
-        Annotation entries: they are considered in a span if the begin is not
-        smaller than `span.begin` and the end is not larger than `span.end`.
+        Annotation/AudioAnnotation entries: they are considered in a span if the
+        begin is not smaller than `span.begin` and the end is not larger than
+        `span.end`.
```
hunterhector, line 1509: notice `Span` type does not have a unit associated with it. That means if we do `in_span(audio_annotation, text_annotation.span)`, the function actually runs, but this is not what we want. could this be a potential issue?

forte/data/data_pack.py
```diff
@@ -1426,8 +1504,9 @@ def in_span(self, inner_entry: Union[int, Entry], span: Span) -> bool:
...
 
         Link entries: if the parent and child of the links are both
         `Annotation` type, this link will be considered in span if both parent
```
hunterhector, line 1512: Here only mentions `Annotation`, but looks like for audio annotation this could work too.

docs/spelling_wordlist.txt
```diff
@@ -139,3 +139,4 @@ docstrings
 numpy
 jsonpickle
 crowdworkers
+AudioAnnotation
```
hunterhector, line 142: You don't need to add this to spelling list if every time you refer to it you use backtip around it

forte/data/data_pack.py
```diff
@@ -1191,23 +1264,50 @@ def get(  # type: ignore
...
+        # If the `entry_type` and `range_annotation` are for different types of
+        # payload, then we yield from an empty list with a warning.
+        if (
+            require_annotations(Annotation)
+            and isinstance(range_annotation, AudioAnnotation)
+        ) or (
+            require_annotations(AudioAnnotation)
+            and isinstance(range_annotation, Annotation)
+        ):
+            logger.warning(
```
hunterhector, line 1301: Is this conditional branch (i.e. the warning branch) tested?

forte/data/data_pack.py
```diff
@@ -1191,23 +1264,50 @@ def get(  # type: ignore
...
-                yield from []
-                return
+        if (
+            len(self.annotations) == 0
+            and isinstance(range_annotation, Annotation)
+            and require_annotations(Annotation)
+        ) or (
+            len(self.audio_annotations) == 0
+            and isinstance(range_annotation, AudioAnnotation)
+            and require_annotations(AudioAnnotation)
```
hunterhector, line 1287: wondering if this branch is tested.

hunterhector (commented):
There are a few things we could follow up with this:

1. Add a few test cases related to the newly updated condition branches, notice that test cases should cover most of these branches since bugs often happen in these less used branches. Check the codecov report if needed. Examples of not tested areas include the warning message, deleting `AudioAnnotation`, and other similar new conditional branches.

The other two items should have their own issues and tickets, could you create two issues based on them?

2. We have generally covered the `get` function but haven't covered the `get_data` function. The differences between these two functions: `get` obtain the classes while `get_data` return them as primitive types and nested dict. `get_data` also can handle m more complex requests. It is worth creating an issue regarding this.  Basically, we should have the `get_data` function also work but not simply raising NotImplementedError as Line935 in data_pack.py. 
3. The `index` and `build_coverage_index` functions should be polished to include AudioAnnotation as well. Related line is here: https://github.com/asyml/forte/blob/7adecbad0cc208a66ee2917060941b60a3fc789a/forte/data/data_pack.py#L1101

hunterhector (approved):

Issue #570, opened 2021-12-08
Generate and attribute structure from a forte entry class based on the fully qualified name


hunterhector:
**Is your feature request related to a problem? Please describe.**
Purpose: Entry names and their attribute structures are needed to create table-backed data structures (instead of class-backed). Given an entry class (or an entry name), we need to be able to find all its entry attributes, similar to what's defined in the ontology JSON class.

For example, for type "sentence[annotation]", we want to obtain all its attribute fields, including begin, end, tid, speaker, part_id, sentiment, and classification. 

**Describe the solution you'd like**
For each type, we want to obtain all the attributes. For example, for sentence we want to get a list of ["speaker", "part_id", "sentiment", "classification", "classifications"]. We want to get the attributes for each type of entry as a dictionary. The solution looks like the following:

```
# input can be a string
entry_name = "ft.onto.base_ontology.Sentence"
# or a class
entry_class = ft.onto.base_ontology.Sentence

# function signature
def get_entry_attribute_by_class(input_entry_class: Union[str, class]):

# output
["speaker", "part_id", "sentiment", "classification", "classifications"]
```

Note that we should only include attributes that we defined, not other default python attributes. So the output from `__dict__` or `vars` won't be directly usable, we will at least filter it.

You can use the functions [here](https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L61) to get the classes.

**Describe alternatives you've considered**
Approach 1:
This is the previous idea, which is discarded since we cannot generate all types without importing them. We have also edited the issue purpose a bit because of this.

While initializing the new data structure, we need a list of all the valid annotations, links and groups. For example, for annotations, we want to have a list of types as strings, ["Token", "Subword", "Sentence", ...], including all the types defined in ft.onto and ftx.onto. Please notice that users could also create their own types of data which inherit "forte.data.ontology.core.Entry".
For each type, we want to obtain all the attributes. For example, for sentence we want to get a list of ["speaker", "part_id", "sentiment", "classification", "classifications"]. We want to get the attributes for each type of entry as a dictionary. For example:
```
type_attributes = {
            "Token": ["pos", "ud_xpos", "lemma", "chunk", "ner", "sense", "is_root", "ud_features", "ud_misc"],
            "Document": ["document_class", "sentiment", "classifications"],
            "Sentence": ["speaker", "part_id", "sentiment", "classification", "classifications"],
        }
```

Approach 2:
[...]

PR #614, opened 2022-02-02
Inferring ontology tree automatically #583

### LINKED ISSUE #583: Inferring ontology tree automatically
The ontology tree is used in several places by Forte, it is mainly used to find all the entries (including subtype) of a top-level type, which is used in:
1. get entries of a particular type 
2. check output consistency between the components.

There could be other use of this piece of information. 

Currently, the ontology tree is loaded by [passing in the JSON file](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L125), which is optional.  There are a couple of issues with this approach. The JSON files may not be ready or synchronized in all environments, and the parameter being optional also makes it sometimes neglected. Also, there could be more than one JSON file making it a bit hard to find them all.

Since the entry classes are available in the python environment, it might be possible to infer them by finding all sub-classes of `Entry`.

We can recursively find subclasses from `Entry` via `__subclasses__`, and populate the `onto_specs_dict` and `merged_entry_tree` around [here](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L318). 

Currently, we haven't evaluated the problems of this approach.

**Additional context**
The solution here can be used to partially address issue #570.

Xuezhi-Liang:
This PR fixes #583 

### Description of changes
Recursive Method to Find Subclasses to generate the ontology tree,

### Test Conducted
Evaluated the Recursive Method in tests/forte/advanced_pipeline_test.py.


hunterhector:
Notice https://github.com/asyml/forte/issues/570 is an immediate follow up to this

qinzzz (commented):
I tried to run your `parse_entry` function on `Entry` class, and there are a few problems:
1) The returned "definitions" is not a complete list of all classes. For example, classes in ft.onto.wikipedia and ft.onto.metric are not found.  `ConferenceGroup` that inherit `Group` and multiple class that inherit `Link` in base_ontology is also not found.
Part of this problem is caused by skipping entrys that start with `TOP_MOST_MODULE_NAME`
2) The returned list of "additional_prefixes" contains duplicate `ftx.onto`.
3) The test case is too simple and cannot reflect actual use case. Please add test cases with real classes in forte, e.g. Entry. All classes in ontology_specs/ should be included in the returning dictionary.

hunterhector:
@feipenghe , @Xuezhi-Liang encountered a problem in building the doc, do you know what happen here?

feipenghe:
> @feipenghe , @Xuezhi-Liang encountered a problem in building the doc, do you know what happen here?

@Xuezhi-Liang  I think it's because hyperlink for "Handling Structured Text Data" expired. Since it's not very important and I might change the hyperlink later to a notebook, you can remove the hyperlink.

Xuezhi-Liang:
This issue is dependent on #issue570. I will recode this once #570 is solved.

hunterhector:
closed by https://github.com/asyml/forte/issues/583#issuecomment-1047413208

PR #616, opened 2022-02-03
Consider remove ontology definitions from pipeline representation serialization #584

### LINKED ISSUE #584: Consider remove ontology definitions from pipeline representation serialization
The current ontology serialized IR looks like the following:

```
components:
- configs:
    serialize_method: jsonpickle
    zip_pack: false
  type: forte.data.readers.string_reader.StringReader
- configs:
    overwrite: false
    user_input: bye
  selector:
    configs: {}
    type: forte.data.selector.DummySelector
  type: advanced_pipeline_test.UserSimulator
forte_ir_version: 0.0.1
states:
  attribute:
    _check_type_consistency: true
    _do_init_type_check: true
    _enable_profiling: true
    _initialized: true
  resource:
    merged_entry_tree:
      attributes: []
      children:
      - attributes: []
        children:
        - attributes:
          - lemma
          - ud_features
          - is_root
          - pos
          children: []
          name: ft.onto.base_ontology.Token
        - attributes: []
          children: []
          name: ft.onto.base_ontology.Document
        - attributes:
```
The `merged_entry_tree` section is quite long and may not really be necessary. Especially if we are able to infer the ontology tree via https://github.com/asyml/forte/issues/583, we can drop this section and make the IR to be more focused on its main objectives.

After we finish https://github.com/asyml/forte/issues/583, during the pipeline [save](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L324) function, we should drop this [part](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L436). Then we should make sure the [init_from_config](https://github.com/asyml/forte/blob/master/forte/pipeline.py#L313) works correctly with this new change.

**Additional context**

Xuezhi-Liang:
This PR fixes #584. 

### Description of changes
The merged_entry_tree section is quite long and may not really be necessary. Especially if we are able to infer the ontology tree via issue #583, we can drop this section and make the IR be more focused on its main objectives.

Depend on issue #583.

### Test Conducted
The test code is already ready, just check [here](https://github.com/asyml/forte/blob/d8d7e2a70e896648af4d445859b828e0c144f422/tests/forte/advanced_pipeline_test.py#L108).


hunterhector:
Since this pr depends on an unmerged one, will review it later

hunterhector:
It is quite confusing when this issue also tries to fix #583  when there is another PR here: https://github.com/asyml/forte/pull/641 also says so. This is not allowed. Please state clearly where will #583 gets solved. This can create problems such as accidentally closing issues.

Please learn this: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword, the PR message here is not just a convention, it will cause actual conseqeuences.

Issue #617, opened 2022-02-03
Update DataStore interface with Group and Link entry types


wanglec:
**Is your feature request related to a problem? Please describe.**
Develop skeleton code for group and link entries and related functions.

**Depends on**
- List issues that this one depends on.

**Describe the solution you'd like**
The interface requires a class definition with abstract methods, including add_link_raw, add_group_raw, next_entry, prev_entry.

**Describe alternatives you've considered**

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

PR #618, opened 2022-02-03
Add group and link skeleton code

wanglec:
This PR fixes https://github.com/asyml/forte/issues/617. 

### Description of changes
Implement skeleton functions for add_link_raw, add_group_raw and next/prev_entry .

### Test Conducted
Unit tests are conducted in tests/forte/data/data_store_test.py.


forte/data/data_store.py
```diff
@@ -315,50 +397,76 @@ def delete_entry(self, tid: int):
...
         Returns:
             An iterator of the entries matching the provided arguments.
-
         """
-        # We find the `type_id` according to `entry_type` and locate the
-        # sortedlist.
-        # We create an iterator to generate entries from the sortedlist.
+        # We find the `type_id` according to `entry_type` and locate the list.
+        # We create an iterator to generate entries from the list.
         yield from self.__elements[self.__type_dict[entry_type]]
```
hunterhector, line 446: entry_type can be str or EntryType, will this lookup work?

forte/data/data_store.py
```diff
@@ -211,15 +221,52 @@ def _new_annotation(self, entry_type: str, begin: int, end: int):
...
+        Args:
+            entry_type (str): Fully qualified name of this annotation.
+            parent_tid (int): Tid of the parent entry.
+            child_tid (int): Tid of the child entry.
+
+        Returns:
+            A list representing a new link type entry data.
+        """
+
+        tid: int = uuid.uuid4().int
```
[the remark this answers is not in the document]
wanglec: yes

forte/data/data_store.py
wanglec: Yes I add another function to convert class to str. Do you think it is better to only allow str in the typing?

hunterhector: Where is that function called? Here I saw `entry_type` is used directly, which means it should be str.

wanglec: It should be up now. I just pushed it. It is called `_get_entry_type_str`.

forte/data/data_store.py
```diff
@@ -188,10 +195,24 @@ def __init__(self, onto_file_path: Optional[str] = None):
...
+        4345314235: [parent_tid, child_tid, tid, entry_type, attr_1, ...,
+                    attr_n, index_id]}
         """
         self.__entry_dict: dict = {}
 
+    def _new_tid(self) -> int:
+        r"""This function generates a new `tid` for an entry."""
+        return uuid.uuid4().int
+
+    def _get_entry_type_str(
```
hunterhector, line 208: Consider implement this as a general utility method? Maybe here: https://github.com/asyml/forte/blob/master/forte/utils/utils.py

Also check that file to see if any utilities can be reused?

wanglec: I just found a similar function in utils. I will import that function directly and remove _get_entry_type_str.

forte/data/data_store.py
```diff
@@ -234,6 +285,41 @@ def add_annotation_raw(
...
+        r"""This function adds a group entry with `member_type` to the
+        `entry_type` list, returns the `tid` and the 2-D index for the
+        inserted entry.
+
+        Args:
+            entry_type (str): Fully qualified name of this group.
+            member_type (str): Fully qualified name of its members.
+
+        Returns:
+            `tid` of the entry and its index in the entry list.
```
hunterhector, line 318: Explain the index in the entry list better.

hunterhector: You should also create a function that takes the "index in the entry list" and return the actual data.

forte/data/data_store.py
```diff
@@ -313,52 +399,82 @@ def delete_entry(self, tid: int):
...
-        # entry from `__entry_dict` when its information is retrieved. We
-        # find the
-        # `entry_type` sortedlist using `type_id`. We bisect the sortedlist to
-        # find the index of entry data. Then, `type_id` and index are passed to
-        # _delete_entry_by_loc.
+        # `entry_type`, `type_id`. If it is an Annotation-Like entry, we obtain
+        # `begin` and `end` indices and use bisection to locate the `index_id`.
+        # If it is not, we obtain the 'index_id' directly from `__entry_dict`.
+        # We remove the entry from `__entry_dict` when its information is
+        # retrieved. Then, the `type_id` and the `index_id` are passed to
```
hunterhector, line 406: NO, given a tid we should always find it directly. Obtaining the `begin` and `end` through a list and do bisect is very indirect. Think about it, If you have a mapping to `begin` and `end` why can't we have a mapping to its position?

forte/data/data_store.py
```diff
@@ -313,52 +399,82 @@ def delete_entry(self, tid: int):
...
             tid (int): Unique id of the entry.
 
         Returns:
             The entry which `tid` corresponds to, its `type_id` and its index
-            in the `entry_type` sortedlist.
-
+            in the `entry_type` list.
         """
+        # If the entry is an annotation, bisect the annotation sortedlist
+        # to find the entry. May use LRU cache to optimize speed.
```
hunterhector, line 433: No, given tid, don't bisect

forte/data/data_store.py
hunterhector: You already have a function `get_entry` below, just call it from here.

forte/data/data_store.py
```diff
@@ -313,52 +399,82 @@ def delete_entry(self, tid: int):
...
+        # Otherwise, use `index_id` to find the index of the entry.
         raise NotImplementedError
 
     def get(
         self, entry_type: Union[str, Type[EntryType]], **kwargs
```
hunterhector, line 438: What is the kwargs for? it is not used

forte/data/data_store.py
```diff
@@ -313,52 +399,82 @@ def delete_entry(self, tid: int):
...
         """
-        # We find the `type_id` according to `entry_type` and locate the
-        # sortedlist.
-        # We create an iterator to generate entries from the sortedlist.
+        # We find the `type_id` according to `entry_type` and locate the list.
+        # We create an iterator to generate entries from the list.
+        if not isinstance(entry_type, str):
+            # convert `entry_type` to fully qualified name
+            entry_type = get_full_module_name(entry_type)
         yield from self.__elements[self.__type_dict[entry_type]]
```
hunterhector, line 454: This is actually oversimplifying things, did you consider the sub-classes?

hunterhector (commented):
The major problem is that we will have a O(1) way to lookup an entry loc based on the tid, there is no need to treat different types of entries differently. All entries need to be obtainable via this lookup in O(1).

forte/data/data_store.py
```diff
@@ -64,18 +65,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
         <attr_n>].
         A group type entry has the following format:
         [<member_type>, <[members_tid_list]>, <tid>, <entry_type>, <attr_1>,
-            <attr_2>, ..., <attr_n>].
+            <attr_2>, ..., <attr_n>, index_id].
```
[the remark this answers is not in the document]
wanglec: Yes in line 83 `It is an extra field to record the location of the entry in the list. When the user adds a new entry to the data store, the `index_id` will be created and appended to the end of the original `entry data` list.` 

I want to add an extra field in the data so that we can have the index when we have tid: 
entry_dict[tid][-1].

Then, we use this index in delete because we also need to remove data from the list.

forte/data/data_store.py
```diff
@@ -206,20 +220,57 @@ def _new_annotation(self, entry_type: str, begin: int, end: int):
...
         """
 
-        tid: int = uuid.uuid4().int
+        tid: int = self._new_tid()
         entry = [begin, end, tid, entry_type]
         entry += len(self._type_attributes[entry_type]) * [None]
         return entry
 
+    def _new_link(self, entry_type: str, parent_tid: int, child_tid: int):
+        r"""This function generates a new link with default fields.
```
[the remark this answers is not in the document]
wanglec: Yes

forte/data/data_store.py
```diff
@@ -206,20 +220,57 @@ def _new_annotation(self, entry_type: str, begin: int, end: int):
...
+        Returns:
+            A list representing a new link type entry data.
+        """
+
+        tid: int = self._new_tid()
+        entry = [parent_tid, child_tid, tid, entry_type]
+        entry += len(self._type_attributes[entry_type]) * [None]
+        return entry
+
+    def _new_group(self, entry_type: str, member_type: str):
```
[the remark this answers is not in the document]
wanglec: yes

forte/data/data_store.py
```diff
@@ -206,20 +220,57 @@ def _new_annotation(self, entry_type: str, begin: int, end: int):
...
+        r"""This function generates a new group with default fields.
+        Called by add_group_raw() to create a new group with
```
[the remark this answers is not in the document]
wanglec: yes

forte/data/data_store.py
```diff
@@ -234,6 +285,41 @@ def add_annotation_raw(
         # self.__entry_dict.
         raise NotImplementedError
 
+    def add_link_raw(
+        self, entry_type: str, parent_tid: int, child_tid: int
+    ) -> Tuple[int, int]:
+        r"""This function adds a link entry with `parent_tid` and `child_tid`
+        to the `entry_type` list, returns the `tid` and the 2-D index for the
```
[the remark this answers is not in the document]
wanglec: Yes I will add more info here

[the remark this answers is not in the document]
wanglec: I think it is unnecessary to return the index id in add_entry functions now, since we already store the index in entry data.

forte/data/data_store.py
wanglec: We may want to support: get within a `begin`, `end` range, etc.

hunterhector: Having `kwargs` will likely be a bad design since it means we don't have an idea of what we will support. In Forte we tend to avoid any kwargs

forte/data/data_store.py
wanglec: yes

forte/data/data_store.py
wanglec: yes

forte/data/data_store.py
wanglec: Sure, I will remove it.

forte/data/data_store.py
wanglec: Do you mean user-defined classes? `get_full_module_name` now works for classes defined in ft/ftx like 'ftx.onto.ag_news.Description', 'ft.onto.base_ontology.Token', etc .

hunterhector: No that one. Notice if you request a type, it is logical to also expect the sub-type of it also gets returned. 

We implemented similar stuff in data pack: https://github.com/asyml/forte/blob/master/forte/data/data_pack.py#L1365

So here we will need to use the parallel-iteration technique we discussed.

hunterhector: Is the plan to leave the implementation to another PR?

wanglec: I had a brute force way to implement it and it is already in the code. I think we can optimize after the issue #583 is done.

hunterhector: ok, once we merge this let's create an issue for that,

wanglec: yes sure

hunterhector (approved):

PR #619, opened 2022-02-03
Skeleton op - Lowest Complexity Data Augmentation Op

### LINKED ISSUE #621: Low Level Data Augmentation Op
Currently, the way to create new Data Augmentation techniques in Forte is not consistent. It can be created as a `Processor` as well as an `Op`. We want to have a consistent way to have all Data Augmentation techniques be just `Op`. The first component of this solution would be a low level DA Op where the users just need to provide the augmentation logic. This Op will also provide utility functions to help generate that logic. Once the logic is created by the user, this `Op` will process this logic to instantiate the augmentation and applying the changes suggested by the user.

The first step of the solution would be to create a Skeleton `Op` which would only require users to implement the augmentation logic. The utility functions (these can be methods to insert/delete/replace annotations) will be provided by the `Op` itself. This can set the tone for future `Ops` which could take the Skeleton Op further and make other DA frameworks with varying levels of complexity and freedom of computation. By doing so, we can set the basis for a consistent setup where all DA methods can be translated to `Op`.

Pushkar-Bhuse:
This PR fixes #621. 

### Description of changes
This is the first stage of migrating all data augmentation techniques in Forte to the `Op` format. Skeleton Op is the framework for by which users can implement their own augmentation logic by using inbuilt utility functions.

This would help in migrating current Data Augmentation techniques which are otherwise used as `Processors` to `Op`. This will also form the basis for higher level `Op` structures with greater simplicity but lesser freedom.

### Test Conducted
The tests included checking if the internal data structures of the `SkeletonOp` remained consisted before, during and after a data augmentation task. This was done by creating a dummy augmenter which inherited the `SkeletonOp`


hunterhector:
Please create a proper issue, and correctly uses `This PR fixes [issue link]`

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,250 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1:
```suggestion
# Copyright 2022 The Forte Authors. All Rights Reserved.
```

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,250 @@
...
+            span: Span = self._replaced_annos[pid][ind][0]
+            if not (span.begin >= end or span.end <= begin):
+                return True
+            if span.begin > end:
+                break
+            ind += 1
+
+        return False
+
+    def inserted_annotation_status(self):
```
hunterhector, line 120: let's annotate the return type on the function signature

hunterhector: If the purpose is to return one of the internal value, consider using `@property`?

hunterhector: Btw, is it necessary or good to expose internal states to users? We developers can still access these internal variables in test cases without this function

hunterhector (commented):
Only take a glance now,  have some simple comments. For the rest, let's discuss and talk about the details later.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
+# Copyright 202 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1:
```suggestion
# Copyright 2022 The Forte Authors. All Rights Reserved.
```

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+)
+from forte.data.multi_pack import MultiPack
+from forte.data.ontology.core import BaseLink, Entry
+from forte.utils.utils import create_class_with_kwargs, get_class
+
+
+__all__ = ["SkeletonOp"]
+
+
+class SkeletonOp(ABC):
```
hunterhector, line 55: Should extend https://github.com/asyml/forte/blob/master/forte/common/configurable.py so that `default_configs` will work properly and get inherited through the class hierarchy.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        self._data_pack_map: Dict[int, int] = {}
+
+        # :attr:`_entry_maps`: {datapack id: Dict{orig tid, new tid}}
+        # It is a map for tracking the annotation ids
+        # before and after the auto align. It maps the
+        # original annotation tid to the new annotation tid.
+        # It is used when copying the Link/Group/MultiPackLink/MultiPackGroup.
+        self._entry_maps: Dict[int, Dict[int, int]] = {}
+
+        self.configs = configs
```
hunterhector, line 105: Use `make_configs` from `Configurable` to merge provided configs with defulat configs

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+    Forte's internal setup.
+    """
+
+    def __init__(self, configs: Union[Config, Dict[str, Any]]) -> None:
+
+        # :attr:`_replaced_annos`: {datapack id: SortedList[span, new text]}
+        # It records the spans replaced by new texts.
+        # It is a map from datapack id to a list of tuples
+        # (span, new text) inserted by :func:`replace`.
+        # The new text will be used for building new data pack.
```
hunterhector, line 72: This sentence `The new text will be used for building new data pack` is slightly not clear.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        ] = defaultdict(lambda: SortedList([], key=lambda x: x[0]))
+
+        # :attr:`_inserted_annos_pos_len`: {datapack id: Dict{position: length}}
+        # It records the position and length of inserted spans,
+        # mapping from datapack id to a dictionary (position -> length)
+        # inserted by :func:`insert`.
+        # The position is the index in the original datapack
+        # of insertion, and the length is the length of the inserted string.
+        self._inserted_annos_pos_len: DefaultDict[
+            int, SortedDict[int, int]
```
hunterhector, line 84: why do we need to create a Dict with the position as the key here? I feel this will make it hard to compute overlap, as compared to `SortedList[Tuple[int, int]]`, where you can bisect it. Basically you can store the [begin, end] (position,position+length)

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        ] = defaultdict(lambda: SortedDict({}))
+
+        # :attr:`_deleted_annos_id`: {datapack id: Set[annotation tid]}
+        # It records the deleted span ids, mapping from datapack id
+        # to a set of annotation tids appended by :func:`delete`.
+        self._deleted_annos_id: DefaultDict[int, Set[int]] = defaultdict(set)
```
hunterhector, line 90: Similar to above, here we use a slightly different data structure again. Maybe we can also store `SortedList[Tuple[int, int]]` so that we can quickly find overlaps?.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        # original annotation tid to the new annotation tid.
+        # It is used when copying the Link/Group/MultiPackLink/MultiPackGroup.
+        self._entry_maps: Dict[int, Dict[int, int]] = {}
+
+        self.configs = configs
+
+    def perform_augmentation(self, input_pack: DataPack) -> DataPack:
+        r"""
+        Function to apply the defined augmentation function and
+        instantiating it into a data pack. This data pack is then
```
hunterhector, line 110: a new data pack right? should specify this in the docstring

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        returned.
+
+        Args:
+            data_pack: The Datapack holding the replaced annotations.
```
hunterhector, line 114: Have you try generating the docstring and take a look at the HTML? I feel the indent here is not correct.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+            replaced_annotations: A SortedList of tuples(span, new string).
+            The text and span of the annotations will be updated
+            with the new string.
+
+        Returns:
+            A new data_pack holds the text after replacement.
```
hunterhector, line 120:
```suggestion
            A new `data_pack` holds the text after replacement.
```

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        This function copies a Link/Group in the data pack.
+        If the children Link/Group does not exist, it will recursively
+        create the children Link/Group. If the children Annotation
+        does not exist, it will abort and return False.
+        Args:
+            entry: The Link/Group in the original data pack to copy.
+            entry_map: The dictionary mapping original entry to copied entry.
+            new_pack: The new data pack, which is the destination of copy.
+        Returns:
+            A bool value indicating whether the copy happens.
```
hunterhector, line 142: whether it is sucessful

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        """
+
+        # If the entry has been copied, return True.
+        if entry.tid in entry_map:
+            return True
+
+        # The entry should be either Link or Group.
+        is_link: bool = isinstance(entry, Link)
```
hunterhector, line 150: This didn't check for group, shalle we check for group and raise error otherwise?

forte/processors/data_augment/algorithms/skeleton_op.py
hunterhector, line 55: Btw, `SkeletonOp` might not be the best name? Something more down to the ground like `BaseOp` or `AbstractDAOp` (any names you like). And we also didn't have the term "data augmentation" in the class name.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        It could be used in tasks such as text generation, where
+        MultiPackLink is used to align the source and target.
+        Args:
+            entry: The MultiPackLink/MultiPackGroup to copy.
+            multi_pack: The multi_pack contains the input entry.
+        Returns:
+            A bool value indicating whether the copy happens.
+        """
+        # The entry should be either MultiPackLink or MultiPackGroup.
+        is_link: bool = isinstance(entry, BaseLink)
```
hunterhector, line 205: MIght check for `MultiPackLink` instead? also could throw exceptions for non-MultiPackLink, non-MultiPackGroup ones?

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+            child_pack: DataPack = child_entry.pack
+            child_pack_pid: int = child_pack.pack_id
+            # The new pack should be present.
+            if (
+                child_pack_pid not in self._data_pack_map
+                or child_pack_pid not in self._entry_maps
+            ):
+                return False
+            new_child_pack: DataPack = multi_pack.get_pack_at(
+                multi_pack.get_pack_index(self._data_pack_map[child_pack_pid])
```
hunterhector, line 224: I feels like this can be done via this func: 
https://github.com/asyml/forte/blob/1fac83be0b2d34d8adb995a2629e31743756e71b/forte/data/multi_pack.py#L275

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+            new_children.append(new_child_entry)
+
+        # Create the new entry and add to the multi pack.
+        new_entry: Entry
+        if is_link:
+            entry = cast(MultiPackLink, entry)
+            new_link_parent: Entry = new_children[0]
+            new_link_child: Entry = new_children[1]
+            new_entry = type(entry)(
+                multi_pack, new_link_parent, new_link_child  # type: ignore
```
hunterhector, line 242: I guess we can inline 239 240 here, but this is minor

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+    def modify_index(
+        self,
+        index: int,
+        # Both of the following spans should be SortedList.
+        # Use List to avoid typing errors.
+        old_spans: List[Span],
+        new_spans: List[Span],
+        is_begin: bool,
+        is_inclusive: bool,
+    ) -> int:
```
hunterhector, line 259: I assume you copy from the original code without modification? If so, I will skip reviewing this.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+            # Use the span's end index as anchor, if possible.
+            delta_index = (
+                new_spans[last_span_ind].end - old_spans[last_span_ind].end
+            )
+        return index + delta_index
+
+    def _apply_augmentations(
+        self,
+        data_pack: DataPack,
+    ) -> DataPack:
```
hunterhector, line 426: Simiarly, is there anything new here or mostly copied? I will skip reviewing this if they are copied.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        if len(self._replaced_annos[pid]) == 0:
+            return False
+        ind: int = (
+            bisect_left(self._replaced_annos[pid], (Span(begin, begin), "")) - 1
+        )
+        if ind < 0:
+            ind += 1
+
+        while ind < len(self._replaced_annos[pid]):
+            span: Span = self._replaced_annos[pid][ind][0]
```
hunterhector, line 653: hmm, so here we only check conflicts with "replacement", didn't check with "delete" and stuff?

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+            A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - `augment_entry`:
+                Defines the entry the processor will augment.
+                It should be a full qualified name of the entry class.
+                For example, "ft.onto.base_ontology.Sentence".
+        """
+        return {
+            "augment_entry": "ft.onto.base_ontology.Token",
+            "other_entry_policy": {},
```
hunterhector, line 764: We don't need a default value for `augment_entry` right? We could in fact create a default value for `other_entry_policy`. Here we also miss the documentation for `other_entry_policy`

tests/forte/processors/data_augment/algorithms/skeleton_op_test.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertEqual(list(delete_list[data_pack.pack_id]), [token_1.tid])
+
+        expectations = [
+            (Span(token_1.begin, token_1.end), ""),
+            (Span(token_1.end, token_1.end), "There"),
+            (Span(token_3.begin, token_3.end), "Last"),
+        ]
+
+        self.assertEqual(list(replace_list[data_pack.pack_id]), expectations)
+
```
hunterhector, line 101: The test cases didn't test for the exceptions.

Also, the code you adapted here that does all the deletion and insertion work, are coupled with a good set of test cases here: https://github.com/asyml/forte/blob/master/tests/forte/processors/base/data_augment_replacement_processor_test.py

What's your plan of keeping these test cases?

hunterhector (commented):
1. Many functions in the skeleton op are not private. But we should probably only expose a few, and keep the rest private. We should probably make only a few selected functions, including the utities to be public.
2. Looking at the codecov report: https://app.codecov.io/gh/asyml/forte/compare/619/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=asyml#D1L146, it seems like the coverage of the skeleton_op is fairly low (only 60%). What's the plan of adapting our existing test cases to here?

forte/processors/data_augment/algorithms/word_splitting_processor.py
```diff
@@ -98,51 +98,47 @@ def default_configs(cls):
         Returns:
             A dictionary with the default config for this processor.
             Additional keys for determining how many words will be split:
-            - alpha: 0 <= alpha <= 1. indicates the percent of the words
-            in a sentence that are changed.
-        Config Values:
-
-            - `other_entry_policy` (dict):
+        The keys to this dictionary are:
+            - alpha:
```
feipenghe, line 102: Several suggestions.
1. At the beginning of the list bulletpoint, there should be an empty line.
2. "-" should align with the start of the previous sentence on a vertical line, in this case "the"
3. "The keys to this dictionary are:" should have the same indentation level as the previous docstring. Or you might not even need this description as the previous docstring is clear enough. (So a simple fix for 2 and 3 is just delete "The keys to this dictionary are:")
4. You can use [building docs command](https://github.com/asyml/forte/blob/d3083beca858d7f254090c3e3185f78f2dfed110/.github/workflows/main.yml#L136) and see how you docstring look like in a html.

forte/processors/data_augment/base_data_augment_processor.py
```diff
@@ -58,20 +58,17 @@ def modify_index(
     r"""
     A helper function to map an index before replacement
     to the index after replacement.
-
```
feipenghe, line 61: It might be better to keep the empty lines there as they separate docstring paragraphs clearly.

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: Actually, this set up for recording deleted annotations is quite simple. Because when updating the new data pack, we just check that if the annotation which we are planning to add now to the data pack exists in the `deleted_annos_id` set or no. If it does, we just continue iterating, which essentially means that this annotation is not added.

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: This is useful when a new annotation is inserted. It helps to check if there has been an insertion before at that position in O(1) time. If there has, then it doesn't go through with recording that insertion. Else it will add that record in this data structure.

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: There is nothing new here

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: If you go below, when adding annotations to the data pack, we check if the annotations exist in the deleted_annos data structure. If yes, they are not added. Similarly, for inserted, we have a separate loop which handles inserted annotations.

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: I think `augment_entry` is an important default config since it determines on what type of ontology will the augmentation happen. Sorry about the `other_entry_policy documentation. I'll add it.

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+                break
+            ind += 1
+
+        return False
+
+    def insert_annotations(
+        self, inserted_text: str, data_pack: DataPack, pos: int
+    ) -> bool:
+        r"""
+        This is a utility function to insert a new annotation. After
```
hunterhector, line 666: Feels like this is not accurate `This is a utility function to insert a new annotation`. The function only insert some new text right?

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: What `_replaced_annos` basically records is all types of operations done on a text, be it insert, delete or replace. the operation of deleting just a part of an annotation can be replicated as a collection of one insert and one delete task (we first delete the annotation and insert the part of it that we want to keep). This maintains a consistent structure. The job of creating the final data pack is determined by the `_replaced_annos` data structure.

hunterhector: I think I understand a bit now, can you fix all the rest of the comments?

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        """
+        pid: int = input_anno.pack.pack_id
+        if self._overlap_with_existing(pid, input_anno.begin, input_anno.end):
+            return False
+
+        self._replaced_annos[pid].add((input_anno.span, ""))
+        self._deleted_annos_id[pid].add(input_anno.tid)
+        return True
+
+    def replace_annotations(
```
hunterhector, line 710: Do you feel there is a need to include `split_annotation` (like you did before) as another utility?

forte/processors/data_augment/algorithms/skeleton_op.py
```diff
@@ -0,0 +1,765 @@
...
+        self, replacement_anno: Annotation, replaced_text
+    ) -> bool:
+        r"""
+        This is a utility function to record specifically a replacement of an annotation.
+        With this function, an annotation can be replaced with another annotation.
```
hunterhector, line 715: `an annotation can be replaced with another annotation`, not really what this function is doing right? We are replacing the text of this annotation with the new text?

forte/processors/data_augment/algorithms/skeleton_op.py
hunterhector: And tbh, we don't know whether this op is only concerning one type of `augment_entry`, so having this `augment_entry` key here is probably meaningless, I think we can remove this key entirely.

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: It is using the same function

tests/forte/processors/data_augment/algorithms/skeleton_op_test.py
Pushkar-Bhuse: This could be done when we integrate the Base Data Augment Processor to use the Skeleton Op

forte/processors/data_augment/algorithms/skeleton_op.py
Pushkar-Bhuse: From the work I did in the `word_splitting` op, I feel that the operation of splitting an annotation can be done by the combination of an `insert` and `delete`. So that is why I feel we dont need a separate split utility.

tests/forte/processors/data_augment/algorithms/skeleton_op_test.py
Pushkar-Bhuse: The ones that needed the creation would take time. But for the rest, like `test_modify_index` and `test_multi_pack_copy_link_or_group`, I will be adding them to the next commit.

forte/processors/data_augment/algorithms/word_splitting_processor.py
Pushkar-Bhuse: Doing that gave me an unexpected indent error

feipenghe: It seems like the it has passed test cases. If you still have issue you can refer to this [edit](https://github.com/asyml/forte/pull/636/commits/fd89c925c96fe879bed46bf302c9b42bb4905874#diff-b522065264cdf426e37a70fb07d873a84c5e537aa20602de61e0c50efa861fccL179).

For other related issues, you can search keyword in this [PR](https://github.com/asyml/forte/pull/636/commits) for example fix.

hunterhector (approved):

PR #628, opened 2022-02-08
implement delete_entry and delete_entry_by_loc

qinzzz:
implement delete_entry and delete_entry_by_loc
change type in entry list to type_id; modify corresponding test case

This PR fixes #598. 

### Description of changes
implement delete_entry and delete_entry_by_loc

### Possible influences of this PR.
Change the type field in entry list from String to Integer;

### Test Conducted
add new tests in data_store_test; test valid and invalid delete operations.

PR #635, opened 2022-02-09
implement add_annotation_raw

### LINKED ISSUE #606: Add new annotations to DataStore
Implement `add_annotation_raw(self, type_id: int, begin: int, end: int)` in forte/forte/data/data_store.py.
This function adds an annotation entry with `begin` and `end` index to the sortedlist at index `type_id` of the array which records all sortedlists It returns the tid and index for the entry.

**Depends on**
Pull request #586

We should create the `entry data` with the format [begin, end, tid, type_id, ,attr_1, ..., attr_n]. Attr_1 to attr_n are attributes for this type and are different for each type. They are initialized as None value. A helper function _new_annotation() can be used to generate an annotation type entry data with default fields. A reference to the entry should be stored in both self.elements and self.entry_dict.

**Additional context**

qinzzz:
This PR fixes #606. 

### Description of changes
implement add_annotation_raw in DataStroe.

### Test Conducted
test_add_annotation_raw in data store test.


forte/data/data_store.py
```diff
@@ -119,6 +119,7 @@ def __init__(self, onto_file_path: Optional[str] = None):
         super().__init__()
         self.onto_file_path = onto_file_path
         self.__entry_type_idx = 3
+        self.TID_INDEX = 2
```
wanglec, line 122: We may use a file to include all constants and their descriptions.

hunterhector, line 122: similarly, consider keep the consts in another place

forte/data/data_store.py
```diff
@@ -309,7 +310,10 @@ def add_annotation_raw(self, type_id: int, begin: int, end: int) -> int:
         # annotation type entry data with default fields.
         # A reference to the entry should be store in both self.__elements and
         # self.__entry_dict.
-        raise NotImplementedError
+        entry = self._new_annotation(type_id, begin, end)
```
hunterhector, line 313: Now I feel there is a problem, where we store the `type_id` into the entry itself since we convert it to an integer.

What if we update the ontology with new types, will the `type_id` keep consistent? 

@wanglec

forte/data/data_store.py
```diff
@@ -309,7 +310,10 @@ def add_annotation_raw(self, type_id: int, begin: int, end: int) -> int:
...
+        self.__elements[type_id].add(entry)
```
hunterhector, line 314: I guess we need to handle some keyerror here (and test them).

forte/data/data_store.py
wanglec: I think we could handle updated ontology in `EntryTypeGenerator`. If we update ontology, we check the current dictionary which maps `type_id` and `entry_type`. If it exists in the dictionary, we do nothing. If it does not exist, we add it to the dictionary with a new `type_id`. Does this seem to work for you?

hunterhector: not sure yet. that actually means that in the generator we somehow need to ensure all the`type_id` are the same for the same time from last time (say some existing data pack serialization on disk). This sounds hard to manage unless we have a deterministic entry name to id generator

hunterhector: This is mainly related to the serialization/deserialization part. Maybe you can think about the workflow while deserializing a pack. There needs to be a `dict` after deserialization that maps from "entry type -> integer ids", and after that you deserialize this data pack, you need to make sure to align this mapping with the current mapping in memory. (i.e align the mapping from disk with the mapping from memory).

wanglec: Yes, do you mean to serialize this map somehow and use it when deserializing the data store?

hunterhector: Yes, we will need to consider this. We need to have a mechanism to ensure the ids mapped to the correct type during steps like serialization.

forte/data/data_store.py
```diff
@@ -118,7 +119,6 @@ def __init__(self, onto_file_path: Optional[str] = None):
         """
         super().__init__()
         self.onto_file_path = onto_file_path
-        self.__entry_type_idx = 3
 
         """
                 The `_type_attributes` is a private dictionary that provides
```
hunterhector, line 124: There is a doc spelling issue in the comment, let's fix those. Like `ints` -> `integers`

hunterhector, line 124: There is a doc spelling issue in the comment, let's fix those. Like `ints` -> `integers`

forte/data/data_store.py
```diff
@@ -244,9 +254,26 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        r"""This function takes a type_id and returns whether a type
+        is an annotation type or not.
+        Args:
+            type_name (str): The name of type in `self.__elements`.
+        Returns:
+            A boolean value whether this type_id belongs to an annotation
+            type or not.
+        """
+        entry_class = get_class(type_name)
+        return issubclass(entry_class, Annotation)
```
mylibrar, line 275: https://github.com/asyml/forte/pull/676#discussion_r824927567. Or maybe we can fix this issue in [#628 ](https://github.com/asyml/forte/pull/628#discussion_r825503299)

forte/data/data_store.py
```diff
@@ -370,34 +406,82 @@ def get_attr(self, tid: int, attr_id: int) -> Any:
...
+        # complexity: O(lgn)
+        # if it's annotation type, use bisect to find the index
+        if self._is_annotation(type_name):
+            entry_index = bisect_left(target_list, entry_data)
+        else:  # if it's group or link, use the index in entry_list
+            entry_index = entry_data[-1]
+
+        if (
+            entry_index >= len(target_list)
+            or target_list[entry_index] != entry_data
```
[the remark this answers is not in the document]
wanglec: seems to be unlikely to happen

forte/data/data_store.py
```diff
@@ -265,7 +292,10 @@ def add_annotation_raw(self, type_name: str, begin: int, end: int) -> int:
         # annotation type entry data with default fields.
         # A reference to the entry should be store in both self.__elements and
         # self.__entry_dict.
-        raise NotImplementedError
+        entry = self._new_annotation(type_name, begin, end)
+        self.__elements[type_name].add(entry)
```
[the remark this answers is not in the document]
wanglec: yes sure

forte/data/data_store.py
```diff
@@ -265,7 +292,10 @@ def add_annotation_raw(self, type_name: str, begin: int, end: int) -> int:
...
+        self.__entry_dict[entry[constants.TID_INDEX]] = entry
```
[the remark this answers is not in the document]
wanglec: yes

forte/data/data_store.py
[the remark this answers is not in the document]
qinzzz: In case the entry_index stored in entry_data is wrong?

forte/data/data_store.py
```diff
@@ -370,34 +406,82 @@ def get_attr(self, tid: int, attr_id: int) -> Any:
         Returns:
             The value of `attr_id` for the entry with `tid`.
         """
-        # We retrieve the entry data from `__entry_dict` using tid.
-        # We locate the attribute using `attr_id` and get the attribute.
         entry = self.__entry_dict[tid]
         return entry[attr_id]
 
     def delete_entry(self, tid: int):
```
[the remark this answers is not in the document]
qinzzz: Sure

forte/data/data_store.py
qinzzz: I will solve it in a new PR #687. This API will be kept.

forte/data/data_store.py
```diff
@@ -183,10 +182,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
     def _new_tid(self) -> int:
         r"""This function generates a new ``tid`` for an entry."""
         return uuid.uuid4().int
+    
+    def _add_new_type(self, type_name: str):
+        """
+        Add a new type input_entry_class into self._type_attributes.
+        This method will call get_entry_attribute_by_class(class)
+        Args:
+        input_entry_class: A string or class type representing a class
```
hunterhector, line 191: indentation

forte/data/data_store.py
```diff
@@ -183,10 +182,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
     def _new_tid(self) -> int:
         r"""This function generates a new ``tid`` for an entry."""
         return uuid.uuid4().int
+    
+    def _add_new_type(self, type_name: str):
+        """
+        Add a new type input_entry_class into self._type_attributes.
+        This method will call get_entry_attribute_by_class(class)
```
hunterhector, line 189: I didn't see this `get_entry_attribute_by_class` called.

forte/data/data_store.py
```diff
@@ -183,10 +182,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
     def _new_tid(self) -> int:
         r"""This function generates a new ``tid`` for an entry."""
         return uuid.uuid4().int
+    
+    def _add_new_type(self, type_name: str):
+        """
+        Add a new type input_entry_class into self._type_attributes.
```
hunterhector, line 188: Quote the variable names.

forte/data/data_store.py
```diff
@@ -183,10 +182,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
...
+        Args:
+        input_entry_class: A string or class type representing a class
+        """
+        # check if type is in dictionary
+        if type_name in self._type_attributes:
+            return
+        # get attribute dictionary
+        self._type_attributes[type_name] = {"attributes":{}, "parent": None}
+        self.__elements[type_name] = SortedList()
+        assert len(self.__elements) == len(self._type_attributes)
```
hunterhector, line 199: Don't use `assert`, raise correct exception.

forte/data/data_store.py
```diff
@@ -201,8 +215,15 @@ def _new_annotation(self, type_name: str, begin: int, end: int) -> List:
 
         tid: int = self._new_tid()
         entry: List[Any]
+
+        self._add_new_type(type_name)
         entry = [begin, end, tid, type_name]
-        entry += len(self._type_attributes[type_name]) * [None]
+        try:
+            entry += len(self._type_attributes[type_name]) * [None]
+        except KeyError as e:
```
hunterhector, line 223: add these except to the docstring.

forte/data/data_store.py
```diff
@@ -183,10 +182,25 @@ def __init__(self, onto_file_path: Optional[str] = None):
...
+        Add a new type input_entry_class into self._type_attributes.
+        This method will call get_entry_attribute_by_class(class)
+        Args:
+        input_entry_class: A string or class type representing a class
+        """
+        # check if type is in dictionary
+        if type_name in self._type_attributes:
+            return
+        # get attribute dictionary
+        self._type_attributes[type_name] = {"attributes":{}, "parent": None}
```
hunterhector, line 197: If I didn't misunderstand, this behavior is not correct.

If the `type_name` is not in `self._type_attributes`, we should check to see if we can import the specific class. Now it looks like it is just accepting a new type name, and assigning empty attributes.

hunterhector: Also, I think the function should also be designed to return the content from `self._type_attributes`, so you won't need to access it again (Like in line 252). And also use the EAFP pattern here.

forte/data/data_store.py
```diff
@@ -224,8 +245,15 @@ def _new_link(
 
         tid: int = self._new_tid()
         entry: List[Any]
+
+        self._add_new_type(type_name)
```
hunterhector, line 249: You should get the type attribute content from this call already, don't need to call again.

forte/data/data_store.py
```diff
@@ -243,14 +271,33 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
         entry = [member_type, [], tid, type_name]
-        entry += len(self._type_attributes[type_name]) * [None]
+        try:
+            entry += len(self._type_attributes[type_name]) * [None]
+        except KeyError as e:
+            raise KeyError(
+                f"Group type with id {type_name} does not exist."
+            ) from e
         return entry
```
hunterhector, line 283: I notice the same lines of code appear multiple times here in all the `_new_xx` function, I think we should merge them together.

forte/data/data_store.py
```diff
@@ -243,14 +271,33 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        r"""This function takes a type_id and returns whether a type
+        is an annotation type or not.
+        Args:
+            type_name (str): The name of type in `self.__elements`.
+        Returns:
+            A boolean value whether this type_id belongs to an annotation
+            type or not.
+        """
+        entry_class = get_class(type_name)
+        return issubclass(entry_class, Annotation)
```
hunterhector, line 295: Please add an issue/todo to make sure we optimize this function.

forte/data/data_store.py
```diff
@@ -307,23 +362,29 @@ def add_group_raw(
     def set_attribute(self, tid: int, attr_name: str, attr_value: Any):
         r"""This function locates the entry data with ``tid`` and sets its
         ``attr_name`` with `attr_value`. It first finds ``attr_id``  according
-        to ``attr_name``. `tid`, `attr_id`, and ``attr_value`` are
+        to ``attr_name``. ``tid``, ``attr_id``, and ``attr_value`` are
         passed to `set_attr()`.
+        Raises KeyError when ``tid`` or ``attr_name`` is not found.
```
hunterhector, line 367: Use the correct `raise` index.

forte/data/data_store.py
```diff
@@ -343,6 +402,7 @@ def get_attribute(self, tid: int, attr_name: str) -> Any:
         ``tid``. It locates the entry data with ``tid`` and finds `attr_id`
         of its attribute ``attr_name``. ``tid`` and ``attr_id``  are passed
         to ``get_attr()``.
+        Raises KeyError when ``tid`` or ``attr_name`` is not found.
```
hunterhector, line 405: Use the `raise` docstring syntax

forte/data/data_store.py
```diff
@@ -370,34 +435,78 @@ def get_attr(self, tid: int, attr_id: int) -> Any:
...
         """
-        # We retrieve the entry data from `__entry_dict` using tid.
-        # We locate the attribute using ``attr_id``  and get the attribute.
         entry = self.__entry_dict[tid]
         return entry[attr_id]
 
     def delete_entry(self, tid: int):
         r"""This function locates the entry data with ``tid`` and removes it
         from the data store. This function first removes it from `__entry_dict`.
+        Raises KeyError when entry with ``tid`` is not found. Raise RuntimeError
```
hunterhector, line 444: Use `raise` docstring syntax.

forte/data/data_store.py
```diff
@@ -370,34 +435,78 @@ def get_attr(self, tid: int, attr_id: int) -> Any:
...
+        try:
+            # get `entry data` and remove it from entry_dict
+            entry_data = self.__entry_dict.pop(tid)
+        except KeyError as e:
+            raise KeyError(
+                f"The specified tid [{tid}] "
+                f"does not correspond to an existing entry data "
+            ) from e
+
+        _, _, tid, type_name = entry_data[:4]
```
hunterhector, line 459: Use constant, not hard-coded 4

forte/data/data_store.py
```diff
@@ -184,6 +183,37 @@ def _new_tid(self) -> int:
...
+        if self.onto_file_path is not None:
+            raise RuntimeError(
+                f"{type_name} is not a valid type in Forte base class or"
+                f"provided by your ontology file"
+            )
+        # get attribute dictionary
+        attributes = EntryTypeGenerator.get_entry_attributes_by_class(type_name)
+
+        attr_dict = {}
+        idx = 4
```
hunterhector, line 205: consts

hunterhector:
Please resolve the conflicts and review all the previous conversations and comments.

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        attr_idx = constants.ENTRY_TYPE_INDEX + 1
+        for attr_name in attributes:
+            attr_dict[attr_name] = attr_idx
+            attr_idx += 1
+
+        self._type_attributes[type_name] = {
+            "attributes": attr_dict,
+            "parent_entry": None,
+        }
+        self.__elements[type_name] = SortedList()
```
hunterhector, line 232: Do we need to create the list now? This will mean there will be empty list in the `self.__elements` dict, which may make serialization awkward.

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
         r"""This function generates a new ``tid`` for an entry."""
         return uuid.uuid4().int
 
+    def _add_new_type(self, type_name: str):
```
hunterhector, line 200: return type missing

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+
+        self._type_attributes[type_name] = {
+            "attributes": attr_dict,
+            "parent_entry": None,
+        }
+        self.__elements[type_name] = SortedList()
+
+        return attr_dict
+
+    def _get_type_attribute_entry(self, type_name: str):
```
hunterhector, line 236: missing return type.

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        Raises:
+            KeyError: when ``type_name`` does not exist.
+        """
+        try:
+            type_attribute_entry = self._type_attributes[type_name]
+        except KeyError as e:
+            raise KeyError(f"Type {type_name} does not exist.") from e
+        return type_attribute_entry
+
+    def _get_entry_attribute_dict(self, type_name: str):
```
hunterhector, line 252: missing return type

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        """Get the attribute dict of an entry.
+        Args:
+            type_name (str): The fully qualified type name of the new entry.
+        Returns:
+            attr_dict (dict): The attributes-to-index dict of an entry.
+        """
+        return self._get_type_attribute_entry(type_name)["attributes"]
+
+    def _get_entry_parent(self, type_name: str):
```
hunterhector, line 261: missing return type

forte/data/data_store.py
```diff
@@ -283,7 +371,15 @@ def add_annotation_raw(self, type_name: str, begin: int, end: int) -> int:
         # annotation type entry data with default fields.
         # A reference to the entry should be store in both self.__elements and
         # self.__entry_dict.
-        raise NotImplementedError
+        entry = self._new_annotation(type_name, begin, end)
+        try:
+            self.__elements[type_name].add(entry)
+        except KeyError:
+            self.__elements[type_name] = SortedList(key=lambda s: (s[0], s[1]))
```
hunterhector, line 378: So you have automatically creating the list here, I feel like there is no need to do this again in the `_add_new_type` function

forte/data/data_store.py
hunterhector, line 200: Further the name is a bit misleading. It is more like `_get_type_info`, since we never really add a new type, they are either in the dict or in python class.

forte/data/data_store.py
qinzzz: Did that in line 433. Forget to delete this line

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
         r"""This function generates a new ``tid`` for an entry."""
         return uuid.uuid4().int
 
+    def _get_type_info(self, type_name: str) -> Dict[str, int]:
+        """
+        Add a new type input_entry_class into ``_type_attributes``.
```
hunterhector, line 202: docstring is off (not really `add a new type`). The main purpose is to get the type, `add a type` is kinda a side-effect (and the user don't need to know that something is added).

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        This method will get all attributes of a type and assign index to them.
```
hunterhector, line 203: Similar to above, when describing the function, `assign index` is not important. You can add a second paragraph providing some implementation details, but if you want to do that, it needs to be done in much more details. For example, "assign index" is to vague

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        Args:
+            type_name (str): A string or class type representing a class.
```
hunterhector, line 205: "or class type"? did we take that? if so, the type annotation is wrong. 

and then, remember we should use "full qualified name"

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+        Returns:
+            attr_dict (dict): The attribute dict of the given type.
```
hunterhector, line 207: what's the format of the "attribute dict"? Docstring needs to provide that.

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+            RuntimeError: when the type is not provided by ontology file and
+            dynamic import is disabled.
+        """
+        # check if type is in dictionary
+        if type_name in self._type_attributes:
+            return self._type_attributes[type_name]
+        if not self._dynamically_add_type:
+            raise RuntimeError(
+                f"{type_name} is not an existing type in current data store."
+                f"Dynamically add type is disabled."
```
hunterhector, line 218: As an exception message, it should explain what to do (which boolean needs to be turned on). Further, `ValueError` might be more appropriate.

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+            attr_idx += 1
+
+        self._type_attributes[type_name] = {
+            "attributes": attr_dict,
+            "parent_entry": None,
+        }
+
+        return attr_dict
+
+    def _get_type_attribute_info(self, type_name: str) -> Dict[str, Any]:
```
hunterhector, line 236: what's the purpose of this function? is it a weaker version of `_get_type_info`? This function also ignores dynamic typing. should we not use this?

forte/data/data_store.py
```diff
@@ -204,8 +284,11 @@ def _new_annotation(self, type_name: str, begin: int, end: int) -> List:
 
         tid: int = self._new_tid()
         entry: List[Any]
+
+        entry_attr_dict = self._get_type_info(type_name)
         entry = [begin, end, tid, type_name]
-        entry += len(self._type_attributes[type_name]) * [None]
+        entry += len(entry_attr_dict) * [None]
```
hunterhector, line 290: `num_attributes_for_type` could be a useful function to add.

tests/forte/data/data_store_test.py
```diff
@@ -248,20 +247,47 @@ def setUp(self) -> None:
...
+        self.assertEqual(empty_data_store._type_attributes["ft.onto.base_ontology.Document"],
+                        self.data_store._type_attributes["ft.onto.base_ontology.Document"])
+        # test the return value
+        self.assertEqual(doc_attr_dict, empty_data_store._type_attributes["ft.onto.base_ontology.Document"]["attributes"])
+        # test add invalid type
+        with self.assertRaisesRegex(
+            ValueError, "Class not found in invalid.Type"
+        ):
+            self.data_store._get_type_info("invalid.Type")
+        self.assertTrue("invalid.Type" not in self.data_store._type_attributes)
```
hunterhector, line 269: We will need more tests for dealing with types in python class and ontology files (such as missing type in both, type appear in both, and other corner cases)

forte/data/data_store.py
hunterhector: There are a couple of counter-arguments:
1.  This creates multiple `get_info` api calls, making it confusing which one to call.
2. Can we really ensure the dynamic-type can be done before call thing this function?
3. Even if 2 is ensured, this function will be as fast as `_get_type_info`, so what's the benefit of having this?

forte/data/data_store.py
```diff
@@ -187,6 +197,76 @@ def _new_tid(self) -> int:
...
+
+        attr_dict = {}
+        attr_idx = constants.ENTRY_TYPE_INDEX + 1
+        for attr_name in attributes:
+            attr_dict[attr_name] = attr_idx
+            attr_idx += 1
+
+        self._type_attributes[type_name] = {
+            "attributes": attr_dict,
+            "parent_entry": None,
```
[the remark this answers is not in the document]
qinzzz: 1. Considering efficiency, I believe it's better to initialize as None by default and fill it when needed.
2. Yes, I designed those two functions to have different purposes.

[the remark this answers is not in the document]
wanglec: line 214 and 234 are both in function `_get_type_info`. Shouldn't they contain the same format of output?

forte/data/data_store.py
wanglec: @qinzzz Indeed, it seems like the only benefit here is they could provide different error messages.

forte/data/data_store.py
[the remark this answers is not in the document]
qinzzz: Oh I see, I will change line214 to return attr_dict

forte/data/data_store.py
qinzzz: Previously I thought set and get attribute function should only get and not add new types, so i used a new function to pre vent them from adding new types. But yes, the new API does not actually provide benefits, since adding a new type if it does not exist is still a reasonable behaviour.

tests/forte/data/data_store_test.py
qinzzz: The parse ontology file function is not merged yet, so I guess we cannot test that part for now

hunterhector: yeah, just try to make sure we remember to test these.

forte/data/data_store.py
```diff
@@ -190,6 +200,104 @@ def _new_tid(self) -> int:
...
+
+        new_entry_info = {
+            "attributes": attr_dict,
+            "parent_entry": set(),
+        }
+        self._type_attributes[type_name] = new_entry_info
+
+        return new_entry_info
+
+    def _get_type_attribute_dict(self, type_name: str) -> Dict[str, int]:
```
hunterhector, line 262: I don't think we really need this function

tests/forte/data/data_store_test.py
hunterhector: need to create issues to track these otherwise they may be forgotten.

forte/data/data_store.py
```diff
@@ -381,16 +503,17 @@ def get_attribute(self, tid: int, attr_name: str) -> Any:
...
         try:
-            attr_id = self._type_attributes[entry_type][attr_name]
+            attr_id = self._get_type_attribute_dict(entry_type)[attr_name]
         except KeyError as e:
             raise KeyError(f"{entry_type} has no {attr_name} attribute.") from e
 
-        return self._get_attr(tid, attr_id)
+        return entry[attr_id]
 
     def _get_attr(self, tid: int, attr_id: int) -> Any:
```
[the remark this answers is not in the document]
qinzzz: Previously we decided not to call _get_attr mainly for efficiency but can leave it for future use.

hunterhector (approved):

PR #641, opened 2022-02-14
Generate entry names and attribute structure from python class hierarchy #570

### LINKED ISSUE #570: Generate and attribute structure from a forte entry class based on the fully qualified name

Xuezhi-Liang:
This PR fixes #570, 

### Description of changes

### Test Conducted


hunterhector:
So it is generally not a good idea to fix two issues in one PR.

tests/forte/advanced_pipeline_test.py
```diff
@@ -280,5 +280,76 @@ def tearDown(self) -> None:
...
+            "definitions"
+        ]
+        definitions_2 = self.pipeline_2.resource.get("onto_specs_dict")[
+            "definitions"
+        ]
+        retDifference = [i for i in definitions_2 if i not in definitions_1]
+        retDifference_expected_union = [
+            i for i in self.expected if i in retDifference
+        ]
+        # self.assertEqual(len(self.expected), len(retDifference_expected_union))
```
qinzzz, line 351: This test does not test anything -- If you want to skip a test, use @unittest.skip;

tests/forte/advanced_pipeline_test.py
```diff
@@ -280,5 +280,76 @@ def tearDown(self) -> None:
...
+            {
+                "entry_name": "ft.onto.wikipedia.WikiAnchor",
+                "parent_entry": "forte.data.ontology.top.Annotation",
+            },
+        ]
+
+        self.pipeline_1: Pipeline = Pipeline()
+        from ft.onto.wikipedia import WikiPage
+
+        self.pipeline_2: Pipeline = Pipeline()
```
qinzzz, line 335: I did not understand the purpose of this test. Why does pipeline_1 and pipeline_2 be different?

qinzzz (commented):
Please reuse your previous PR for the same issue instead of opening a new one.

tests/forte/data/entry_type_generator_test.py
```diff
@@ -1,19 +1,31 @@
...
-        self.assertLessEqual(mock_get_type_attributes.call_count, 1)
-        EntryTypeGenerator.get_type_attributes()
-        self.assertLessEqual(mock_get_type_attributes.call_count, 1)
+        type_attributes = EntryTypeGenerator.get_type_attributes()
+        self.assertLessEqual(len(type_attributes['Annotation']['Token']), 9)
+        self.assertLessEqual(len(type_attributes['Annotation']['Sentence']), 5)
+        # type_attributes = self.entry_type_generator.get_type_attributes()
+        # self.assertLessEqual(mock_get_type_attributes.call_count, 1)
+        # EntryTypeGenerator.get_type_attributes()
+        # self.assertLessEqual(mock_get_type_attributes.call_count, 1)
```
qinzzz, line 28: Do not comment out the old test, just write a new function.

qinzzz: This test fails. Please fix and recommit (reuse this PR)

feipenghe:
I have fixed the documentation issues in this [PR](https://github.com/asyml/forte/commit/0c1dec1311f27eae150287a8aa405632b265e03e#diff-8aa2183d412dbc2206045adf49d10dfc10c416aea9267594ac7525e9c90b9078L196) and it's merged into master branch. You can try checkout this file from master branch to fix this. This issue is due to updated version of myst parser.

Xuezhi-Liang:
All fixed. Please help review it again. Thanks. @qinzzz

tests/forte/data/entry_type_generator_test.py
```diff
@@ -1,10 +1,17 @@
 import unittest
 from unittest.mock import patch
 from forte.data.entry_type_generator import EntryTypeGenerator
+from ft.onto.base_ontology import Token
+from ft.onto.wikipedia import WikiPage
```
hunterhector, line 5: Before a full review from qinxin, I would like to ask a question.

I heard that we have to import a type here to make it appear in the generated list. Can we get around this? For our use case, it is impossible to import them beforehand.

hunterhector: In the comment section of the highest voted answer here: https://stackoverflow.com/questions/3862310/how-to-find-all-the-subclasses-of-a-class-given-its-name, they have mentioned a method.

hunterhector: let us know if you find this impossible though, we will consider alternate approaches.

Xuezhi-Liang: OK， I will try to find it is possible to get around this.

Xuezhi-Liang: > In the comment section of the highest voted answer here: https://stackoverflow.com/questions/3862310/how-to-find-all-the-subclasses-of-a-class-given-its-name, they have mentioned a method.

I also used this method to generate the ontology tree.  But please note if the class definition of a subclass hasn't been executed yet - for example, if the subclass's module hasn't been imported yet - then that subclass doesn't exist yet, and __subclasses__ won't find it.

Xuezhi-Liang: OK. I have attached two screenshots of the test result for easy viewing.  

image
image

Xuezhi-Liang: > 

Also,  we do not need to import all the types, Like the below images, we just import Token, all classes that belong to base_ontology are appear in the generated list. 
image

forte/pipeline.py
```diff
@@ -222,6 +222,64 @@ def __init__(
...
+        if len(entry_subclass.__subclasses__()) > 0:
+            all_attributes[entry_subclass.__name__] = {}
+        parent_entry = entry_subclass.__module__ + "." + entry_subclass.__name__
+        for subclass in entry_subclass.__subclasses__():
+            entry_name = subclass.__module__ + "." + subclass.__name__
+            if (
+                not len(entry_name.split(".")) < 3
+                and not len(parent_entry.split(".")) < 3
+            ):
+                if "__annotations__" in subclass.__dict__.keys():
```
qinzzz, line 267: Use try..except to save time

forte/pipeline.py
```diff
@@ -222,6 +222,64 @@ def __init__(
...
+        all_subclasses = []
+        all_attributes: Dict = {}
+        if len(entry_subclass.__subclasses__()) > 0:
+            all_attributes[entry_subclass.__name__] = {}
+        parent_entry = entry_subclass.__module__ + "." + entry_subclass.__name__
+        for subclass in entry_subclass.__subclasses__():
+            entry_name = subclass.__module__ + "." + subclass.__name__
+            if (
+                not len(entry_name.split(".")) < 3
+                and not len(parent_entry.split(".")) < 3
```
qinzzz, line 265: Why use hard coding 3 here?

forte/pipeline.py
```diff
@@ -222,6 +222,64 @@ def __init__(
         # Indicate whether do type checking during pipeline initialization
         self._do_init_type_check: bool = do_init_type_check
 
+    def find_entry_inherit(self, entry, filter_list: List[str]):
+        r"""Find all sub_classes of Entry Class.
```
qinzzz, line 226: This class can not only find inherit for `Entry`, but also other base classes.

tests/forte/data/entry_type_generator_test.py
hunterhector: So based on xuezhi's investigation (thanks on that!). We will have to import at least something that contains the ontology class.

However, this will make this function not match our expectations.  Since we have no way to ensure we import every possible ontology class before running this function. So we may consider using an alternative approach to solve this problem, which does not require we import all classes upfront.

Thanks to Xuezhi on this, if we decided to use the alternate approach, we can consider this issue solved as not-fixable.

We will still work on the attribute structure given a entry name, but that will be a slightly different issue.

forte/pipeline.py
Xuezhi-Liang: If you have to import all realted types, import them in pipeline.py instead of the test file. You may not want users to import all classes every time they use this function. I tested locally and this should work.------------ At first, for user convenience, I imported them to pipeline.py. Yes, it works locally, but it could not pass the PyLint check in the PR, as shown in the figure below, so I imported them in the test case.
image

tests/forte/data/entry_type_generator_test.py
Xuezhi-Liang: Thanks, @hunterhector， Once the team has the final decision, please let me know, as #583 and #584 meet the same problems.

Xuezhi-Liang:
Closed this PR as the description changed. Will open a new PR.

PR #648, opened 2022-02-19
Ontology namespace consistency

### LINKED ISSUE #647: Ensure Ontology namespace consistency
**Describe the bug**
In reference to this [issue](https://github.com/asyml/forte-medical/issues/1) and comments from [this](https://github.com/asyml/forte-medical/pull/4) PR,
We want to be building ontologies in appropriate namespaces and want the structure to be consistent across all forte repos, to ensure disambiguation. 

The structure should be of the form 
```
- forte
- ftx 
  |_ medical
      |_ clinical.py
  |_ news
      |_ ag_news.py
  |_ mcq
      |_ race_qa.py
```

Also, this should also be reflected in `setup.py` as so:
`packages=find_namespace_packages(include=['ftx.*'])`

Piyush13y:
This PR fixes #647 

### Description of changes

- The structure of the package namespaces updated using ftx folder. 
- `setup.py` updated as well
- All usages of ftx.* packages updated using appropriate namespaces.

Should generate ontologies in the expected format and not cause conflicts.


forte/ontology_specs/race_qa.json
```diff
@@ -4,12 +4,12 @@
     "base_ontology.json"
   ],
   "additional_prefixes": [
-    "ftx.onto.race_qa"
+    "ftx.mcqs"
```
hunterhector, line 7: I feel like we can keep these to be `ftx.onto` if they don't conflict with our projects? Making fewer changes might be easier for us.

forte/ontology_specs/medical.json
```diff
@@ -4,11 +4,11 @@
     "base_ontology.json"
   ],
   "additional_prefixes": [
-    "ftx.medical"
+    "ftx.medical.clinical"
```
hunterhector, line 7: so we would probably want to make sure we don't have the same module defined here and forte-medical. Since the code here is for demo/example purposes, we may use a package path like `ftx.demo.clinical` or `ftx.onto.clinical`.

forte/ontology_specs/ag_news.json
```diff
@@ -4,11 +4,11 @@
       "base_ontology.json"
     ],
     "additional_prefixes": [
-        "ftx"
+        "ftx.news"
     ],
     "definitions": [
       {
-        "entry_name": "ftx.onto.ag_news.Description",
+        "entry_name": "ftx.news.ag_news.Description",
```
hunterhector, line 11: probably no need to create a subpackage name for each example, we can put them all under `ftx.onto`?

setup.py
```diff
@@ -17,7 +17,8 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     license="Apache License Version 2.0",
-    packages=setuptools.find_packages(
+    packages=setuptools.find_namespace_packages(
```
hunterhector, line 20: Wait, this probably will not work. The original `find_packages` finds packages using `__init__` file. I am wondering whether this change will omit our core source folder `forte`.

hunterhector (commented):
I am not sure if the new find packages command is correct. Have you checked their behaviors before and after (i.e. check the output of `packages`)?

forte/ontology_specs/race_qa.json
Piyush13y: Makes sense, reverted these changes. Now, building all example ontologies in `ftx.onto.*`

setup.py
Piyush13y: I tried reinstalling `forte` with the updated `setup.py` file and checked the site-packages that got installed.
I could see forte root source folder in addition to ft and ftx (Not sure if this is what you meant though)

Screen Shot 2022-02-19 at 9 58 12 PM

Also, wouldn't our build pipeline on git throw an error too for most of our import statements in the source code, if it were to skip forte's root source folder?

hunterhector: The CI pipeline fails for other reasons (due to our recent updates on Texar-pytorch), so it did fail so I don't know what happen to the full test.

I am not familiar with the behavior of `find_namespace_packages` when both types of packages exist so I have to be careful here. Would be great if we could find supporting documentation for this.

hunterhector: By updating the master the pipeline seems to work, this is good news but let's be a little cautious and confirm the actual behavior of `find_namespace_packages`.

Piyush13y: [This](https://discuss.python.org/t/mixing-namespace-and-non-namespace-packages-bad-practice/7536/11) thread discusses a similar scenario I think. It says packaging works if we name and situate our packages correctly, but some devs suggest its not a best practice. Couldn't find much intricate documentation on namespace packaging, it's mostly vague. Let me know what you think.

hunterhector: The thread you refer to is a different scenario, which is a scenario that we should avoid. 

In that thread, there is a package named `myproject` with an `__init__.py`, but he also wants to use `myproject. datagen` as a namespace package. This is not advised because if you install two projects using different methods, it will not work.

In our case, we plan to have `ftx` in multiple places, and we don't want to have an `__init__.py` at `ftx` level.  So this is a standard namespace packaging setup. This will work correctly.  We also have another package `forte`, which won't conflict with `ftx` anyways.

My question is that: what's the behavior of `find_namespace_packages`? Our previous approach is `find_packages` and we know that it will try to find all packages by looking for `__init__` files. We want to know whether `find_namespace_packages` will do the same thing and then in addition to that, find the namespace packages passed through the argument. In other words, is `find_namespace_packages` doing `find_packages`'s job and some additional work.

Piyush13y: Regarding that, setup tools documentation says `find_namespace_packages` is backwards compatible to `find_packages`. Hence, that should imply it does all that `find_packages` as a utility does and more. 

[Reference](https://setuptools.pypa.io/en/latest/setuptools.html)

hunterhector: thanks for the reference, I didn't find it myself. it should be good now.

hunterhector (approved):

PR #657, opened 2022-02-24
Pipeline documentation

feipenghe:
This PR provides a documentation for building NLP pipeline.

### Description of changes
* Add reader and processor documentations.
* Change Section to Caption so that subsections are shown on the sphinx documentation left bar. 
* Export jupyter notebook to rst so that sphinx documentation can directly render the notebook (rather than using extra packages)

### Test Conducted


docs/toc/processor.rst
```diff
@@ -0,0 +1,95 @@
...
+==========
+
+A pipeline component that wraps inference model and set up inference related work.
+
+
+
+Usage
+------
+
+* wrap model by setting `self.model`
```
hunterhector, line 11: There is no requirement to use `self.model`, this is all up to the implementation.

docs/toc/processor.rst
```diff
@@ -0,0 +1,95 @@
...
+* set up inference related behaviors
+
+
+Functions
+------------------
+
+* `initialize()`: initilize model and inference related attributes.
```
hunterhector, line 18: `model` or `inference` are not the required parts of a processor. From the framework perspective, `initialize` is a step that's run when a pipeline is initialized before any data (datapacks) is processed. That's the message we are going to provide.

docs/toc/processor.rst
```diff
@@ -0,0 +1,95 @@
...
+* `process()`: using loaded model to make predictions and write the prediction results out
```
hunterhector, line 19: Again, `model`, `prediction`, `write prediction` are all not essential parts of `process`. The `process` function has access to the data pack, and the `process` function is called at certain stages during the pipeline. Those are the important information.

hunterhector: Please try to understand the lifecycle of a pipeline, when are stuff called, and then come back to explain them.

docs/toc/processor.rst
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+
+Processor Class Hierarchy
+------------------------------
+
+Here we provide a simplified class hierarchy for :class:`PlainTextReader` to show the relations of processors which are subclasses of `PipelineComponent`.
```
hunterhector, line 27: Why showing a hierarchy? And this is also not a hierarchy for readers

docs/toc/reader.rst
```diff
@@ -0,0 +1,142 @@
+Reader
+=======
+
+A pipeline component that reads data from a file path into a data iterator.
```
hunterhector, line 4: No, reader is not for "file path" only.

docs/toc/reader.rst
```diff
@@ -0,0 +1,142 @@
...
+
+Usage
+------
+
+* Read data from a path and load it into a data iterator. It returns one datum at a time.
```
hunterhector, line 9: what do you mean by return a datum a time. what's the definition of a datum, and what's the definition of `a time`

docs/toc/reader.rst
```diff
@@ -0,0 +1,142 @@
...
+
+            Returns: Iterator over paths to .txt files
+            """
+            return dataset_path_iterator(text_directory, self.configs.file_ext)
+
+        def _cache_key_function(self, text_file: str) -> str:
+            return os.path.basename(text_file)
+
+        # pylint: disable=unused-argument
+        def text_replace_operation(self, text: str):
```
hunterhector, line 85: As an tutorial, you should explain these functions, including `text_replace_operation`

docs/toc/reader.rst
```diff
@@ -0,0 +1,142 @@
...
+            """
+            record_meta["ft.onto.base_ontology.Document"] = set()
+
+
+Reader Class Hierarchy
+------------------------
+
+Here we provide a simplified class hierarchy for `PlainTextReader` to show the relations of readers which are subclasses of `PipelineComponent`.
+
+* `PipelineComponent`: As the hierarchy suggests, readers are subclasses of ~PipelineComponent
```
hunterhector, line 122: Again, there is no need to explain this now. You should probably focus on explaining why or when to use `PackReader`/`MultiPackReader`.

hunterhector (commented):
The understanding is basically wrong, need several revisions

docs/toc/processor.rst
feipenghe: It's supposed to be `MachineTranslationProcessor`. The hierarchy is to show it's essentially a `PipelineComponent`, and how to decide what processor to inherit when users need to customize their own.

hunterhector: This should probably be explained better, a class hierarchy would be too boring for a reader

hunterhector (approved):

PR #659, opened 2022-02-28
Generate and attribute structure from a forte entry class based on the fully qualified name #570

### LINKED ISSUE #570: Generate and attribute structure from a forte entry class based on the fully qualified name

Xuezhi-Liang:
This PR fixes #570 

### Description of changes

### Test Conducted


forte/data/entry_type_generator.py
```diff
@@ -1,4 +1,5 @@
 import functools
```
hunterhector, line 1: Let's include the apache license

forte/data/entry_type_generator.py
```diff
@@ -10,6 +11,15 @@ def _get_type_attributes():
     return type_attributes
 
 
+def _get_entry_attribute_by_class(input_entry_class: str):
+    modname, _, clsname = input_entry_class.rpartition(".")
```
hunterhector, line 15: Maybe we should use the utilities [here](https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L79) to get class and import, which would help future debugs:

forte/data/entry_type_generator.py
```diff
@@ -10,6 +11,15 @@ def _get_type_attributes():
     return type_attributes
 
 
+def _get_entry_attribute_by_class(input_entry_class: str):
```
hunterhector, line 14: though docstring for private classes is not required, still helpful to have for clarity

forte/data/entry_type_generator.py
```diff
@@ -10,6 +11,15 @@ def _get_type_attributes():
...
+    modname, _, clsname = input_entry_class.rpartition(".")
+    mod = importlib.import_module(modname)
+    try:
+        return list(getattr(mod, clsname).__annotations__.keys())
```
hunterhector, line 18: So we are relying on the class annotations to find the attributes? This might be OK, but this means we need to enforce all entry classes are generated from our `JSON` files (previously, an alternative way is that one can actually modify the classes). 

That being said, this solution may still be OK. I just would like to confirm is this the only way to do this? If this is the case, we need to make a note to make sure our documentation enforce typing very strictly.

forte/data/entry_type_generator.py
```diff
@@ -10,6 +11,15 @@ def _get_type_attributes():
...
+    except AttributeError:
+        return []
+
+
 class EntryTypeGenerator:
```
hunterhector, line 23: Need class level docstring.

forte/data/entry_type_generator.py
```diff
@@ -10,6 +11,15 @@ def _get_type_attributes():
...
     def __init__(self):
```
hunterhector, line 24: Unused `__init__`, maybe delete it.

forte/data/entry_type_generator.py
```diff
@@ -39,3 +49,25 @@ def get_type_attributes():
```
hunterhector, line 49: 1. Missing return type.
2. I think we need to rewrite the docstring a bit. First, we need to make sure we include  `Return`. And we also need to explain the behavior of this function. For example, it might be important that we mention this actually relies on type annotation.

forte/data/entry_type_generator.py
```diff
@@ -39,3 +49,25 @@ def get_type_attributes():
 
         """
         return _get_type_attributes()
+
+    @staticmethod
+    def get_entry_attribute_by_class(input_entry_class: str):
```
hunterhector, line 54: 1. missing return type
2. wrong documentation format, missing arguments, missing return types.  
3. To be more accurate, we should probably write the docstring with [doctest format](https://docs.python.org/3/library/doctest.html) (which will ensure the documentation will also be tested!) , i.e. using `>>>` and return values. We often use doctest for utilities, like [here](https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L240).

tests/forte/data/entry_type_generator_test.py
```diff
@@ -15,5 +16,37 @@ def test_get_type_attributes(self, mock_get_type_attributes):
...
+            "ft.onto.base_ontology.Title": [],
+            "ft.onto.metric.SingleMetric": ["value"]
+        }
+        for entry_name in entry_name_attributes_dict.keys():
+            attribute_result = EntryTypeGenerator.get_entry_attribute_by_class(
+                entry_name
+            )
+            self.assertEqual(
+                attribute_result, entry_name_attributes_dict[entry_name]
+            )
```
hunterhector, line 48: So there is nothing particularly wrong with the test cases. However, if one day we updated the ontology file, this test may fail.

One idea is that you can create some dummy example ontology class here (for example, copy the `Sentence` class here but rename it). That way we can test on those classes.

hunterhector:
btw, when we are adding new files, we need to make sure they are picked up by the documentation runner, which means we need to add them to the appropriate files [here](https://github.com/asyml/forte/tree/master/docs/code).

We are updating our guide to make these more explicitly, but it is still good to refer to our docstring guide [here](https://github.com/asyml/forte/blob/master/CONTRIBUTING.md#docstring)

forte/data/entry_type_generator.py
Xuezhi-Liang: Will add it in the next PR.

forte/data/entry_type_generator.py
Xuezhi-Liang: OK, In new PR will modify this.

forte/data/entry_type_generator.py
Xuezhi-Liang: Will delete in new PR

tests/forte/data/entry_type_generator_test.py
Xuezhi-Liang: Will refactor the test cases in the new PR.

forte/data/entry_type_generator.py
Xuezhi-Liang: This script is created by @qinzzz. Can I delete something like [get_type_attributes()](https://github.com/asyml/forte/blob/366692ba6edfdbcf4b2738cca75ff976b16c456e/forte/data/entry_type_generator.py#L19) function, cause we already changed  the issue's description.

forte/data/entry_type_generator.py
Xuezhi-Liang: Modified in new PR.

forte/data/entry_type_generator.py
Xuezhi-Liang: Now we have two ways to find the attributes, one is __annotations__, the other one is __dataclass_fields__.  like the below image shows.
image

However, when we use __dataclass_fields__, It will also find the attributes of the parent class. as the below image shows.
image

So from my side,  __annotations__ may be the only way to do this.

Xuezhi-Liang:
> btw, when we are adding new files, we need to make sure they are picked up by the documentation runner, which means we need to add them to the appropriate files [here](https://github.com/asyml/forte/tree/master/docs/code).
> Already added entry_type_generator.py
> We are updating our guide to make these more explicitly, but it is still good to refer to our docstring guide [here](https://github.com/asyml/forte/blob/master/CONTRIBUTING.md#docstring)

forte/data/entry_type_generator.py
hunterhector: Need to add docstring to the class to explain what the class does.

forte/data/entry_type_generator.py
hunterhector: yes, please add typing and modify docsting accordingly. When @qinzzz drafted the class she didn't know the final outcome. So now is the time to finalize them.

forte/data/entry_type_generator.py
hunterhector: hmm, interesting. So `dataclass` fields might actually be better (less relying to type annotation). And we might actually want everything from all the ancestor classes. @qinzzz do you need the attributes from all the ancestors.

Xuezhi-Liang: If our Entry name is 'ft.onto.base_ontology.Token' or similar, we can use 'annotation' or '__dataclass_fields__' to get all the attributes.  And the difference is shown above.

However, when I use the entry name as 'entry_type_generator_test.TokenTest' in the test. The class does not have the '__dataclass_fields__' parameter. So in this case, we can non use '__dataclass_fields__' to do this.  As the below images shows.

image
image

So 'annotations' may be the only way to do this.

Xuezhi-Liang:
We use `__dataclass_fields__` to get type attributes based on all the discussions above.  @qinzzz  Could you please view if the code is ok?  Please inform @hunterhector to merge and close this issue if it's ok.

BTW, @hunterhector mentioned that it seems we do not need to design an entry-type generator class if we only have one function here.

forte/data/entry_type_generator.py
```diff
@@ -39,3 +52,39 @@ def get_type_attributes():
 
         """
         return _get_type_attributes()
+
+    @staticmethod
+    def get_entry_attributes_by_class(input_entry_class_name: str):
+        """Get type attributes by class name. input_entry_class_name should be
+        an object class dotted name path.
+        Please note must add "data class" decorator to all the class and their
```
hunterhector, line 60: Let's use the real name `dataclass` (remeber to use backtip), and provide a link to python: https://docs.python.org/3/library/dataclasses.html.  `dataclass` is a python concept and we need to follow the exact name. If you want to mention the decorator, literally just say `@dataclass`

Please also rephrase the sentence to make it grammatically correct, for example:

Note: This function is only applicable to classes decorated as Python `dataclass` since it relies on the `__dataclass_fields__` to find out the attributes.

forte/data/entry_type_generator.py
```diff
@@ -39,3 +52,39 @@ def get_type_attributes():
...
+        ancestors.
+
+
+        Args:
+            input_entry_class_name: An object class dotted name path.
+
+        Returns:
+             A list of attributes corresponding to the input class.
```
hunterhector, line 68: The documentation didn't mention what would happen if the input class is decorated by `dataclass`

hunterhector (dismissed):

hunterhector (approved):

Issue #634, opened 2022-02-09
Texer numpy version requirement is outdated


feipenghe:
**Describe the bug**
Recent developments shows that numpy version requirement <1.19 is outdated, and it causes incompatibility with recent tensorflow versions and potentially other packages in the future.

**To Reproduce**
No particular way to reproduce.

**Expected behavior**
It's expected `texar` numpy version is updated to be compatible with numpy>1.21 

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**
An example issue occured in the git runner is here https://github.com/asyml/forte-wrappers/runs/5120444152?check_suite_focus=true#step:8:83

According to @hunterhector , 
> About the Numpy issue,  it is restricted here https://github.com/asyml/texar-pytorch/blob/master/requirements.txt#L3 
Previously I restricted this to 1.19.5 since there will be CI errors. I remember that’s due to mypy checks since Numpy introduce its own typing since 1.20 (before that we are using generated stubs). So without this 1.19.5 limit, it will raise a lot of linting errors. Now this restriction is a bit too strict and will be outdated soon.

hunterhector:
This PR https://github.com/asyml/texar-pytorch/pull/352 has solved the problem, can we update the versions on TensorFlow now?

feipenghe:
This issue has been fixed.

PR #661, opened 2022-02-28
Upgrade Numpy version

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/634#event-6141238942. 

### Description of changes
Update numpy version and pass the git CI test.

### Test Conducted
General git CI test

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 sortedcontainers==2.1.0
-numpy>=1.16
+numpy>=1.19
 jsonpickle==1.4
 pyyaml==5.4
 smart-open==1.8.4


hunterhector (commented):
do we need to constraint the version on Forte? What will be broken without upgrading this?

feipenghe:
> do we need to constraint the version on Forte? What will be broken without upgrading this?

Do you mean numpy version? I thought this is the change I need to make. I am not sure what else I need to do to fix it.

hunterhector:
> > do we need to constraint the version on Forte? What will be broken without upgrading this?
> 
> Do you mean numpy version? I thought this is the change I need to make. I am not sure what else I need to do to fix it.

I don't understand what you are trying to fix here, you said "pass the git CI test", but which one is failed before and is it fixed now?

Here is what happened:
1. You created an issue here: https://github.com/asyml/forte/issues/634
2. This issue shows a problem in the workflow here: https://github.com/asyml/forte-wrappers/runs/5120444152?check_suite_focus=true#step:8:83
3. Based on what you reported, we allow texar-pytorch to work with numpy >=1.20, https://github.com/asyml/texar-pytorch/pull/352, no <1.19 anymore

So now you should check whether the problems are fixed. Not to create a PR that **adds more restrictions to the numpy version**. 

And btw, you **cannot** fix a comment, if you want to fix something, find an issue or create an issue.

PR #665, opened 2022-03-02
change elements from list to dict

### LINKED ISSUE #663: Make self.elements a dictionary
Currently self.elements is a list object. Every type's sortedlist is stored in index i, where i is defined by type_index_dict. We found that although using a list can save space, such index mapping is brittle when the list of type changes. Thus, we want to convert the list to a dictionary where keys are type names and values are sortedlists. 

**Depends on**

Change the self.elements to a dictionary, where keys are type names and values are sortedlists. Get rid of self.__type_index_dict as we do not need a type-index mappinng.

**Additional context**

qinzzz:
This PR fixes #663. 

### Description of changes
Changed structure self.elements in DataStroe from list to dict. Deleted type_index_dict.

### Test Conducted
Updated the structure in data store test.


forte/data/data_store.py
```diff
@@ -191,7 +168,7 @@ def __init__(self, onto_file_path: Optional[str] = None):
                 ...
             ]
         """
-        self.__elements: List = []
+        self.__elements: dict = {}
```
wanglec, line 171: I think we also need to update the description for elements.

hunterhector: yeah, let's update the description

forte/data/data_store.py
```diff
@@ -454,19 +431,18 @@ def get(
...
-                for entry in self.__elements[id]:
+            for type in self.__elements:
+                if issubclass(get_class(type), entry_class):
+                    all_types.append(type)
+            for type in all_types:
+                for entry in self.__elements[type]:
                     yield entry
         else:
-            for entry in self.__elements[type_id]:
+            for entry in self.__elements[type_name]:
```
hunterhector, line 445: Consider add `KeyError`?

PR #676, opened 2022-03-11
Implement get_entry, prev_entry, and next_entry for DataStore

### LINKED ISSUE #646: implement prev_entry and next_entry
Given an entry represented by `tid`, the user may want to get the previous or next entry with the same type. They can use function prev_entry() and next_entry() to do that.
We provide to two function signatures `next_entry(self, tid: int) -> List` and `prev_entry(self, tid: int) -> List`.

**Depends on**

Given `tid` of an entry, we can call `get_entry()` to find the current entry's index. Then in `self.__elements`, we can retrieve the entry list with the same type. We can use the entry index we get can to find the previous and next record in `entry_list`.

**Additional context**

mylibrar:
This PR fixes #646. 

### Description of changes
Implement get_entry, prev_entry, and next_entry for DataStore.

### Possible influences of this PR
This PR is subject to change after merging #628, #635, and #665.

### Test Conducted
`test_get_entry` and `test_next_entry` in `data_store_test.py`


forte/data/data_store.py
```diff
@@ -493,4 +520,9 @@ def prev_entry(self, tid: int) -> List:
         Returns:
             The previous entry of the same type as the `tid` entry.
```
hunterhector, line 521: Update docstring to include the `IndexError`

forte/data/data_store.py
```diff
@@ -480,7 +501,13 @@ def next_entry(self, tid: int) -> List:
         Returns:
             The next entry of the same type as the `tid` entry.
         """
```
hunterhector, line 503: Update docstring to include the `IndexError`

hunterhector: Also update the docstring to explain the format of the `next entry`

forte/data/data_store.py
```diff
@@ -433,10 +434,30 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+        if entry_type not in self.__type_index_dict:
+            raise KeyError(f"Entry of type {entry_type} is not found.")
+        type_id = self.__type_index_dict[entry_type]
+        index_id = -1
+
         # If the entry is an annotation, bisect the annotation sortedlist
         # to find the entry. May use LRU cache to optimize speed.
         # Otherwise, use `index_id` to find the index of the entry.
-        raise NotImplementedError
+        if issubclass(get_class(entry_type), (Annotation, AudioAnnotation)):
```
hunterhector, line 449: `issubclass` check and `get_class` might slow down the process. @wanglec @qinzzz do we have the class hierarchy somewhere?

Another idea is that we should cache the subclass information (e.g. using an `is_annotation_dict`) and only check it the first time.

forte/data/data_store.py
```diff
@@ -480,7 +501,13 @@ def next_entry(self, tid: int) -> List:
...
-        raise NotImplementedError
+        _, type_id, index_id = self.get_entry(tid=tid)
+        entry_list = self.__elements[type_id]
+        if index_id >= len(entry_list) - 1:
+            raise IndexError(
+                f"Entry with tid {tid} is the last entry in entry list."
+            )
+        return entry_list[index_id + 1]
 
     def prev_entry(self, tid: int) -> List:
```
hunterhector, line 512: It should be important to document the list ordering of the non-annotation types. What would happen if we query `prev_entry` or `next_entry` on those? @wanglec

wanglec: if it is a non-annotation type, it will be sorted in the insertion order, which means the next and prev would return the next/previous inserted entry.

wanglec: We also change the structure of `self.__elements` in data_pack.py. Sorry for the confusion in the string description.
it looks like this:
{"ft.onto.base_ontology.Document": SortedList([]), "ft.onto.base_ontology.Sentence":  SortedList([]) }, etc.

forte/data/data_store.py
mylibrar: #628 provides a [_is_annotation()](https://github.com/asyml/forte/pull/628/files#diff-f1c68405498aa1cba27f81862f28d3edecfdc2d8501394bb3cac9d83dfa1283dR250) method and this line will be updated when #628 is merged.

forte/data/data_store.py
mylibrar: Thanks for the update! I've fixed the implementation based on #665.

forte/data/data_store.py
```diff
@@ -399,21 +400,40 @@ def _delete_entry_by_loc(self, type_id: int, index_id: int):
...
         # If the entry is an annotation, bisect the annotation sortedlist
         # to find the entry. May use LRU cache to optimize speed.
         # Otherwise, use `index_id` to find the index of the entry.
-        raise NotImplementedError
+        index_id = -1
+        if issubclass(get_class(entry_type), (Annotation, AudioAnnotation)):
+            entry_list = self.__elements[entry_type]
+            begin: int = entry_list.bisect_left(entry)
+            for i, e in enumerate(entry_list[begin:]):
+                if e[2] == entry[2]:
```
mylibrar, line 429: We can make use of the [constant](https://github.com/asyml/forte/pull/635/files#diff-bed4b362357669e7f878c5c84fde9f9b6ed8f1a1d88b81a3e143a547ac8fa9c1R3) to replaced the hardcoded index here after #635 is merged.

forte/data/data_store.py
```diff
@@ -477,21 +477,40 @@ def _delete_entry_by_loc(self, type_name: str, index_id: int):
...
+        entry_type = entry[constants.ENTRY_TYPE_INDEX]
+        if entry_type not in self.__elements:
+            raise KeyError(f"Entry of type {entry_type} is not found.")
+
         # If the entry is an annotation, bisect the annotation sortedlist
         # to find the entry. May use LRU cache to optimize speed.
         # Otherwise, use ``index_id`` to find the index of the entry.
-        raise NotImplementedError
+        index_id = -1
+        if self._is_annotation(entry_type):
```
hunterhector, line 502: If we don't need `index_id`, does it mean that we don't need to bisect the list? The `entry` content is already obtained through `self.__entry_dict`.

It seems that we only need `index_id` when we are doing operations like `prev`, `next`, or some sort of iteration. If we simply want to take the entry value, `tid` lookup seems to be sufficient.

Also, @wanglec @qinzzz any comments?

forte/data/data_store.py
```diff
@@ -533,9 +552,20 @@ def next_entry(self, tid: int) -> List:
...
-            The next entry of the same type as the ``tid`` entry.
+            A list of attributes representing the next entry of the same type
+            as the ``tid`` entry.
+
+        Raises:
+            IndexError: An error occurred accessing the next entry of the last
+                element in entry list.
         """
-        raise NotImplementedError
+        _, entry_type, index_id = self.get_entry(tid=tid)
```
hunterhector, line 562: The first return value `_` actually contains entry information right? Why do we need to discard it and use the `index_id` again. @wanglec @qinzzz

forte/data/data_store.py
hunterhector: > if it is a non-annotation type, it will be sorted in the insertion order, which means the next and prev would return the next/previous inserted entry.

Let's clarify these behaviors in docstring

forte/data/data_store.py
qinzzz: Yes, if we only need the `entry list` and `entry name`, we can directly get them from `entry_dict`. We can probably add a separate function for getting `index_id`?

qinzzz: Since `delete_entry` function also needs to get `index_id`, we can implement this in a new pull request.

forte/data/data_store.py
```diff
@@ -477,21 +477,40 @@ def _delete_entry_by_loc(self, type_name: str, index_id: int):
...
+
         # If the entry is an annotation, bisect the annotation sortedlist
         # to find the entry. May use LRU cache to optimize speed.
         # Otherwise, use ``index_id`` to find the index of the entry.
-        raise NotImplementedError
+        index_id = -1
+        if self._is_annotation(entry_type):
+            entry_list = self.__elements[entry_type]
+            begin: int = entry_list.bisect_left(entry)
+            for i, e in enumerate(entry_list[begin:]):
```
qinzzz, line 505: Since every entry is unique, I believe we only need to check entry_list[begin]. If it does not have the same `tid`, then it means the entry does not exist. Correct me if I'm wrong.

forte/data/data_store.py
```diff
@@ -533,9 +552,20 @@ def next_entry(self, tid: int) -> List:
...
+
+        Raises:
+            IndexError: An error occurred accessing the next entry of the last
+                element in entry list.
         """
-        raise NotImplementedError
+        _, entry_type, index_id = self.get_entry(tid=tid)
+        entry_list = self.__elements[entry_type]
+        if index_id >= len(entry_list) - 1:
+            raise IndexError(
```
qinzzz, line 565: if `index_id == len(entry_list) - 1`, we can return None instead of raising an exception since its still an expected input.

forte/data/data_store.py
hunterhector: we can actually add a function in this PR since (prev, next) will need the id.

forte/data/data_store.py
qinzzz: `index_id` is not in the entry's list, and next_entry only need entry's index and type informaton.

forte/data/data_store.py
```diff
@@ -546,9 +576,19 @@ def prev_entry(self, tid: int) -> List:
...
+        Raises:
+            IndexError: An error occurred accessing the previous entry of the
+                first element in entry list.
         """
-        raise NotImplementedError
+        _, entry_type, index_id = self.get_entry(tid=tid)
+        if index_id <= 0:
+            raise IndexError(
+                f"Entry with tid {tid} is the first entry in entry list."
+            )
```
qinzzz, line 590: Same here, it;s better for index_id=0 to return None, not expection

forte/data/data_store.py
hunterhector: we should call another `get_entry_index` function (mentioned above) instead of this.

forte/data/data_store.py
hunterhector: I actually have a question about the index here. Why do we need to access the `entry_list` via `index_id + 1`?

hunterhector: And this may help me understand why an out-of-bound index is expected.

wanglec: So we first find the `index_id` of the current entry with `tid`. Then we find the next entry, which has index `index_id+1`.

hunterhector: Oh, forgot we are working on `next_entry` here, but why is `len(entry_list) - 1` valid but return None?

qinzzz: I think since we may use next_entry to iterate through a list, we don't want to get an exception when reaching the end of the list. We only need `None` to indicate that this is the last one.

forte/data/data_store.py
wanglec: So `get_entry_index` will only return the `index_id`, right?

hunterhector: yes, so `get_entry` from `tid` directly get the value, while `get_enty_index` can provide an index for `delete`, `next` purposes.

forte/data/data_store.py
```diff
@@ -524,31 +558,48 @@ def get(
...
+            of the last element in entry list.
         """
-        raise NotImplementedError
-
-    def prev_entry(self, tid: int) -> List:
+        _, entry_type = self.get_entry(tid=tid)
+        index_id: int = self.get_entry_index(tid=tid)
+        entry_list = self.__elements[entry_type]
+        if index_id >= len(entry_list) - 1:
+            return None
```
hunterhector, line 580: I think the idea is to only return `None` at the edges. For indices that's out of the list, should still be `IndexError`

forte/data/data_store.py
hunterhector: Here are some related thoughts:

1. In Python the correct way to deal with iterator/loop end is to raise an iteration for `StopIteration`, what you proposed can be achieved by providing iterators of a certain entry type.
2. Consider we expand the behavior of `next_entry` with a parameter `step = 1` (which mimics the step parameter in list iteration), when should we use None?

forte/data/data_store.py
```diff
@@ -477,21 +477,55 @@ def _delete_entry_by_loc(self, type_name: str, index_id: int):
...
+        and its ``type_name``.
 
         Args:
             tid (int): Unique id of the entry.
 
         Returns:
-            The entry which ``tid`` corresponds to, its ``type_id`` and its index
-            in the ``type_name`` list.
+            The entry which ``tid`` corresponds to and its ``type_name``.
+
```
hunterhector, line 489: `Raises`

tests/forte/data/data_store_test.py
```diff
@@ -386,33 +392,36 @@ def test_is_annotation(self):
...
+                5,
+                1234,
+                "ft.onto.base_ontology.Document",
+                None,
+                "Postive",
+                None,
+            ],
+        )
+        # First entry in list does not have a previous entry.
+        self.assertIsNone(self.data_store.prev_entry(1234))
```
hunterhector, line 424: asserRaises

hunterhector (approved):

PR #689, opened 2022-03-17
Given a string, return if it's a subclass of a type #687

### LINKED ISSUE #687: Given a string, return if it's a subclass of a type
In DataStore, class types are represented as strings. In certain cases, we need to know if a string `class_type` is the subclass of a class, e.g. Annotation, Link or Group. Therefore, we want to implement a new internal function `_is_subclass(type_name, cls)`.

**Depends on**
None

We want an efficient solution to this problem. If one `type_name` is queried for multiple times, we can cache the result of its superclass and return immediately. Moreover, for every type, we can store the parent class information in type_attributes. 
Note that we accpect two types of class: the class defined in forte, or the classes in user provided ontology file. A flag "include_onto_file" can be configured in the function parameter.

**Additional context**

Xuezhi-Liang:
This PR fixes #687. 

### Description of changes

### Test Conducted


Xuezhi-Liang:
My solution is that we can first get the class by the string name and then find their base(father) class. Last, determine if their base class is the same class users want to check. Please check the first version PR #689.

hunterhector:
Looking at the original issue, there is the following solution proposed:

```
We want an efficient solution to this problem. If one type_name is queried for multiple times, we can cache the result of its superclass and return immediately. Moreover, for every type, we can store the parent class information in type_attributes.
Note that we accpect two types of class: the class defined in forte, or the classes in user provided ontology file. A flag "include_onto_file" can be configured in the function parameter.
```

This is not done in this PR, please reconsider the requirement

qinzzz:
Maybe we can keep a filed "parent class" in the _type_attribute dictionary as a cache. When a class_name is queried, we can check if the class's parent already exists in the dictionary; if so, we can check if the parent matches the given class in an iterative way (Or we can store all parent classes as a list to save more time). If not, we use `issubclass()` to check and store the result. 
Note that if user provides an onto_file_path, we should have stored the parent classic the _type_attribute dictionary, so that we will not (and actually cannot) use `issubclass` to check it.

forte/data/data_store.py
```diff
@@ -247,6 +247,23 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+            type_name: A fully qualified name of an entry class
+            cls: A class.
+
+        Returns:
+            A boolean value whether this type belongs to a clc class
+            type or not.
+
+        """
+        entry_class = get_class(type_name)
+        return cls == entry_class.__base__
```
qinzzz, line 265: Should also consider the case when `cls` is the grandparent of `entry_class`.

Xuezhi-Liang: OK， Got it.

forte/data/data_store.py
```diff
@@ -247,6 +247,23 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+    def _is_subclass(type_name: str, cls) -> bool:
+        r"""This function determines if a string class_type is
+        the subclass of a class.
+
+        Args:
+            type_name: A fully qualified name of an entry class
+            cls: A class.
+
+        Returns:
+            A boolean value whether this type belongs to a clc class
```
feipenghe, line 260: Check sphinx rendering effect it might shows abnormally as it's expecting a variable name here.

hunterhector:
need to resolve the conflicts for this PR to proceed.

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
         entry += len(self._type_attributes[type_name]) * [None]
         return entry
 
+    def _is_subclass(
+        self, type_name: str, cls, include_onto_file: bool = False
+    ) -> bool:
+        r"""This function determines if a string class_type is
```
hunterhector, line 253:
```suggestion
        r"""This function determines if a string ``class_type`` is
```

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        the subclass of a class.
+
+        Args:
+            type_name: A fully qualified name of an entry class
+            cls: A class.
```
hunterhector, line 258: `A class` is a very under specified documentation, please elaborate.

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+            include_onto_file: A boolean value whether the class is in user provided ontology file
+
+        Returns:
+            A boolean value whether this `type_name` class is the subclass of `cls` or not.
+
+        """
+
+        if include_onto_file:
+            if type_name in self._type_attributes["parent_class"].keys():
```
hunterhector, line 267: The key for `_type_attributes` should be entry type class names, why `parent_class` string is used here.

hunterhector: The structure for `_type_attributes` should be `self._type_attributes[type_name]["parent_class"]`

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+                if (
+                    cls.__module__ + "." + cls.__name__
```
hunterhector, line 269: We should use utilities instead of writing this again: https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L114

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+    def _is_subclass(
+        self, type_name: str, cls, include_onto_file: bool = False
+    ) -> bool:
+        r"""This function determines if a string class_type is
+        the subclass of a class.
+
+        Args:
+            type_name: A fully qualified name of an entry class
+            cls: A class.
+            include_onto_file: A boolean value whether the class is in user provided ontology file
```
hunterhector, line 259: I don't think we need to use `include_onto_file` value here. This function should check `_type_attributes` and python environment. I don't understand what this boolean is doing here.

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+            if type_name in self._type_attributes["parent_class"].keys():
+                if (
+                    cls.__module__ + "." + cls.__name__
+                    in self._type_attributes["parent_class"][type_name]
+                ):
+                    return True
+                return False
+            return False
+        else:
+            if "parent_class" not in self._type_attributes.keys():
```
hunterhector, line 276: If you want to check key existence, don't use `keys`, directly do `a in dict`, not `a in dict.keys()`

forte/data/data_store.py
```diff
@@ -247,6 +247,59 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+                    cls.__module__ + "." + cls.__name__
+                    in self._type_attributes["parent_class"][type_name]
+                ):
+                    return True
+            else:
+                entry_class = get_class(type_name)
+                self._type_attributes["parent_class"][type_name] = []
+                if issubclass(entry_class, cls):
+                    self._type_attributes["parent_class"][type_name].append(
+                        cls.__module__ + "." + cls.__name__
```
hunterhector, line 290: There are many calls of `cls.__module__ + "." + cls.__name__`, should be done once in one place.

forte/data/entry_type_generator.py
```diff
@@ -77,7 +77,7 @@ def get_entry_attributes_by_class(input_entry_class_name: str) -> list:
         ["speaker", "part_id", "sentiment", "classification", "classifications"].
         The solution looks like the following:
 
-        ... code-block::python
+        .. code-block::python
```
hunterhector, line 80: This file can be deleted after merge with master.

tests/forte/data/data_store_test.py
```diff
@@ -18,7 +18,7 @@
 import logging
 import unittest
 from sortedcontainers import SortedList
-
+from forte.data.ontology.top import Annotation, Group, Link
```
hunterhector, line 21: Let's import these locally in `test_is_subclass` function.

forte/data/data_store.py
```diff
@@ -247,6 +248,50 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        Returns:
+            A boolean value whether ``type_name``  class is the``cls``
+            subclass or not.
+
+        """
+        if type_name not in self._type_attributes:
+            self._type_attributes[type_name] = {}
+        if "parent_class" not in self._type_attributes[type_name]:
+            self._type_attributes[type_name].update(parent_class=[])
+        cls_qualified_name = cls.__module__ + "." + get_qual_name(cls)
```
hunterhector, line 269: https://github.com/asyml/forte/blob/master/forte/utils/utils.py#L37

forte/data/data_store.py
```diff
@@ -250,6 +251,44 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+
+        Returns:
+            A boolean value whether ``type_name``  class is the``cls``
+            subclass or not.
+
+        """
+        if type_name not in self._type_attributes:
+            self._type_attributes[type_name] = {}
+        if "parent_class" not in self._type_attributes[type_name]:
+            self._type_attributes[type_name].update(parent_class=[])
```
hunterhector, line 271: why using `update` instead of simply `["parent_class"]=[]`

forte/data/data_store.py
```diff
@@ -250,6 +251,44 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        cls_qualified_name = get_full_module_name(cls)
+        type_name_parent_class = self._type_attributes[type_name][
+            "parent_class"
+        ]
+        if len(type_name_parent_class) > 0:
+            return bool(cls_qualified_name in type_name_parent_class)
```
hunterhector, line 277: If we need to check containment from `type_name_parent_class`, probably making it a `set`?

forte/data/data_store.py
```diff
@@ -250,6 +251,44 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        else:
+            entry_class = get_class(type_name)
+            if issubclass(entry_class, cls):
+                type_name_parent_class.append(cls_qualified_name)
+                cls_base_class = cls.__base__
```
hunterhector, line 282: I find that `__base__` is undocumented and its behavior is non-guaranteed. Probably should not use it.

forte/data/data_store.py
```diff
@@ -250,6 +251,44 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+                if cls_base_class is not None:
+                    cls_base_qualified_name = get_full_module_name(
+                        cls_base_class
+                    )
+                    type_name_parent_class.append(cls_base_qualified_name)
```
hunterhector, line 287: I don't understand why we are adding the `base` of `cls` to the list.

tests/forte/data/data_store_test.py
hunterhector: this comment is not addressed

forte/data/data_store.py
Xuezhi-Liang: OK, will change in the next PR.

forte/data/data_store.py
Xuezhi-Liang: It‘s my fault. Fixed in the new PR.

forte/data/data_store.py
Xuezhi-Liang: ok, changed in the new PR

tests/forte/data/data_store_test.py
```diff
@@ -21,7 +21,7 @@
 from typing import Optional, Dict
 from dataclasses import dataclass
 from forte.data.data_store import DataStore
-from forte.data.ontology.top import Annotation, Generics
+from forte.data.ontology.top import Annotation, Generics, Group, Link
```
Xuezhi-Liang, line 24: I saw some test cases already imported `Annotation` here, So do we need to import these locally in the `test_is_subclass` function?

forte/data/data_store.py
```diff
@@ -250,6 +251,46 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+            else:
+                cls_subclasses = cls.__subclasses__()
+                for cls_subclass in cls_subclasses:
+                    if issubclass(entry_class, cls_subclass):
+                        cls_subclass_qualified_name = get_full_module_name(
+                            cls_subclass
+                        )
+                        type_name_parent_class.add(cls_qualified_name)
+                        type_name_parent_class.add(cls_subclass_qualified_name)
+                return bool(cls_qualified_name in type_name_parent_class)
```
hunterhector, line 292: This `else` block is a bit confusing to me. What are we trying to achieve here? `issubclass` check can check for any direct/indirect subclassing, listing the `subclass` of `cls` won't add anything new.

tests/forte/data/data_store_test.py
hunterhector: Let's try to move all of them to local since importing here will change some test behaviors.

forte/data/data_store.py
```diff
@@ -250,6 +251,46 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+        if type_name not in self._type_attributes:
+            self._type_attributes[type_name] = {}
+        if "parent_class" not in self._type_attributes[type_name]:
+            self._type_attributes[type_name]["parent_class"] = set()
+        cls_qualified_name = get_full_module_name(cls)
+        type_name_parent_class = self._type_attributes[type_name][
+            "parent_class"
+        ]
+        if len(type_name_parent_class) > 0:
+            return bool(cls_qualified_name in type_name_parent_class)
```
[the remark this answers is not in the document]
qinzzz: If `len(type_name_parent_class) > 0`, I believe it does not necessarily mean that the set contains all possible parents, since you only add one class in line 281 and returns.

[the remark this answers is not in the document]
hunterhector: I guess a simple approach is to check:

```
if cls_qualified_name in type_name_parent_class:
  return True
else:
  # check is_subclass
```

Basically, we are caching the result for this pair, and if not, we can use `is_subclass`. @qinzzz  what do you think

Xuezhi-Liang:
Based on the recent comments, @qinzzz requirements and suggested solution.  This function accepts two types of class: the class defined in forte, or the classes in the user-provided ontology file.  Note that if the user provides an onto_file_path, we should have stored the parent classic in the `_type_attribute` dictionary, so that we will not (and actually cannot) use `issubclass` to check it. So it is reasonable that add the "include_onto_file" flag in the function parameter and we may still need it.  I modified the code, please help check it again.

forte/data/data_store.py
```diff
@@ -253,6 +254,47 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+    ) -> bool:
+        r"""This function takes a fully qualified ``type_name`` class name,
+        ``cls`` class and returns whether ``type_name``  class is the``cls``
+        subclass or not. This function accpect two types of class: the class defined
+        in forte, or the classes in user provided ontology file.
+
+        Args:
+            type_name: A fully qualified name of an entry class.
+            cls: An entry class.
+            include_onto_file: A boolean value whether ``type_name`` class
```
hunterhector, line 268: Let's simply rename this variable to avoid confusion.

```suggestion
            no_dynamic_subclass: A boolean value controlling where to look for subclasses. If True, this function will not check the subclass relations via `issubclass` but rely on pre-populated states only.
```

forte/data/data_store.py
```diff
@@ -253,6 +254,47 @@ def _new_group(self, type_name: str, member_type: str) -> List:
...
+            self._type_attributes[type_name] = {}
+        if "parent_class" not in self._type_attributes[type_name]:
+            self._type_attributes[type_name]["parent_class"] = set()
+        cls_qualified_name = get_full_module_name(cls)
+        type_name_parent_class = self._type_attributes[type_name][
+            "parent_class"
+        ]
+
+        if include_onto_file:
+            return bool(cls_qualified_name in type_name_parent_class)
```
hunterhector, line 286: remove the `bool` wrap

tests/forte/data/data_store_test.py
```diff
@@ -647,6 +647,26 @@ def test_get_entry_attribute_by_class(self):
...
+            )
+        )
+        # ``Generics`` is the grandfather of "data_store_test.SingleMetricTest" class.
+        self.assertTrue(
+            self.data_store._is_subclass(
+                "data_store_test.SingleMetricTest", Generics
+            )
+        )
+
+
```
qinzzz, line 669: Add test in:
1. query class that already exists in type_attributes (you can add more info into the type_attributes for testing)
2. get every ancestors of a class
3. test with the flag `no_dynamic_subclass` on

hunterhector: Looks like 3 tests suggested are included. Resolving this.

Issue #688, opened 2022-03-16
Improve the procedure to convert Forte pipeline to YAML


zhanyuanucb:
**Is your feature request related to a problem? Please describe.**
Collaborating with some submission tools to submit a Forte pipeline often requires saving the pipeline config. It will be convenient to have a method to tell Forte pipeline from which dir the submission tool will pick up the serialized pipeline.

**Describe the solution you'd like**
Add an `export` method to `Pipeline`. This method takes in a `commit_name` and checks if a hidden environment variable `FORTE_EXPORT_PATH` is assigned. If that is the case, then this method will name the pipeline config `<commit_name>.yml` and dump it to that dir.

**Describe alternatives you've considered**

**Additional context**

PR #690, opened 2022-03-17
Forte export

zhanyuanucb:
This PR fixes https://github.com/asyml/forte/issues/688 

### Description of changes
Added `export` method to `Pipeline` that checks a predefined environment variable `FORTE_EXPORT_PATH` and export pipeline config into it.

### Test Conducted
Added tests in tests/forte/pipeline_test.py
- `export` should do nothing if `FORTE_EXPORT_PATH` is not defined
- `export` should raise `FileNotFoundError` if `FORTE_EXPORT_PATH` is assigned but does not exists
- `export` should dump pipeline config to `FORTE_EXPORT_PATH` if it exists
-  users should be able to export a pipeline multiple times, with different export names


forte/pipeline.py
```diff
@@ -466,6 +467,19 @@ def save(self, path: str):
...
+        r"""Export pipeline to FORTE_EXPORT_PATH
+        FORTE_EXPORT_PATH is an environment variable
+        The pipeline is saved by :meth:`~forte.pipeline.Pipeline.save`
+        Args:
+            name: Export name of the pipeline
+        """
+        export_path = os.environ.get("FORTE_EXPORT_PATH")
+        if export_path:
+            if not os.path.exists(export_path):
+                raise FileNotFoundError()
```
hunterhector, line 480: Please add clear message in the exception

forte/pipeline.py
```diff
@@ -466,6 +467,19 @@ def save(self, path: str):
...
+    def export(self, name: str):
+        r"""Export pipeline to FORTE_EXPORT_PATH
+        FORTE_EXPORT_PATH is an environment variable
+        The pipeline is saved by :meth:`~forte.pipeline.Pipeline.save`
+        Args:
+            name: Export name of the pipeline
+        """
+        export_path = os.environ.get("FORTE_EXPORT_PATH")
+        if export_path:
+            if not os.path.exists(export_path):
```
hunterhector, line 479: Actually, we might want to create the export path if not found.

forte/pipeline.py
```diff
@@ -466,6 +467,19 @@ def save(self, path: str):
         with open(path, "w", encoding="utf-8") as f:
             yaml.safe_dump(self._dump_to_config(), f)
 
+    def export(self, name: str):
```
hunterhector, line 470: I remember we discussed a method to generate a name when none is given (and use auto increment numbers for new files)

hunterhector: Anything can be a user's job here, the discussion and spirit behind all this would be to make sure care as little as possible.

Now, we allow someone to use a name for sure, but it can be even simpler if we just use a default name, and the user don't need to worry about it if he is not planning to use multiple pipeline submission.

Uniqueness check can be another thing, one approach is we always overwrite the previous one if the name is not unique, that's reasonable too.

forte/pipeline.py
```diff
@@ -466,6 +470,28 @@ def save(self, path: str):
...
+
+        Returns:
+            Optional[str]: Export path of pipeline config YAML.
+        """
+        export_path: Optional[str] = os.environ.get("FORTE_EXPORT_PATH")
+        if export_path:
+            if not os.path.exists(export_path):
+                os.makedirs(export_path, exist_ok=True)
+            export_path = os.path.join(
+                export_path, f"{name}-{self._export_count}.yml"
```
aurickq, line 489: Better to just require the `name` argument (not default it to `pipeline`) and raise a `ValueError` if there is a conflict. adding a suffix will be unintuitive to users who write the lines
```
pipeline.export("my-pipeline")
pipeline.export("my-new-pipeline")
```
but pipelines with the names `my-pipeline-1` and `my-new-pipeline-2` show up on the other end

aurickq: D'oh just saw @hunterhector's other comment. Ignore this one.

zhanyuanucb: But I think your comment also makes sense.
How about this:
- If user didn't specify an export name, then the default one will be `pipeline-<self._export_count>.yml`
- The `self._export_count` only increments if the pipeline is exported with the default name
- If user provides an export name, then we'll use it. And raise `ValueError` if conflicts.

forte/pipeline.py
```diff
@@ -466,6 +470,28 @@ def save(self, path: str):
         with open(path, "w", encoding="utf-8") as f:
             yaml.safe_dump(self._dump_to_config(), f)
 
+    def export(self, name: str = "pipeline") -> Optional[str]:
+        r"""Export pipeline to FORTE_EXPORT_PATH
+        FORTE_EXPORT_PATH is an environment variable
+        The pipeline is saved by :meth:`~forte.pipeline.Pipeline.save`
```
hunterhector, line 476: Should make the docstring reflect the following behaviors:
1. The directory will be created if not found
2. A default name will be used to store the output
3. Raise exception when provided name is taken

forte/pipeline.py
```diff
@@ -466,6 +471,43 @@ def save(self, path: str):
...
+        """
+        export_path: Optional[str] = os.environ.get("FORTE_EXPORT_PATH")
+        if export_path:
+            os.makedirs(export_path, exist_ok=True)
+
+            if name:
+                export_name = f"{name}.yml"
+            else:
+                export_name = f"pipeline-{self._export_count}.yml"
+                self._export_count += 1
```
hunterhector, line 502: Actually, this implementation will likely fail. We should check the duplicate file in the directory and then increment the suffix if necessary. Storing a state will cause several problems, the main one is this value will be lost when the pipeline is rerun.

forte/pipeline.py
```diff
@@ -466,6 +471,43 @@ def save(self, path: str):
...
+    def export(self, name: Optional[str] = None) -> Optional[str]:
+        r"""Export pipeline to FORTE_EXPORT_PATH
+        FORTE_EXPORT_PATH is an environment variable
+        The pipeline is saved by :meth:`~forte.pipeline.Pipeline.save`
+        This method will have the following behaviors:
+        - FORTE_EXPORT_PATH will be created if assigned but not found.
+        - If name is not provided, a default one will be used.
+        - When the pipeline is exported multiple time with default name,
+        the export name will be `pipeline-1.yml`, `pipeline-2.yml`...
+        - Raise exception if provided name is taken.
```
hunterhector, line 483: Docstring suggestions

1. The indent of the list doesn't work properly on the generated doc, see the effect here: https://asyml-forte--690.org.readthedocs.build/en/690/code/pipeline.html#forte.pipeline.Pipeline.export
2. Should also make it explicit that: "If the `name` argument is provided, no suffix will be appended"
3. Should make it explicit what exception will be raised (ValueError in this case)
4. In addition to saying the pipeline saved by `save`, also explain the behavior explicitly: the pipeline will be serialized to a file (and then explain the details of the file
5. Explicitly stated that `FORTE_EXPORT_PATH` should be a directory and all serialized files will be stored here.

forte/pipeline.py
zhanyuanucb: So instead using count as suffix, how about using timestamp like `2022.03.18.07.39.27`?

forte/pipeline.py
zhanyuanucb: sry... should have read the CONTRIBUTING.md closely

hunterhector (approved):
Looks very nice, thank you!

PR #692, opened 2022-03-23
Parse user specified ontology file

qinzzz:
This PR handles the parameter onto_file_path in DataStore's init method.

### Description of changes
If onto_file_path is not None, build tyoe_attribute dictionary based on the provided ontology file.

### Test Conducted
test_check_onto_file in data store test.


forte/data/data_store.py
```diff
@@ -474,3 +476,85 @@ def prev_entry(self, tid: int) -> List:
             The previous entry of the same type as the ``tid`` entry.
         """
         raise NotImplementedError
+
+    def parse_onto_file(self):
```
hunterhector, line 480: Try to reuse the ontology parsing code from [OntologyGenerator](https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_generator.py#L191)

forte/data/data_store.py
```diff
@@ -474,3 +476,85 @@ def prev_entry(self, tid: int) -> List:
...
+                raise KeyError(
+                    str(e)
+                    + "\n"
+                    + "Failed to parse onto file. Onto file should"
+                    + "be a json file and contains all required fields."
+                ) from e
+
+    def _get_entry_attributes_by_class(
+        self, input_entry_class_name: str
+    ) -> list:
```
hunterhector, line 519: Return type annotation should be `typing.List` not `list`

forte/data/data_store.py
```diff
@@ -474,3 +476,85 @@ def prev_entry(self, tid: int) -> List:
...
+
+            # return
+            # ["speaker", "part_id", "sentiment", "classification", "classifications"]
+
+        """
+        class_ = get_class(input_entry_class_name)
+        try:
+            return list(class_.__dataclass_fields__.keys())
+        except AttributeError:
+            return []
```
hunterhector, line 560: This function looks like belonging to another PR, why it is here?

hunterhector (commented):
try to reuse the ontology code generator

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
             The previous entry of the same type as the ``tid`` entry.
         """
         raise NotImplementedError
+
+    def parse_onto_file(self):
+        r"""If user provides a customized ontology json file, we parse
+        this file and set self._type_attributes accordingly. It not only
+        includes the classes in the provided ontology file, but also import
+        classes from parent entry files.
```
hunterhector, line 484: First, Consider improving the documentation:

1. Some small grammar issues (e.g. a user)
2. Quote with backtip when referring to variables.
3. Try to explain in a way that people without prior knowledge can understand:
    a. when referring to an internal variable like `self._type_attributes`, explain what it is.
    b. explain the functionality clearly and exact. For example, "provides a customized ontology json" is not very accurate, what do you mean by "custom"? We should say "This function will populate the types and attributes used in the `data_store` with an ontology specification file". Try to avoid ambiguous words like "ontology json file". The same applies to other sentences too. 

Second, in the current design, can the function be private?

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+        """
+        if self.onto_file_path is None:
+            return
+
+        ontoGen = OntologyCodeGenerator()
```
hunterhector, line 489: camel cases shouldn't be used for variables.

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+        ontoGen.parse_ontology_spec(
+            self.onto_file_path, merged_scheme, merged_dict
+        )
+
+        self._type_attributes = {}
+
+        for onto in merged_scheme:
+            entry_name = onto["entry_name"]
+            attr_dic = {}
+            idx = 4
```
hunterhector, line 500: Use a const?

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+        merged_scheme, merged_dict = [], []
+        ontoGen.parse_ontology_spec(
+            self.onto_file_path, merged_scheme, merged_dict
+        )
+
+        self._type_attributes = {}
+
+        for onto in merged_scheme:
+            entry_name = onto["entry_name"]
+            attr_dic = {}
```
hunterhector, line 499:
```suggestion
            attr_dict = {}
```

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+            self._type_attributes[entry_name] = entry_dic
+
+    def _get_entry_attributes_by_class(
+        self, input_entry_class_name: str
+    ) -> List:
+        """Get type attributes by class name. `input_entry_class_name` should be
+        a fully qualified name of an entry class.
+
+        The `dataclass` module<https://docs.python.org/3/library/dataclasses.html> can add
+        generated special methods to user-defined classes. There is an in-built function
```
hunterhector, line 518:
```suggestion
        generated special methods for user-defined classes. There is a built-in function
```

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+        called `__dataclass_fields__` that is called on the class object, and it returns
```
hunterhector, line 519:
```suggestion
        called `__dataclass_fields__` that is called on the class object, and it returns
```

This sentence reads weird to me, maybe rephrase?

forte/data/data_store.py
```diff
@@ -474,3 +476,79 @@ def prev_entry(self, tid: int) -> List:
...
+            `dataclass` since it relies on the `__dataclass_fields__` to find out the attributes.
+
+
+        Args:
+            input_entry_class_name: A fully qualified name of an entry class.
+
+        Returns:
+            A list of attributes corresponding to the input class.
+
+        For example, for Sentence we want to get a list of
```
hunterhector, line 534: This `Sentence` comes from nowhere, we should explain this is an entry, and use the full qualified name.

hunterhector (commented):
In additional to the comments, there are a few more general steps missing:
1. We always need to associate a PR with an issue for tracking purposes.
2. Need to resolve conflicts.

forte/data/data_store.py
hunterhector: Something like "customized ontology specification" is still there, the meaning of "custom"  is not clear, I don't think we need to mention anything about "custom", we just need to mention that if a spec is passed we will use it to populate some internal states.

Now it is also unclear where it is passed?

forte/data/data_store.py
hunterhector: Still not fixed: https://github.com/asyml/forte/blob/dfd9c2ad1e31d9b4670967a3b67964f35b574372/forte/data/data_store.py#L990

forte/data/data_store.py
hunterhector: not fixed: https://github.com/asyml/forte/blob/dfd9c2ad1e31d9b4670967a3b67964f35b574372/forte/data/data_store.py#L991

forte/data/data_store.py
hunterhector: not fixed. What is a `Sentence`?

tests/forte/data/data_store_test.py
```diff
@@ -777,6 +788,63 @@ def test_is_subclass(self):
             )
         )
 
+    def test_check_onto_file(self):
+        data_store_init = DataStore()
+        print("data_store_init", data_store_init)
```
hunterhector, line 793: let's not include the `print` statements in the final PR

forte/data/data_store.py
```diff
@@ -937,7 +940,46 @@ def _parse_onto_file(self):
...
+        )
+
+        for onto in merged_scheme:
+            entry_name = onto["entry_name"]
+            if entry_name in DataStore._type_attributes:
+                continue
+
+            attr_dict = {}
+            idx = constants.ATTR_BEGIN_INDEX
+            if "attributes" in onto:
```
hunterhector, line 956: All these ontology keywords have consts, don't hard-code: https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_const.py#L21

forte/data/data_store.py
```diff
@@ -937,7 +940,46 @@ def _parse_onto_file(self):
...
+                for d in onto["attributes"]:
+                    name = d["name"]
+                    attr_dict[name] = idx
+                    idx += 1
+            entry_dict = {}
+            entry_dict["parent_class"] = set()
+            entry_dict["parent_class"].add(onto["parent_entry"])
+            entry_dict["attributes"] = attr_dict
+            DataStore._type_attributes[entry_name] = entry_dict
```
hunterhector, line 965: Are you trying to parse the parent information yourself?

1. You should be using existing functions
    - https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_generator.py#L680
    - Here is an example usage: https://github.com/asyml/forte/blob/6405c96f179faef3dd58d9d5232962675fceb652/forte/pipeline.py#L672
2. Do your code here properly traverse the tree to make sure we have all the ancestor information available in the ontology file? 

Also notice the EntryTree object actually provide all the attributes: https://github.com/asyml/forte/blob/master/forte/data/ontology/code_generation_objects.py#L798

Also, note that the attributes are in `set` (unordered), remember to sort them.

forte/data/data_store.py
```diff
@@ -937,7 +940,46 @@ def _parse_onto_file(self):
...
+            if entry_name in DataStore._type_attributes:
+                continue
+
+            attr_dict = {}
+            idx = constants.ATTR_BEGIN_INDEX
+            if "attributes" in onto:
+                for d in onto["attributes"]:
+                    name = d["name"]
+                    attr_dict[name] = idx
+                    idx += 1
```
hunterhector, line 960: You have to pay attention to the details, what is the order of the attributes? Can this be consistent? If we rely on this `dict` order, then it is arbitrary and there is no consistency guarantee.

forte/data/data_store.py
```diff
@@ -281,13 +284,13 @@ def _get_type_attribute_dict(self, type_name: str) -> Dict[str, int]:
         """
         return self._get_type_info(type_name)["attributes"]
 
-    def _get_type_parent(self, type_name: str) -> str:
+    def _get_type_parent(self, type_name: str) -> Set[str]:
         """Get a set of parent names of an entry type. The set is a subset of all
         ancestors of the given type.
```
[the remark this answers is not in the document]
qinzzz: "parent_class" is stored for every class in type_attributes and updated dynamically. This function is for accessing the data structure efficiently. For example, `DataStore._type_attributes[type_name][
            "parent_class"
        ]` can use this function call.

forte/data/data_store.py
```diff
@@ -937,7 +940,46 @@ def _parse_onto_file(self):
         """
```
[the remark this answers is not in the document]
qinzzz: 3. Does `files containing its parent entry` sound better? I gave an example after this sentence, maybe it will help understanding.
4. I am using `ontologyGenerator.parse_ontology_spec(
            self._onto_file_path, merged_scheme, merged_dict)` here, and this function will import the files along the ontology tree and store them in `merged_scheme`. When importing a class from file A, it will import every class in A.

forte/data/data_store.py
qinzzz: I used parse_ontology_spec in OntologyCodeGenerator, but the returned merged_scheme does not have the same structure as used in DataStore. This piece of code is trying to convert them to store in `_type_attributes`.

forte/data/data_store.py
hunterhector: this is still there

forte/data/data_store.py
hunterhector: this is still there.

forte/data/data_store.py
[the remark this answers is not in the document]
hunterhector: 1. This function does not add any "efficiency", it is just a shorthand.
2. As I mentioned in my comment, please understand that this function will *confuse* people who uses the API. This function only provides an uncomplete parent set and does not provide methods to check things dynamically. If someone calls this without this knowledge, it is just gonna make mistakes. We should provide an actual useful utility like `_is_subtype(type_name, parent_type_name)`

forte/data/data_store.py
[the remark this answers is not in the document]
hunterhector: 3. I actually (still) don't understand what you are trying to say here. What files? I didn't see any file operations in this function. Try to rethink/re-read what you write, and imagine if someone other than the developers can understand what is written.
4.
  - "import the files along the ontology", what do you mean by "import the files"? what file? Why are there files on the ontology tree?
  - No, this will only ensure that we import entry classes provided in the ontology definition, it does not have the ability to import everything under the same namespace.

forte/data/data_store.py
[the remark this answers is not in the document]
qinzzz: We already have `_is_subclass`. This function can just be deleted.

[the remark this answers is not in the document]
hunterhector: I see. You can provide similar utilities like `_is_subclass` if needed, but this ambiguous function call would better be removed.

forte/data/data_store.py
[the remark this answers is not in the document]
qinzzz: When the provided ontology spec file is `forte/ontology_specs/medical.json`, the output of `ontologyGenerator .parse_ontology_spec` would be every entry type in `ft.onto.base_ontology` plus type in `ftx.onto.clinical`, because "base_ontology.json" is in the imports path, which is because one of the type's parent_entry is in "base_ontology.json", as far as I understand. The function will recursively import every file in the `imports` path until it gets to the current file.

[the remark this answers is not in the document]
hunterhector: Not sure if I understand now, but you can consider reusing the wording in the original docstring you used [here](https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_generator.py#L613).

Also, note that no matter what we do, we cannot guarantee that we import everything under a certain namespace, we can only import everything written in the specification file.

So now I see that you mean the "files" during the "import" section? It is actually not really easy to import them when running this function. The call of `ontologyGenerator.parse_ontology_spec` is safe. Since there are many caveats not considered. For example, where to import the other JSON files? Note that [this need to be set](https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_generator.py#L213) when using this code.

It is much harder to deal with all the imports at this point, so we should take one JSON schema file here only and use that. That's why I suggest using the [merged entry tree here](https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_generator.py#L685). Note that this is used in another place [here](https://github.com/asyml/forte/blob/642d57680c26fc1a36b31cb560e3f11785f9df11/forte/pipeline.py#L674). I am not sure if `merged entry tree` will give you all the classes (including the ones in top or core, but you should check that).

The process is simpler:
1. The user obtains a single ontology specification file that contains all the entry definitions he wanted to use, either manually or through the `-m` option described [here](https://asyml-forte.readthedocs.io/en/latest/toc/ontology_generation.html#ontology-generation-steps)
2. We take this one file and import all the types. In the docstring, write it very clear that we can only import the types in the file, remember, there is no way we can import everything under a certain namespace.


If you really want to use `ontologyGenerator.parse_ontology_spec`, then you need to either:
1. Make it very clear that we need to pass the import path (where the import ontology files will be placed). You can try to use a few default import paths (like the same directory), but throw exceptions if any one is not found.
2. Or make it very clear that we will not use any types not in the spec file provided (will not follow imports)

hunterhector:
let's pass the CI?

hunterhector:
@mylibrar

forte/data/data_store.py
```diff
@@ -1800,26 +1810,69 @@ def _parse_onto_file(self):
...
+            # sort the attribute dictionary
+            for d in sorted(entry_node.attributes):
+                name = d
+                attr_dict[name] = idx
+                idx += 1
+
+            entry_dict = {}
+            entry_dict["parent_class"] = set()
+            entry_dict["parent_class"].add(entry_node.parent.name)
+            entry_dict["attributes"] = attr_dict
```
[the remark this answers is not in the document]
mylibrar: They are added to `forte/common/constants.py`:
```python
TYPE_ATTR_KEY = "attributes"
PARENT_CLASS_KEY = "parent_class"
```

forte/data/data_store.py
```diff
@@ -1800,26 +1810,69 @@ def _parse_onto_file(self):
...
         information accordingly.
 
-        For every ontology, this function will import paths containing its parent entry and
-        merge all classes contained in the imported file into the dictionary. For example,
-        if an ontology has a parent entry in ``ft.onto.base_ontology``, all classes in
-        ``ft.onto.base_ontology`` will be imported and stored in the internal dictionary.
-        A user can use classes both in the ontology specification file and their parent
-        entries's paths.
+        The ontology specification file should contain all the entry definitions users
+        wanted to use, either manually or through the `-m` option. This function will
```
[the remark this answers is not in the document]
mylibrar: Added a little context [here](https://github.com/asyml/forte/pull/692/files#diff-f1c68405498aa1cba27f81862f28d3edecfdc2d8501394bb3cac9d83dfa1283dR1898).

tests/forte/data/data_store_test.py
```diff
@@ -1175,6 +1185,37 @@ def test_is_subclass(self):
...
+            "ft.onto.test.MedicalEntityMention": {
+                "attributes": {
+                    "umls_entities": 4,
+                    "umls_link": 5,
+                },
+                "parent_class": {"ft.onto.test.EntityMention"},
+            }
+        }
+        data_store_from_file = DataStore(onto_file_path="tests/forte/data/ontology/test_specs/test_check_onto_file.json")
+        self.assertDictContainsSubset(expected_type_attributes, data_store_from_file._type_attributes)
```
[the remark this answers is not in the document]
hunterhector: which types are not in `data_store_from_file._type_attributes`? Will this test still succeed when `data_store_from_file._type_attributes` is an empty set?

hunterhector (commented):
For this PR let's focus on making the test case covering what we want.

tests/forte/data/data_store_test.py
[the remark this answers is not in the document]
mylibrar: This line is trying to verify that `data_store_from_file._type_attributes` contains all the items in `expected_type_attributes`.

Issue #666, opened 2022-03-02
implement co_iterator in DataStore


qinzzz:
**Is your feature request related to a problem? Please describe.**
This is a new function in DataStore class. Given two or more type names, iterator their entry lists from beginning to end together.

**Depends on**
- List issues that this one depends on.

**Describe the solution you'd like**
The function input will be a list of target type_name strings. For every single type, their entry lists are sorted by the `begin` and `end` fields. The co_iterator function will iterate those sorted lists together, and yield each entry in sorted order. This tasks is quite similar to merging several sorted list to one sorted list. If two entries have both the same `begin` and `end` field, then their order is decided by the order of user input type_name (the type that first appears in the target type list will return first).

**Describe alternatives you've considered**

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

PR #703, opened 2022-03-29
implemented DataStore co_iterator

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/666.

### Description of changes
* implemented a `co_iterator` in `DataStore` 
* Internally, I used min heap to track the entries and the priority is determined by tuple (`begin_index`,`end_index`, type name index in the input `type_names`)  

### Possible influences of this PR
* help iterating two or more types of data entries in order efficiently 

### Test Conducted
* tested sorting by begin_index, end_index, type name index in the input `type_names` separately
* tested  input type names  that doesn't exist in `type_names`


forte/data/data_store.py
```diff
@@ -490,6 +492,72 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+        n = len(type_names)
+        # suppose the length of type_names is N and the length of entry list of
+        # one type is M
+        # then the time complexity of using min-heap to iterate
+        # is O(M*log(N))
+
+        entries = []
+        for tn in type_names:
+            if tn in self.__elements:
+                entries.append(self.__elements[tn][0])
```
hunterhector, line 523: We will probably use some constants defined here: https://github.com/asyml/forte/pull/635/files#diff-bed4b362357669e7f878c5c84fde9f9b6ed8f1a1d88b81a3e143a547ac8fa9c1, probably blocked until then.

forte/data/data_store.py
```diff
@@ -490,6 +492,72 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+            else:
+                raise ValueError(
+                    f"Input parameter types name {tn} is not"
+                    "available. Please input avaible ones in this DataStore"
+                    f"object: {list(self.__elements.keys())}"
+                )
+        pointers = [0] * n  # record the current entry index for elements
+        # compare tuple (begin, end, order of type name in input argument type_names)
+        h = []
+        begin_idx, end_idx, type_name_idx = 0, 1, 3
```
hunterhector, line 533: same as above, would probably need to use constants.

forte/data/data_store.py
```diff
@@ -490,6 +492,72 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+        for p_idx in range(n):
+            h.append(
+                (
+                    (entries[p_idx][begin_idx], entries[p_idx][end_idx], p_idx),
+                    entries[p_idx][type_name_idx],
+                )
+            )  # p_idx
+
+        heapify(h)
```
hunterhector, line 542: Can you add some comments explaining what the heap is doing? And btw, could we build the heap on the fly instead of doing it later.

forte/data/data_store.py
hunterhector: The constants for entry information are here: https://github.com/asyml/forte/blob/master/forte/common/constants.py

If you didn't find the const you need, you can consider add one.

hunterhector: The constants for entry information are here: https://github.com/asyml/forte/blob/master/forte/common/constants.py

If you didn't find the const you need, you can consider add one.

forte/data/data_store.py
```diff
@@ -493,6 +495,129 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
         # Otherwise, use ``index_id`` to find the index of the entry.
         raise NotImplementedError
 
+    def co_iterator(self, type_names: List[str]) -> Iterator[List]:
+        r"""
+        Given two or more type names, iterate their entry lists from beginning to end together.
```
hunterhector, line 500: 1.  Need to emphasize that `co_iterator` works on annotation-like types. Actually, it is better to reflect that in the function name, maybe rename this to `co_iterator_annotation_like` (In the future we may create iterators for more complex objects like bounding boxes)
2. Back-tip the variable names.

For entries that have the exact same `begin`, `end`, and `type_name`, what's the order? It is fine to say the order is undefined (or arbitrary), but this behavior needs to be mentioned.

forte/data/data_store.py
```diff
@@ -493,6 +495,129 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+        # is O(M*log(N))
+
+        # Initialize the first entry of all entry lists
+        # it avoids empty entry lists or non-existant entry list
+        first_entries = []
+        for tn in type_names:
+            if tn in self.__elements:
+                if len(self.__elements[tn]) > 0:
+                    first_entries.append(self.__elements[tn][0])
+                else:
```
hunterhector, line 542: Here you can actually simplify a little bit by catching the `IndexError` to find out that the entry list is empty.

forte/data/data_store.py
```diff
@@ -493,6 +495,129 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+
+        while h:
+            # NOTE: we push the tuple to the heap
+            # but not the actual entry. But we can retrieve
+            # the entry by the tuple's data.
+            # In the following comments,
+            # the current entry means the entry that
+            # the popped tuples represents
+            # the current entry list means the entry list
+            # where the current entry locates at.
```
hunterhector, line 590: These couple lines of comments are a bit confusing.

forte/data/data_store.py
```diff
@@ -493,6 +495,129 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+
+            # retrieve the popped entry tuple (minimum item in the heap)
+            # and get the p_idx (the index of the current entry list in self.__elements)
+            entry_tuple = heappop(h)
+            type_name: str = entry_tuple[1]
+            _, _, p_idx = entry_tuple[0]
```
hunterhector, line 596: Just a small tip, I think you can do:
`(_, _, p_idx), type_name = entry_tuple`

forte/data/data_store.py
```diff
@@ -493,6 +495,129 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+            # get the index of current entry
+            # and locate the entry represented by the tuple for yielding
+            pointer = pointers[p_idx]
+            entry = self.__elements[type_name][pointer]
+            # check whether there is next entry in the current entry list
+            # if there is, then we push the new entry's tuple into the heap
+            if pointer + 1 < len(self.__elements[type_name]):
+                pointers[p_idx] = pointer + 1
```
hunterhector, line 604: How about `pointers[p_idx] += 1`?

forte/data/data_store.py
feipenghe: edited based on the review.

forte/data/data_store.py
```diff
@@ -493,6 +496,137 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+            # raise error if it's empty list
+            if not self.__elements[tn]:
+                raise IndexError(
+                    f"Entry list of type name ({tn}) is"
+                    " empty. Please check data in this DataStore "
+                    " to see if empty list is expected"
+                    f" or remove {tn} from input parameter type_names"
+                )
+            else:
+                first_entries.append(self.__elements[tn][0])
```
hunterhector, line 561: What I am suggesting is actually 
```
        for tn in type_names:
           element_list = []
           try:
               element_list = self.__elements[tn]               
               first_entries.append(element_list[0])               
           except KeyError: 
               raise ValueError(
                    f"Input argument `type_names` to the function contains"
                    f" a type name [{tn}], which is not recognized."
                    f" Please input available ones in this DataStore"
                    f" object: {list(self.__elements.keys())}"
                )
          except IndexError:
                raise IndexError(
                    f"Entry list of type name ({tn}) is"
                    " empty. Please check data in this DataStore "
                    " to see if empty list is expected"
                    f" or remove {tn} from input parameter type_names"
                )
```

hunterhector: @qinzzz @wanglec could you comment on whether `self.__elements[tn]` can be an empty list. Consider a scenario where you delete all elements of that type, will this list entry be deleted? Do we need to revisit the `delete` logic?

I would advise that `self.__elements[tn]` won't be an empty list, and the key should be deleted. This not also help making the serialization case (no point to serialize information for empty lists).

wanglec: I think it makes sense to delete the empty list and the corresponding key in `delete_entry`.

hunterhector: @feipenghe let's make modifications in `delete_entry` to ensure that.

forte/data/data_store.py
```diff
@@ -493,6 +496,134 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+                if self.__elements[tn]:
+                    first_entries.append(self.__elements[tn][0])
+                    pointers[i] = 0
+            except KeyError as e:  # self.__elements[tn] will be catched here.
+                raise ValueError(
+                    f"Input argument `type_names` to the function contains"
+                    f" a type name [{tn}], which is not recognized."
+                    f" Please input available ones in this DataStore"
+                    f" object: {list(self.__elements.keys())}"
+                ) from e
```
hunterhector, line 559: We can still keep the `IndexError` even though we know `__elements` should be non-empty, but one more catch here will sometimes help us debug unexpected behaviors

forte/data/data_store.py
feipenghe: Edited it.

forte/data/data_store.py
```diff
@@ -493,6 +498,137 @@ def get_entry(self, tid: int) -> Tuple[List, int, int]:
...
+        # the index of the entry type name in input parameter ``type_names``)
+        h: List[Tuple[Tuple[int, int, int], str]] = []
+        for p_idx in range(n):
+            entry_tuple = (
+                (
+                    first_entries[p_idx][constants.BEGIN_INDEX],
+                    first_entries[p_idx][constants.END_INDEX],
+                    p_idx,
+                ),
+                first_entries[p_idx][constants.ENTRY_TYPE_INDEX],
```
hunterhector, line 584: Is the sorting here performed in the following order:
1. compare begin
2. compare end
3. compare `p_idx`
4. compare name

In docstring, it says the order is:
```
            - start index of the entry.
            - end index of the entry.
            - the index of the entry type name in input parameter ``type_names``.
```

Notice `p_idx` will be different for different types, so the sorting essentially prefers pointers that are smaller, right? This is kinda a weird behavior.

feipenghe: `p_idx` is actually the index of the type name in input parameter `type_names`.
Suppose `type_names = ["Token", "Document"]` then the `p_idx` for `Token ` is 0 and `p_idx` for `Document ` is 1.

hunterhector: i see, so you put it there for sorting purposes.

hunterhector (approved):

Issue #700, opened 2022-03-25
example chat bot configuration error. #700


bhaskar2443053:
**Describe the bug**
following the read.me in chartbot example. encounters configuration error.

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
python download_models.py --model-name chatbot-bert
python download_models.py --model-name indexer
python download_models.py --model-name srl
python chatbot_example.py
error appears on 4th command:
 
(forte) bhaskarrao@LM00824BRAO chatbot % python chatbot_example.py
Traceback (most recent call last):
  File "/Users/bhaskarrao/opt/anaconda3/envs/forte/lib/python3.9/site-packages/forte/common/configurable.py", line 105, inmake_configs
    final_configs = Config(
  File "/Users/bhaskarrao/opt/anaconda3/envs/forte/lib/python3.9/site-packages/texar/torch/hyperparams.py", line 160, in __init__
    parsed_hparams = self._parse(
  File "/Users/bhaskarrao/opt/anaconda3/envs/forte/lib/python3.9/site-packages/texar/torch/hyperparams.py", line 226, in _parse
    raise ValueError(
ValueError: Unknown hyperparameter: query_pack_name. Only hyperparameters named 'kwargs' hyperparameters can contain new entries undefined in default hyperparameters.
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/Users/bhaskarrao/Documents/forte/examples/chatbot/chatbot_example.py", line 128, in <module>
    main(all_config)
  File "/Users/bhaskarrao/Documents/forte/examples/chatbot/chatbot_example.py", line 75, in main
    query_pipeline = setup(config)
  File "/Users/bhaskarrao/Documents/forte/examples/chatbot/chatbot_example.py", line 44, in setup
    query_pipeline.add(component=SearchProcessor(), config=config.searcher)
  File "/Users/bhaskarrao/opt/anaconda3/envs/forte/lib/python3.9/site-packages/forte/pipeline.py", line 829, in add
    self.component_configs.append(component.make_configs(config))
  File "/Users/bhaskarrao/opt/anaconda3/envs/forte/lib/python3.9/site-packages/forte/common/configurable.py", line 109, inmake_configs
    raise ProcessorConfigError(
forte.common.exception.ProcessorConfigError: Configuration error for the processor forte.processors.ir.search_processor.SearchProcessor.

**Expected behavior**

**Screenshots**
[...]

PR #711, opened 2022-03-30
fix chatbot example

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/700. 

### Description of changes
* add required packages in `README.md`
* ~~add `__init__.py` in `ftx` for importing~~
* corrected some updated configuration

After the change, the chatbot example should be able to run without configuration error.

### Test Conducted


examples/chatbot/README.md
```diff
@@ -1,23 +1,39 @@
...
+After setting up `forte` environment by installing packages in `requirements.txt` in the project root folder.
+```
+pip install termcolor
+pip install tensorflow==2.4 # version compantible with numpy==1.19.5
+pip install faiss-gpu
+```
+
+Then we need to install forte external pacakges
+```
+pip install src/faiss
```
hunterhector, line 26: This instruction is not clear, since it didn't specify the repo you need to download and the directory you are in.

hunterhector (commented):
We deliberately remove `__init__.py` from `ftx`, and because we use find_namespace_package `https://github.com/asyml/forte/blob/master/setup.py#L20`. It should work with `pip install` even if the file does not present.

It is important that we don't have this file to avoid conflicting with our other projects.

examples/chatbot/README.md
feipenghe: changed the description.

bhaskar2443053:
HI @feipenghe ,
the chatbot example still throws this error, while running it.

image

PR #714, opened 2022-03-30
Move get entry attributes to DataStore

### LINKED ISSUE #713: Move get entry attributes to DataStore
get_entry_attributes_by_class is implemented in a new class, which is not unnecessary. We should make it a private class of DataStore, as It is only used when adding new types in DataStore.

**Depends on**

Delete EntryTypeGenerator and move the function to DataStore.

**Additional context**

qinzzz:
This PR fixes #713 . 

### Description of changes
Move get_entry_attributes_by_class from TypeEntryGenerator to DataStore. Delete TypeEntryGenerator.

### Test Conducted


hunterhector (dismissed):

hunterhector (approved):
Approving for now.

One comment: the `_type_attributes` contains information about the parent class. During inheritance check, we also need to check for all ancestors, we may need some additional fields to cache that computation.

Issue #719, opened 2022-03-31
Move out dependency of texar to extra_requires in forte


feipenghe:
**Is your feature request related to a problem? Please describe.**
We want to separate out the texar-pytorch dependency to extra_requires in forte. Right now the following modules include texar:

examples/

forte/train_preprocessor.py

forte/common - HParams, which will be moved out to [asyml-utilities](https://github.com/asyml/asyml-utilities/pull/1)

forte/data/

forte/models/

forte/processors/

We can address each of the module listed above

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #720, opened 2022-03-31
Move texar to extra require

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/719. 

### Description of changes
* raise `ImportError` with suggestions when the extra library `texar-pytorch` is not installed

### Test Conducted


examples/tagging/tagging_trainer.py
```diff
@@ -15,7 +15,6 @@
 from typing import Iterator, Dict
 
 import torch
-from texar.torch.data import Batch
```
hunterhector, line 18: For examples, you should should tell the users to install texar instead of throwing the exception during excecution

forte/common/configuration.py
```diff
@@ -18,7 +18,15 @@
 """
 from typing import Dict
 
-from texar.torch import HParams
+
+try:
+    from texar.torch import HParams
```
hunterhector, line 23: This is not the correct solution, please talk tou @mylibrar

tests/forte/data/vocabulary_test.py
```diff
@@ -17,12 +17,21 @@
...
 from ddt import ddt, data, unpack
-from texar.torch.data import SpecialTokens
+
 
 from forte.common import InvalidOperationException
 from forte.data import dataset_path_iterator
 from forte.data.vocabulary import Vocabulary, FrequencyVocabFilter
 
+try:
+    from texar.torch.data import SpecialTokens
```
hunterhector, line 27: No need to remove from test cases

hunterhector:
@feipenghe @mylibrar Please notice that the solution cannot be simply throwing exceptions at run time. 

If we decide Texar would be extra required, we need to provide a good extra requirement install option in `setup.py`, and in the documentation here: https://github.com/asyml/forte#download-and-installation, explain which ones require which extra requirement.

Note that you also cannot simply say `pip install forte[texar]`, it is not intuitive. It should be related to the function itself. For example, if the users need to use the models (srl, ner), the extra requirement should be `pip install forte[forte-models]`. If the function is related to data augmentation models, the extra requirement could be `pip install forte[data-aug-models]`.

Here is the goal: the user shouldn't encounter any exception if he follows the documentation. The exceptions are pointers for a user to notice the documentation, not documentations by themselves.

CONTRIBUTING.md
```diff
@@ -135,7 +135,7 @@ the [Google Python Style guide](http://google.github.io/styleguide/pyguide.html)
...
 
-We also recommend using tools `pre-commit` that automates the checking process before each commit since checking format is a repetitive process. We have the configuration file `.pre-commit-config.yaml` that lists several plugins including `black` to check format in the project root folder. Developers only need to install the package by `pip install pre-commit`.
+We also recommend using tools `pre-commit` that automates the checking process before each commit since checking format is a repetitive process. We have the configuration file `.pre-commit-config.yaml` that lists several plugins including `black` to check format in the project root folder. Developers only need to install the package by `pip install pre-commit`. All the package versions in the `.pre-commit-config.yaml` must be consistent with package versions in [workflow configuration](https://github.com/asyml/forte/blob/master/.github/workflows/main.yml). For example, `black` package version should be set to the same.
```
hunterhector, line 138: it will be easier if the pre-commit hook can take the version from the workflow file or a version file, making it easier to sync up.

README.md
```diff
@@ -50,11 +50,17 @@ pip install src/spacy
...
 
-* `pip install forte[ner]`: Install packages required for [ner_trainer](https://github.com/asyml/forte/blob/master/forte/trainer/ner_trainer.py)
+* `pip install forte[forte-models-ner]`: Install packages required for [ner_trainer](https://github.com/asyml/forte/blob/master/forte/trainer/ner_trainer.py)
 * `pip install forte[test]`: Install packages required for running [unit tests](https://github.com/asyml/forte/tree/master/tests).
 * `pip install forte[example]`: Install packages required for running [forte examples](https://github.com/asyml/forte/tree/master/examples).
 * `pip install forte[wikipedia]`: Install packages required for reading [wikipedia datasets](https://github.com/asyml/forte/tree/master/forte/datasets/wikipedia).
-* `pip install forte[augment]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
+* `pip install forte[forte-processors-data_aug]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
```
hunterhector, line 57: The names are a bit long here, and probably not semantically related, we will need to discuss this.

tests/forte/data/entry_type_generator_test.py
```diff
@@ -0,0 +1,131 @@
+import unittest
```
hunterhector, line 1: Please check. this class should have been deleted in master.

tests/forte/data/data_store_test.py
```diff
@@ -138,8 +138,8 @@ def setUp(self) -> None:
         }
```
hunterhector, line 138: I don't think we should touch any `data_store` files in this PR.

forte/data/entry_type_generator.py
```diff
@@ -0,0 +1,98 @@
+#  Copyright 2022 The Forte Authors. All Rights Reserved.
```
hunterhector, line 1: Let's not touch irrelevant files in this PR.

setup.py
```diff
@@ -39,15 +39,37 @@
         'importlib-resources>=5.1.4;python_version<"3.7"',
     ],
     extras_require={
-        "ner": ["torch>=1.1.0", "torchtext==0.4.0", "tqdm>=4.36.1"],
+        "forte-models-ner": [
```
hunterhector, line 42: forte-models

setup.py
```diff
@@ -39,15 +39,37 @@
...
         # transformers 4.10.0 will break the translation model we used here
-        "augment": ["transformers>=3.1, <=4.9.2", "nltk"],
         "stave": ["stave>=0.0.1.dev12"],
         "audio_ext": ["soundfile>=0.10.3"],
-        "remote": ["fastapi>=0.65.2", "uvicorn>=0.14.0",],
+        "remote": [
+            "fastapi>=0.65.2",
+            "uvicorn>=0.14.0",
+        ],
+        "forte-models-data_rl": ["texar-pytorch>=0.1.1"],
```
hunterhector, line 62: forte-rl-aug

setup.py
```diff
@@ -39,15 +39,37 @@
...
+        "forte-models-srl": ["texar-pytorch>=0.1.1"],  # srl and srl_new
+        "forte-processors-misc": ["texar-pytorch>=0.1.1"],
```
hunterhector, line 64: https://github.com/asyml/forte/blob/master/forte/processors/misc/vocabulary_processor.py#L50, copy the special tokens when texar not found

setup.py
```diff
@@ -39,15 +39,37 @@
...
+        "forte-processors-nlp": ["texar-pytorch>=0.1.1"],
```
hunterhector, line 65: subword_tokenize

hunterhector: move https://github.com/asyml/forte/blob/master/forte/processors/nlp/subword_tokenizer.py#L22 to "try" block

setup.py
```diff
@@ -39,15 +39,37 @@
...
-        "augment": ["transformers>=3.1, <=4.9.2", "nltk"],
         "stave": ["stave>=0.0.1.dev12"],
         "audio_ext": ["soundfile>=0.10.3"],
-        "remote": ["fastapi>=0.65.2", "uvicorn>=0.14.0",],
+        "remote": [
+            "fastapi>=0.65.2",
+            "uvicorn>=0.14.0",
+        ],
+        "forte-models-data_rl": ["texar-pytorch>=0.1.1"],
+        "forte-models-srl": ["texar-pytorch>=0.1.1"],  # srl and srl_new
```
hunterhector, line 63: not needed

setup.py
```diff
@@ -39,15 +39,37 @@
...
+        "forte-processors-misc": ["texar-pytorch>=0.1.1"],
+        "forte-processors-nlp": ["texar-pytorch>=0.1.1"],
+        "forte-processors-third_party": ["texar-pytorch>=0.1.1"],
```
hunterhector, line 66: texar-encoder

Add try block here.
https://github.com/asyml/forte/blob/master/forte/processors/third_party/pretrained_encoder_processors.py

setup.py
```diff
@@ -39,15 +39,37 @@
...
+        "forte-processors-data_aug": [
+            "transformers>=3.1, <=4.9.2",
```
hunterhector, line 68: mt

setup.py
```diff
@@ -39,15 +39,37 @@
...
+            "nltk",
+            "texar-pytorch>=0.1.1",
```
hunterhector, line 70: uda

setup.py
```diff
@@ -39,15 +39,37 @@
...
+            "uvicorn>=0.14.0",
+        ],
+        "forte-models-data_rl": ["texar-pytorch>=0.1.1"],
+        "forte-models-srl": ["texar-pytorch>=0.1.1"],  # srl and srl_new
+        "forte-processors-misc": ["texar-pytorch>=0.1.1"],
+        "forte-processors-nlp": ["texar-pytorch>=0.1.1"],
+        "forte-processors-third_party": ["texar-pytorch>=0.1.1"],
+        "forte-processors-data_aug": [
+            "transformers>=3.1, <=4.9.2",
+            "nltk",
```
hunterhector, line 69: augment

README.md
```diff
@@ -50,13 +50,27 @@ pip install src/spacy
...
-* `pip install forte[augment]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
+* `pip install forte[data_aug]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
+
+* `pip install forte[subword_tokenize]`: Install packages required for [subword_tokenizer](https://github.com/asyml/forte/tree/master/forte/processors/nlp/subword_tokenizer.py)
+* `pip install forte[texar-encoder]`: Install packages required for [texar encoder](https://github.com/asyml/forte/tree/master/forte/processors/third_party/pretrained_encoder_processors.py)
 * `pip install forte[stave]`: Install packages required for [StaveProcessor](https://github.com/asyml/forte/blob/master/forte/processors/stave/stave_processor.py).
 * `pip install forte[audio_ext]`: Install packages required for [AudioReader](https://github.com/asyml/forte/blob/master/forte/data/readers/audio_reader.py).
+* `pip install forte[remote]`: Install packages required for [AudioReader](https://github.com/asyml/forte/blob/master/forte/data/readers/audio_reader.py).
```
hunterhector, line 63: Not for `audio_reader`

README.md
```diff
@@ -50,13 +50,27 @@ pip install src/spacy
...
+
+* `pip install forte[extractor]`: Install packages required for [extractor](https://github.com/asyml/forte/blob/master/forte/data/extractors)
+* `pip install forte[rl-aug]`: Install packages required for [data_rl](https://github.com/asyml/forte/tree/master/forte/models/da_rl)
+* `pip install forte[uda]`: Install packages required for [uda](https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/UDA.py).
+* `pip install forte[mt]`: Install packages required for [mt](https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/machine_translator.py).
+* `pip install forte[dict]`: Install packages required for [dict](https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/dictionary.py).
```
hunterhector, line 69: This is actually in the `data_aug` module, should we merge them?

README.md
```diff
@@ -50,13 +50,27 @@ pip install src/spacy
...
+
+* `pip install forte[emb_sim_rep_op]`: Install packages required for [AudioReader](https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/embedding_similarity_replacement_op.py).
```
hunterhector, line 71: you are writing `AudioReader` again. And this is also in the `data_aug` module, maybe simply merge too?

forte/processors/stave/stave_processor.py
```diff
@@ -29,8 +29,18 @@
...
-from stave_backend.lib.stave_project import StaveProjectWriter
+
+try:
+    from stave_backend.lib.stave_viewer import StaveViewer
+    from stave_backend.lib.stave_project import StaveProjectWriter
+except ImportError:
+    print(
+        " `stave` is not installed correctly."
+        " Please refer to [extra requirement for stave_processors.py](stave"
+        " extra requirement hyperlink])"
```
hunterhector, line 40: so are we going to add a hyperlink here?

setup.py
```diff
@@ -39,15 +39,50 @@
...
+        "extractor": ["texar-pytorch>=0.1.1"],
+        "rl-aug": ["texar-pytorch>=0.1.1"],
+        "uda": [
+            "texar-pytorch>=0.1.1",
+        ],
+        "mt": [
+            "transformers>=3.1",
+        ],
+        "dict": ["nltk"],
+        "emb_sim_rep_op": ["texar-pytorch>=0.1.1"],
```
hunterhector, line 85: So, `rl-aug`, `uda`, `mt`, `dict`, `emb_sim_rep_op` are all under data aug modules, since we have a section for `data-aug` that have similar dependencies, maybe we just merge them. Actually having less options here would be better.

forte/processors/stave/stave_processor.py
feipenghe: yes. it will be on the next PR as I reported in the https://github.com/asyml/forte/issues/721

README.md
feipenghe: merged it.

README.md
```diff
@@ -50,13 +50,21 @@ pip install src/spacy
...
-* `pip install forte[augment]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
+* `pip install forte[data_aug]`: Install packages required for [data augmentation module](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
+
+* `pip install forte[subword_tokenize]`: Install packages required for [subword_tokenizer](https://github.com/asyml/forte/tree/master/forte/processors/nlp/subword_tokenizer.py)
+* `pip install forte[texar-encoder]`: Install packages required for [texar encoder](https://github.com/asyml/forte/tree/master/forte/processors/third_party/pretrained_encoder_processors.py)
 * `pip install forte[stave]`: Install packages required for [StaveProcessor](https://github.com/asyml/forte/blob/master/forte/processors/stave/stave_processor.py).
 * `pip install forte[audio_ext]`: Install packages required for [AudioReader](https://github.com/asyml/forte/blob/master/forte/data/readers/audio_reader.py).
+* `pip install forte[remote]`: Install packages required for [AudioReader](https://github.com/asyml/forte/processors/misc/remote_processor.py).
```
hunterhector, line 63: still have AudioReader here

hunterhector (commented):
For many of the imports, they are done in the module-level scope, which means importing another file from the same folder (which is the same module because they are connected by __init__) will cause the exceptions to be thrown. These are not expected.

feipenghe:
Tried to move some import into local scope but some imported modules are used in many class methods. Few possible ways:
1. import everywhere when the module is used (many repetitive code), also function signature cannot use it
2. import in class `__init__` method and function signature cannot use it as well.

hunterhector:
> Tried to move some import into local scope but some imported modules are used in many class methods. Few possible ways:
> 
> 1. import everywhere when the module is used (many repetitive code), also function signature cannot use it
> 2. import in class `__init__` method and function signature cannot use it as well.

which ones in specific? There are two ways to deal with this:
1. If this file is more independent, we can move it to another folder (with a meaningful folder/module name)
2. Remove it from the `__init__.py` file of that folder.

In both case, the usage/references to these files need to be changed.

forte/processors/nlp/subword_tokenizer.py
```diff
@@ -55,6 +55,15 @@ def initialize(self, resources: Resources, configs: Config):
         super().initialize(resources, configs)
         if not self.configs.tokenizer_configs.pretrained_model_name:
             raise ValueError("Please specify a pretrained bert model")
+        try:
+            from texar.torch.data.tokenizers.bert_tokenizer import BERTTokenizer
+        except ImportError:
+            print(
```
hunterhector, line 61: no `print`

feipenghe:
> > Tried to move some import into local scope but some imported modules are used in many class methods. Few possible ways:
> > 
> > 1. import everywhere when the module is used (many repetitive code), also function signature cannot use it
> > 2. import in class `__init__` method and function signature cannot use it as well.
> 
> which ones in specific? There are two ways to deal with this:
> 
> 1. If this file is more independent, we can move it to another folder (with a meaningful folder/module name)
> 2. Remove it from the `__init__.py` file of that folder.
> 
> In both case, the usage/references to these files need to be changed.

For example, https://github.com/asyml/forte/blob/b748ed71794850461800fab586302dd620352351/forte/models/da_rl/aug_wrapper.py and https://github.com/asyml/forte/blob/b748ed71794850461800fab586302dd620352351/forte/processors/data_augment/algorithms/UDA.py

feipenghe:
> > > Tried to move some import into local scope but some imported modules are used in many class methods. Few possible ways:
> > > 
> > > 1. import everywhere when the module is used (many repetitive code), also function signature cannot use it
> > > 2. import in class `__init__` method and function signature cannot use it as well.
> > 
> > 
> > which ones in specific? There are two ways to deal with this:
> > 
> > 1. If this file is more independent, we can move it to another folder (with a meaningful folder/module name)
> > 2. Remove it from the `__init__.py` file of that folder.
> > 
> > In both case, the usage/references to these files need to be changed.
> 
> For example, https://github.com/asyml/forte/blob/b748ed71794850461800fab586302dd620352351/forte/models/da_rl/aug_wrapper.py and https://github.com/asyml/forte/blob/b748ed71794850461800fab586302dd620352351/forte/processors/data_augment/algorithms/UDA.py

`__init__.py` for `uda.py` doesn't seem to import it so I think I can just import in the head of the file. For `aug_wrapper` I think I can remove the importing from its `__init__.py` as well?

examples/tagging/tagging_trainer.py
```diff
@@ -130,6 +131,15 @@ def train(self):
 
         while epoch < self.config_data.num_epochs:
             epoch += 1
+
+            try:
+                from texar.torch.data import Batch
```
hunterhector, line 136: I think this is better imported outside the `while` loop. You can actually install it under the class scope.

examples/tagging/tagging_trainer.py
```diff
@@ -130,6 +131,15 @@ def train(self):
...
+            except ImportError:
+                print(
```
hunterhector, line 138: don't use print.

hunterhector:
Also, make sure you test the imports after installing Forte without texar.

forte/processors/misc/vocabulary_processor.py
```diff
@@ -27,6 +24,17 @@
 ]
 
 
+class SpecialTokens:
```
mylibrar, line 27: Now we can import this from asyml-utilities
```python
from asyml_utilities.special_tokens import SpecialTokens
```

setup.py
```diff
@@ -32,22 +32,47 @@
         "funcsigs>=1.0.2",
         "typed_ast>=1.4.3",
         "jsonschema>=3.0.2",
-        "texar-pytorch",
```
mylibrar, line 35: We might want to put `asyml-utilities` here

forte/train_preprocessor.py
```diff
@@ -18,7 +18,16 @@
...
 
 import torch
-from texar.torch.data import DataIterator, Batch
+
+try:
+    from texar.torch.data import DataIterator, Batch
+except ImportError as e:
+    raise ImportError(
+        " `texar-pytorch` is not installed correctly."
+        " Please refer to [extra requirement for aug wrapper](pip install forte[data_aug])"
```
hunterhector: not `data_aug`

forte/data/data_pack_dataset.py
```diff
@@ -21,8 +21,17 @@
...
 import torch
-from texar.torch import HParams
-from texar.torch.data import IterDataSource, DatasetBase, Batch
+
+try:
+    from texar.torch.data import IterDataSource, DatasetBase, Batch
+except ImportError as e:
+    raise ImportError(
+        " `texar-pytorch` is not installed correctly."
+        " Please refer to [extra requirement for aug wrapper](pip install forte[data_aug])"
```
hunterhector, line 30: also not `data aug`, please double-check these error messages.

.github/workflows/main.yml
```diff
@@ -84,7 +83,20 @@ jobs:
...
+          pip install --use-feature=in-tree-build --progress-bar off .
+          # needs to remove it in after torch dependency is removed
+          pip uninstall -y texar-pytorch
+          pytest tests/forte/texar_nondependency_test.py
+          # install lastest texar pytorch
+          git clone https://github.com/asyml/texar-pytorch.git
+          cd texar-pytorch
+          pip install --progress-bar off .
+          cd ..
+          rm -rf texar-pytorch
```
hunterhector, line 99: why don't we  conduct these steps before the `Install Texar` or even the `Install deep learning frameworks` step?

feipenghe: Since we still have dependencies on other libraries in the dependency matrix. In this PR, we only need to remove `texar`. In the later PR that we set up a dependency env for different modules, these lines of code will be removed.

mylibrar (approved):

Issue #729, opened 2022-04-08
Quick start guide


feipenghe:
**Describe the bug**
We want a quick start guide that can get users familiar with forte faster.

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

PR #730, opened 2022-04-08
Quick start guide

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/729. 

### Description of changes
* Merge `overview.rst` into `README.md`
* change markdown hyperlinks to the web link directly
* add file linking to `README.md`

### Test Conducted


README.md
```diff
@@ -59,54 +58,105 @@ Some components or modules in forte may require some [extra requirements](https:
...
+            words = re.sub(r'[^\w\s]','',sentence.text).split(" ")
+            # second step: split stripped sentences into words and write
+            # words into the data pack
+            for w in words:
+                Word(pack=input_pack, value=w)
+
+    @classmethod
+    def default_configs(cls) -> Dict[str, Any]:
+        return {"test": "test, successor"}
+```
```
hunterhector, line 133: Maybe explain a little bit what is done here, like "get", or how "Word"s are created and stored.

README.md
```diff
@@ -59,54 +58,105 @@ Some components or modules in forte may require some [extra requirements](https:
...
+class Word(Generics):
+    """A dummy generic type for words."""
+    value: Optional[str] = None
+    def __init__(self, pack, value):
+        super().__init__(pack)
+        self.value = value
+```
+Next we can write a simple customized processor for the first task, split sentences into words.
+There are two steps for the sentence processing.
+First, we need to strip all punctuation as splitted words should not contain them.
```
hunterhector, line 107: Past tense of `split` should still be `split`

README.md
hunterhector: How about just using a class from base_ontology instead of creating a new `Word` class here? I don't think people are ready to understand these classes now.

README.md
```diff
@@ -59,54 +69,114 @@ Some components or modules in forte may require some [extra requirements](https:
...
+
+Forte not only makes it easy to integrate with arbitrary 3rd party tools (Check out these [examples](https://github.com/asyml/forte/tree/master/examples)!), but also brings technology to you by offering a miscellaneous collection of deep learning modules via Texar, and a convenient model-data interface for casting tasks to models.
+
+
+
+* [Examples](https://github.com/asyml/forte/tree/master/examples)
+* [Documentation](https://asyml-forte.readthedocs.io/)
+* Currently we are working on some interesting tutorials such as [MT pipeline](https://github.com/asyml/forte/wiki) and [Text classification pipeline](https://asyml-forte.readthedocs.io/en/latest/notebook_tutorial/text_classification_pipeline.html)
+
+## Project Info
```
hunterhector, line 178: I don't find this `Project Info` header useful

docs/spelling_wordlist.txt
```diff
@@ -160,3 +160,4 @@ LinkType
 GroupType
 PathLike
 ElementType
+splitted
```
hunterhector, line 163: don't add this

feipenghe: It's because the headers under this would be too redundant on the left bar if we keep the original header levels. https://asyml-forte--730.org.readthedocs.build/en/730/

README.md
feipenghe: It's because the headers under this would be too redundant on the left bar if we keep the original header levels. https://asyml-forte--730.org.readthedocs.build/en/730/

README.md
```diff
@@ -59,54 +69,114 @@ Some components or modules in forte may require some [extra requirements](https:
...
+New User doesn't need to pay too much attention about it and can just use it as it is.
 
-for pack in Pipeline().set_reader(
-        TerminalReader()
-).add(
-    SpacyProcessor(), {"processors": ["sentence", "ner"]}
-).initialize().process_dataset():
+```python
+@dataclass
+class Word(Generics):
```
hunterhector, line 97: Let's just don't add the new `Word` class here. It is too deep for "Quick Start", just use the pre-built classes from base_ontology.

README.md
```diff
@@ -59,54 +69,114 @@ Some components or modules in forte may require some [extra requirements](https:
...
+            # first step: strip all punctuation
+            words = re.sub(r'[^\w\s]','',sentence.text).split(" ")
+            # second step: split stripped sentences into words and write
+            # words into the data pack
+            for w in words:
+                # we pass the input pack and word strings
+                # into `Word` class, and the `Word.value` store
+                # `input_pack` will internally add `Word` entries into `input_pack`'s internal storage
+                # while `Word` initializes itself.
+                Word(pack=input_pack, value=w)
```
hunterhector, line 134: Use something more useful here, `value` doesn't help too much.  Just use https://github.com/asyml/forte/blob/master/ft/onto/base_ontology.py#L61.

Another reason is this `Word` class is for test purposes, *never* use it for any production thing. `Word` is not even inheriting an Entry class. So this will give a lot of wrong information.

If you want to assign some useful values, how about just using NLTK pos tagger: https://www.nltk.org/book/ch05.html

feipenghe: OK. I didn't consider the difference between `Generics` and `Entry`. I am found that there is a NLTK tagger in forte wrapper. Maybe I can import `fortex.nltk.NLTKWordTokenizer` and simplify the implementation of  `fortex.nltk.NLTKPOSTagger` and put the code here?

hunterhector: 1. If you use `Word`, you will want to use the `begin` and `end`, not use a meaningless `value` here. This example will just give people the wrong impression
2. Yes, make it really simple, something you expect one can understand the second he read the code. You can also consider use this whitepsace tokenize too. https://github.com/asyml/forte/blob/master/forte/processors/misc/simple_processors.py#L43

README.md
```diff
@@ -59,54 +69,94 @@ Some components or modules in forte may require some [extra requirements](https:
...
-out these [examples](./examples)!), but also brings technology to you by
-offering a miscellaneous collection of deep learning modules via Texar, and a
-convenient model-data interface for casting tasks to models.
-
-## Library Example
-
-A simple code example that runs Named Entity Recognizer from Spacy (required
-installing forte spacy wrapper)
+## Quick Start Guide
+Here we provide an example for user to get started with Forte easier. Consider a case we want to tag tokens' POS and extract entities from the sentence. We can write two processors for the two separate tasks.
```
hunterhector, line 77:
```suggestion
Writing NLP pipelines with Forte is easy. The following example creates a simple pipeline that analyzes the sentences, tokens, and named entities from a piece of text.
```

hunterhector (approved):

PR #732, opened 2022-04-09
add abbreviation replacement data augmentation op and test

### LINKED ISSUE #736: Add abbreviation replacement data augmentation
Add a new replacement-based data augmentation method that replaces phrases with abbreviations.

**Additional context**
Original method [here](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/abbreviation_transformation)

abbeyyyy:
This PR fixes [#736](https://github.com/asyml/forte/issues/736).

### Description of changes
Add [Abbreviation Replacement](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/abbreviation_transformation) Augmentation Method 

This PR provides a new replacement-based data augmentation method

### Test Conducted
Test cases included in abbreviation_replacement_op_test.py


hunterhector:
A couple of preparations for the PR:
1. We will need to create an issue and associate a PR with the issue: https://github.com/asyml/forte/blob/master/CONTRIBUTING.md#pull-requests
2. Accept the PR invitation then you can run the workflow CI without approval

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+                self.data = json.load(json_file)
+
+    def single_annotation_augment(
+        self, input_anno: Annotation
+    ) -> Tuple[bool, str]:
+        r"""
+        This function replaces a word from an abbreviation dictionary.
+
+        Args:
+            input_anno (Annotation): The input annotation.
```
feipenghe, line 71: We no longer need to include input parameter annotation type in docstring.
https://github.com/asyml/forte/blob/master/CONTRIBUTING.md#docstring

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+from forte.common.configuration import Config
+
+__all__ = [
+    "AbbreviationReplacementOp",
+]
+
+
+class AbbreviationReplacementOp(SingleAnnotationAugmentOp):
+    r"""
+    This class is a replacement op utilizing a pre-defined
```
Pushkar-Bhuse, line 34: The docstring should be more comprehensive. This is what the user is going to see if they want to use this DA op.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2020 The Forte Authors. All Rights Reserved.
```
Pushkar-Bhuse, line 1: 2022*

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+            - dict_path: the `url` or the path to the pre-defined
+             abbreviation json file. The key is a word / phrase we want to replace.
+             The value is an abbreviated word of the corresponding key.
+    """
+
+    def __init__(self, configs: Union[Config, Dict[str, Any]]):
+        super().__init__(configs)
+        if "dict_path" in configs.keys():
+            self.dict_path = configs["dict_path"]
+        else:
```
Pushkar-Bhuse, line 50: An if-else loop is not needed here as you are already setting a default value in the `default_configs`

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+            self.data = json.loads(r.text)
+        except requests.exceptions.RequestException:
+            with open(self.dict_path, encoding="utf8") as json_file:
+                self.data = json.load(json_file)
+
+    def single_annotation_augment(
+        self, input_anno: Annotation
+    ) -> Tuple[bool, str]:
+        r"""
+        This function replaces a word from an abbreviation dictionary.
```
Pushkar-Bhuse, line 68: Again, we should add a better description of what this function will do.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+            input_anno: The input annotation.
+        Returns:
+            A tuple, where the first element is a boolean value indicating
+            whether the replacement happens, and the second element is the
+            replaced string.
+        """
+        # If the replacement does not happen, return False.
+        if random.random() > self.configs.prob:
+            return False, input_anno.text
+        if input_anno.text in self.data.keys():
```
Pushkar-Bhuse, line 80: Since you are returning from the function if the program enters the earlier `if` statement, you dont need to add this `if`

Pushkar-Bhuse: Also, I am not sure is this check (`input_anno.text in self.data.keys()`) is necessary.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+        r"""
+        Returns:
+            A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - prob: The probability of replacement,
+              should fall in [0, 1]. Default value is 0.1
+            - dict_path: the `url` or the path to the pre-defined
+              abbreviation json file. The key is a word / phrase we want
+              to replace. The value is an abbreviated word of the
+              corresponding key.
```
Pushkar-Bhuse, line 97: I'd recommend adding the default value of `dict_path` in the docstring as well since this is what will be rendered in the documentation and it would be easier for users to see.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,104 @@
...
+            return False, input_anno.text
+
+    @classmethod
+    def default_configs(cls) -> Dict[str, Any]:
+        r"""
+        Returns:
+            A dictionary with the default config for this processor.
+        Following are the keys for this dictionary:
+            - prob: The probability of replacement,
+              should fall in [0, 1]. Default value is 0.1
```
Pushkar-Bhuse, line 93: The default value below is 0.5. Make sure you check the documentation thoroughly.

tests/forte/processors/data_augment/algorithms/abbreviation_replacement_op_test.py
```diff
@@ -0,0 +1,54 @@
...
+            configs={
+                "prob": 1.0,
+            }
+        )
+
+    def test_replace(self):
+        data_pack = DataPack()
+        text = "see you later"
+        data_pack.set_text(text)
+        token = Token(data_pack, 0, len(text))
```
Pushkar-Bhuse, line 38: The `Token` class is generally used for a single word. When annotating the whole sequence, you should use the `Sentence` class. Also, if your SingleAnnotationOp augments an annotation other than a `Token`, you must specify that in the `default_configs`. For your reference, look at the implementation of the `BackTranslationOp` and its test cases. Even that Op augments sentences.

tests/forte/processors/data_augment/algorithms/abbreviation_replacement_op_test.py
```diff
@@ -0,0 +1,54 @@
...
+        data_pack.add_entry(token)
+
+        augmented_data_pack = self.abre.perform_augmentation(data_pack)
+
+        augmented_token = list(
+            augmented_data_pack.get("ft.onto.base_ontology.Token")
```
Pushkar-Bhuse, line 44: I think you should take the comment above into consideration and rework your test cases accordingly.

Pushkar-Bhuse (commented):
Your implementation of the Op seems fine but it looks like you might have not gotten the underlying intricacies of how to modify the `SingleAnnotationAugmentOp` for different type of annotations. Just take a closer look at that once.

tests/forte/processors/data_augment/algorithms/abbreviation_replacement_op_test.py
hunterhector: We also have Document https://github.com/asyml/forte/blob/master/ft/onto/base_ontology.py#L136 for the whole article.

I know it is just a test case so it doesn't matter too much, but still worth noting.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
abbeyyyy: I was thinking if the input phrase does not have a corresponding abbreviation, an error will occur.

Pushkar-Bhuse (approved):
I think the changes look good.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,106 @@
...
+]
+
+
+class AbbreviationReplacementOp(SingleAnnotationAugmentOp):
+    r"""
+    This class is a replacement op utilizing a pre-defined
+    abbreviation dictionary to replace word or phrase
+    with an abbreviation. The abbreviation dictionary can
+    be user-defined, we also provide a default dictionary.
+    `prob` indicates the probability of replacement.
```
hunterhector, line 38: What does "probability of replacement" mean? For example, if `prob` is 0.4, is the replacement happen 40% of the case or the other way around. Does it mean that 40% of the words will be replaced, etc. Let's specify this clearly.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
hunterhector: 1. When checking dict existence, use `text in self.data`, don't need to call the `keys()`.
2. Now we can see that the `prob` only applies to the annotation that has an abbreviation, which should probably be specified in the class docstring.

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,106 @@
...
+            A tuple, where the first element is a boolean value indicating
+            whether the replacement happens, and the second element is the
+            replaced string.
+
+        """
+        # If the replacement does not happen, return False.
+        if random.random() > self.configs.prob:
+            return False, input_anno.text
+        if input_anno.text in self.data.keys():
+            result: str = self.data[input_anno.text]
```
hunterhector, line 75: Something about this replacement:
1. Do we need to consider the case? Maybe we should lower case your dictionary and user input.
2. How about substrings? For example, in `"see you later": "syl8r"`, what if we have an input "i will see you later", it looks like we won't replace this?

hunterhector: Maybe you need to consider using an [Aho-Corasick](http://en.wikipedia.org/wiki/Aho-Corasick%20algorithm) data sturcture here: https://pyahocorasick.readthedocs.io/en/latest/

forte/processors/data_augment/algorithms/abbreviation_replacement_op.py
```diff
@@ -0,0 +1,106 @@
...
+
+            - prob: The probability of replacement,
+              should fall in [0, 1]. Default value is 0.5.
+
+            - dict_path: the `url` or the path to the pre-defined
+              abbreviation json file. The key is a word / phrase we want
+              to replace. The value is an abbreviated word of the
+              corresponding key. Default dictionary is from a web-scraped
+              slang dictionary
+              ("https://github.com/abbeyyyy/JsonFiles/blob/main/abbreviate.json").
```
hunterhector, line 95: Did we adopt the dictionary from another library?

tests/forte/processors/data_augment/algorithms/abbreviation_replacement_op_test.py
```diff
@@ -0,0 +1,86 @@
...
+                "JsonFiles/main/abbreviate.json",
+                "prob": 1.0,
+            }
+        )
+
+    def test_replace(self):
+        data_pack_1 = DataPack()
+        text_1 = "I will see you later!"
+        data_pack_1.set_text(text_1)
+        phrase_1 = Phrase(data_pack_1, 7, len(text_1) - 1)
```
hunterhector, line 40: I see that you have to first identify the phrase before doing the match, which is not a very typical use case.

Issue #731, opened 2022-04-08
Dependency Matrix test for Modules


feipenghe:
**Describe the bug**
Some components as listed in [here](https://github.com/asyml/forte/blame/master/README.md#L53) have different dependencies. Our current [test case that installs packages all at once](https://github.com/asyml/forte/blob/f73ef0bfc94afb938e3cfcb812cf022890ab692e/.github/workflows/main.yml#L87)

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**
We expect to have a dependency matrix to test those modules' in their own environment with minimally required packages.

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

This is a followup issue of https://github.com/asyml/forte/pull/720

PR #738, opened 2022-04-12
Add dependency matrix in main.yml to test modules in their independent environment

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/731. 

### Description of changes
* Add dependences for individual modules in `setup.py` to `main.yml`

### Test Conducted


.github/workflows/main.yml
```diff
@@ -73,30 +82,25 @@ jobs:
...
           pip install --use-feature=in-tree-build --progress-bar off .[models,test,wikipedia,data_aug,nlp,ir,texar-encoder,stave,audio_ext,remote,extractor]
+      - name: Test models module
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .[test]
+          if [ ${{ matrix.modules.module }} != "" ]
+          then
+            pip install --use-feature=in-tree-build --progress-bar off .[${{ matrix.modules.module }}]
+            pip install --progress-bar off pytest==5.1.3 coverage
+          fi
       - name: Test backbone Forte import test
```
hunterhector, line 96: `backbone Forte import` should be done before testing the `modules`, otherwise, some dependencies will be installed.

.github/workflows/main.yml
```diff
@@ -73,30 +82,25 @@ jobs:
...
         run: |
           # Try to install Forte backbone only and test basic imports.
           pip install --use-feature=in-tree-build --progress-bar off .
           # needs to remove it in after torch dependency is removed
           pip uninstall -y texar-pytorch
```
hunterhector, line 101: I recalled you mentioned this can be deleted? Or do we need to delete it after removing "torch

.github/workflows/main.yml
```diff
@@ -73,30 +82,25 @@ jobs:
...
-          git clone https://github.com/asyml/texar-pytorch.git
-          cd texar-pytorch
-          pip install --use-feature=in-tree-build --progress-bar off .
-          cd ..
-          # Remove them to avoid confusion.
-          rm -rf texar-pytorch
       - name: Install Forte
         run: |
           pip install --use-feature=in-tree-build --progress-bar off .[models,test,wikipedia,data_aug,nlp,ir,texar-encoder,stave,audio_ext,remote,extractor]
+      - name: Test models module
```
hunterhector, line 88: step name is not correct.

README.md
```diff
@@ -60,19 +60,21 @@ pip install src/spacy
 Some components or modules in forte may require some [extra requirements](https://github.com/asyml/forte/blob/master/setup.py#L45):
 
 
-* `pip install forte[data_aug]`: Install packages required for [data augmentation modules](https://github.com/asyml/forte/tree/master/forte/processors/data_augment).
```
hunterhector, line 63: any reasons that we re-oder these lines? The previous order is based on whether they are actively developed by us.

feipenghe: I wanted to maintain the same order as those modules in `setup.py` so that I could check them in parallel. I can rearrange the order in `setup.py` instead.

.github/workflows/main.yml
```diff
@@ -31,6 +31,14 @@ jobs:
             tensorflow-version: 2.5.0
         notebook-details:
           - { dep: "huggingface nltk", extra: "'tensorflow>=2.5.0,<2.8.0,termcolor>=1.1.0'"}
+        modules:
+          - { module: "models", test_file: "tests/forte/processors/subword_tokenizer_test.py tests/forte/processors/pretrained_encoder_processors_test.py"}
+          - { module: "data_aug", test_file: "tests/forte/processors/subword_tokenizer_test.py tests/forte/processors/pretrained_encoder_processors_test.py"}
```
hunterhector, line 36: The test files here are wrong

.github/workflows/main.yml
```diff
@@ -31,6 +31,14 @@ jobs:
...
+          - { module:  "nlp",test_file: "tests/forte/models"}
```
hunterhector, line 37: should the module be `model`?

.github/workflows/main.yml
```diff
@@ -31,6 +31,14 @@ jobs:
...
+          - { module: "ir",test_file: "tests/forte/processors/ir/"}
+          - { module: "stave",test_file: "tests/forte/data/readers/stave_reader_test.py tests/forte/processors/stave_processor_test.py"}
+          - { module: "audio_ext",test_file: "tests/forte/data/readers/audio_reader_test.py"}
+          - { module: "extractor",test_file: "tests/forte/train_preprocessor_test.py forte/data/extractors"}
```
hunterhector, line 41: I also don't see `remote` here, can you double check on all the test cases to make sure these are what you want?

.github/workflows/main.yml
```diff
@@ -31,6 +31,14 @@ jobs:
             tensorflow-version: 2.5.0
         notebook-details:
           - { dep: "huggingface nltk", extra: "'tensorflow>=2.5.0,<2.8.0,termcolor>=1.1.0'"}
+        modules:
+          - { module: "models", test_file: "tests/forte/processors/subword_tokenizer_test.py tests/forte/processors/pretrained_encoder_processors_test.py"}
```
hunterhector, line 35: should be `nlp`?

.github/workflows/main.yml
feipenghe: Since the test import only test nondependency on texar but it still has dependency on `torch`, `torchtext` etc. 
Therefore, there are two ways for testing non dependency:
1. Remove install/uninstall script and also remove test importings that have dependency on other extra packages such as torch, torchtext and transformers.
2. test non dependency for each package independently just as this script with install/uninstall script.

.github/workflows/main.yml
feipenghe: I have fixed all test files and there is not test file for remote processors.

hunterhector: @mylibrar do we miss a test case for `remote`? Do we need to make an issue to track it?

mylibrar: Test cases for `remote` are included in https://github.com/asyml/forte/blob/master/tests/forte/advanced_pipeline_test.py

.github/workflows/main.yml
hunterhector: what i don't understand is why `texar-pytorch` appears here, shouldn't we  test this after we install Forte only? Before installing any extra dependencies?

feipenghe: The test file `texar_nondependency_test.py` is only testing `texar`'s nondependency but those classes in the test file still have dependencies on other extra required packages. 
- install forte
- install all extra required packages
- uninstall texar and test `texar`'s nondependency and install texar back

feipenghe: If we only keep classes have dependencies on forte in `texar_nondependency_test.py`, then we can have pipeline 
- install forte
- test nondependency for all extra required packages
- install required packages to test packages separately (dependency matrix)
- install all extra required packages and test all files (regular testing)

hunterhector: Which extra requires packages are needed? We should only test the backbone Forte here, so it shouldn't require other extra requires.

feipenghe: It requires `torch`, `transformers` and `torchtext` from the previous output. I can check and remove classes actually use them.
The other way is to build up nondependency one by one as current install/uninstall pipeline and finally have a backbone test.

.github/workflows/main.yml
```diff
@@ -69,34 +80,28 @@ jobs:
...
-          rm -rf texar-pytorch
+          pytest tests/forte/texar_nondependency_test.py
+          pip uninstall -y torch
+      - name: Test modules need extra packages
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .[test]
+          if [ ${{ matrix.modules.module }} != "" ]
+          then
+            pip install --use-feature=in-tree-build --progress-bar off .[${{ matrix.modules.module }}]
+            pip install --progress-bar off pytest==5.1.3 coverage
```
hunterhector, line 96: I think we have mentioned this, we shouldn't do the matrix test together with the main test, this will make the CI time really long because the normal test is also done N times (N is the number of matrix combinations).

tests/forte/pipeline_test.py
```diff
@@ -50,10 +50,10 @@
 from forte.pipeline import Pipeline
 from forte.processors.base import (
     PackProcessor,
-    RequestPackingProcessor,
     MultiPackProcessor,
 )
 from forte.processors.base.batch_processor import (
+    RequestPackingProcessor,
```
hunterhector, line 56: Can this be imported like before?

tests/forte/processors/stave_processor_test.py
```diff
@@ -26,8 +26,11 @@
 from forte.data.data_pack import DataPack
 from forte.pipeline import Pipeline
 from forte.data.readers import OntonotesReader
-from forte.processors.base import PackProcessor, RequestPackingProcessor
-from forte.processors.base.batch_processor import Predictor
+from forte.processors.base import PackProcessor
+from forte.processors.base.batch_processor import (
+    Predictor,
+    RequestPackingProcessor,
```
hunterhector, line 32: Is it possible that we don't change the import here?

tests/forte/processors/timeprofiling_test.py
```diff
@@ -22,8 +22,11 @@
 from forte.data.data_pack import DataPack
 from forte.pipeline import Pipeline
 from forte.data.readers import StringReader
-from forte.processors.base import PackProcessor, RequestPackingProcessor
-from forte.processors.base.batch_processor import Predictor
+from forte.processors.base import PackProcessor
+from forte.processors.base.batch_processor import (
+    Predictor,
+    RequestPackingProcessor,
```
hunterhector, line 28: Same as above.

.github/workflows/main.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           - { module: "wikipedia", test_file: "tests/forte/datasets/wikipedia"}
           - { module:  "nlp",test_file: "tests/forte/processors/subword_tokenizer_test.py tests/forte/processors/pretrained_encoder_processors_test.py"}
           - { module: "extractor",test_file: "tests/forte/train_preprocessor_test.py forte/data/extractors"}
+          - { module: "remote", test_file: "tests/forte/advanced_pipeline_test.py"}
```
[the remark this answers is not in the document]
feipenghe: I think this is outdated `main.yml` I have moved remote processor test out in the the latest one.

tests/forte/pipeline_test.py
feipenghe: Yes. I will revert it.

.github/workflows/main.yml
```diff
@@ -45,7 +56,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
+
       - name: Install dependencies
+        if: matrix.modules.module == 'main'
```
hunterhector, line 61: The `module main` syntax is making the complex quite complex, how about just a add a new job step?

tests/forte/processors/machine_translation_processor_test.py
```diff
@@ -24,7 +24,9 @@
 from forte.data.multi_pack import MultiPack
 from forte.pipeline import Pipeline
 from forte.data.readers import MultiPackSentenceReader
-from forte.processors.third_party import MicrosoftBingTranslator
+from forte.processors.third_party.machine_translation_processor import (
+    MicrosoftBingTranslator,
```
hunterhector, line 28: can this also be reverted?

.github/workflows/main.yml
feipenghe: I am not sure how to add a job step to separate the two tasks. Is there any more instructions?

hunterhector: we have the `build` job: https://github.com/asyml/forte/blob/master/.github/workflows/main.yml#L12 and the `docs` job: https://github.com/asyml/forte/blob/master/.github/workflows/main.yml#L134

Also check here: https://github.com/asyml/forte-wrappers/blob/main/.github/workflows/main.yml, we have `build`, `pytest` and `docs`

hunterhector:
also, don't touch unrelated files (even just format changes), to avoid conflicts

.github/workflows/main.yml
```diff
@@ -131,6 +118,65 @@ jobs:
           fi
 
 
+  test_modules:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.6, 3.7 ]
+        torch-version: [ 1.5.0, 1.6.0 ]
+        tensorflow-version: [ 1.15.0 ]
```
hunterhector, line 127: Are the modules here also tested in the main test? If so, there is no need to have all the python, pytorch versions here.

.github/workflows/main.yml
```diff
@@ -131,6 +118,51 @@ jobs:
...
 
+  test_modules:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.6 ]
+        torch-version: [ 1.5.0 ]
+        tensorflow-version: [ 1.15.0 ]
+        notebook-details:
+          - { dep: "huggingface nltk", extra: "'tensorflow>=2.5.0,<2.8.0,termcolor>=1.1.0'"}
```
hunterhector, line 129: these notebook details, are they needed?

.github/workflows/main.yml
```diff
@@ -131,6 +118,47 @@ jobs:
...
+            ${{ runner.os }}-
+      - name: install forte
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .
+      - name: Test modules need extra packages
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .[test]
+          pip install --use-feature=in-tree-build --progress-bar off .[${{ matrix.modules.module }}]
+          pip install --progress-bar off pytest==5.1.3 coverage
+          # coverage run -m pytest tests ${{ matrix.modules.test_file }}
```
hunterhector, line 160: The test line is commented out. And you also don't need to run `coverage` here.

This command should be `pytest ${{ matrix.modules.test_file }}`

Don't test the `tests` folder here `pytest test` too

feipenghe: OK. I have changed it to `pytest ${{ matrix.modules.test_file }}`. I think I still need to comment it out and remove comment after move packages dependency out.

README.md
```diff
@@ -76,134 +59,111 @@ Some components or modules in forte may require some [extra requirements](https:
...
 * `pip install forte[nlp]`: Install packages required for additional NLP supports, such as [subword_tokenizer](https://github.com/asyml/forte/tree/master/forte/processors/nlp/subword_tokenizer.py) and [texar encoder](https://github.com/asyml/forte/tree/master/forte/processors/third_party/pretrained_encoder_processors.py)
-* `pip install forte[extractor]`: Install packages required for extractor-based training system, [extractor](https://github.com/asyml/forte/blob/master/forte/data/extractors), [train_preprocessor](https://github.com/asyml/forte/tree/master/forte/train_preprocessor.py) and [tagging trainer](https://github.com/asyml/forte/tree/master/examples/tagging/tagging_trainer.py)
+* `pip install forte[extractor]`: Install packages required for extrator-based training system, [extractor](https://github.com/asyml/forte/blob/master/forte/data/extractors), [train_preprocessor](https://github.com/asyml/forte/tree/master/forte/train_preprocessor.py) and [tagging trainer](https://github.com/asyml/forte/tree/master/examples/tagging/tagging_trainer.py)
```
hunterhector, line 62: why are you changing all these lines, even change a word from correct to misspelling?

docs/requirements.txt
```diff
@@ -33,7 +33,7 @@ elasticsearch==7.5.1
 
 # code generation
 typed_astunparse~=2.1.4
-typed_ast>=1.5.0
+typed_ast~=1.4.0
```
hunterhector, line 36: why are we reverting the changes?

.github/workflows/main.yml
hunterhector: why?

examples/classification/bank_customer_intent.py
```diff
@@ -113,9 +113,7 @@
         "label",
     ],
     "index2class": index2class,
-    "text_fields": [
-        "ft.onto.base_ontology.Body"
-    ],
+    "text_fields": ["ft.onto.base_ontology.Body"],
```
hunterhector, line 116: I would say, revert every file that's not relevant to this PR. don't touch them

.github/workflows/main.yml
```diff
@@ -131,6 +118,47 @@ jobs:
...
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: install forte
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .
+      - name: Test modules need extra packages
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .[test]
+          pip install --use-feature=in-tree-build --progress-bar off .[${{ matrix.modules.module }}]
+          pip install --progress-bar off pytest==5.1.3 coverage
```
hunterhector, line 159: now since you don't use `coverage`, you don't need to install coverage. Please really think about it when you are making changes, you have to *understand* what you are doing, not just following instructions.

README.md
feipenghe: Sorry I checked changed file sections and it shows partial format changes which I thought was unnecessary. I will git reset it,

examples/classification/bank_customer_intent.py
feipenghe: I tried to revert it but I think it's because `black` is updated so it won't pass CI if I push the original commit.

README.md
hunterhector: just don't change anything that's relevant, event format changes in the future.

examples/classification/bank_customer_intent.py
hunterhector: the thing is if the master passes it should pass. you should keep black to be the same version from local to your side so that you don't need to change anything unnecessary.

.github/workflows/main.yml
```diff
@@ -131,6 +118,63 @@ jobs:
...
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: install forte
+        run: |
+          pip install --use-feature=in-tree-build --progress-bar off .
+      - name: install dependencies
+        run: |
+          sudo apt-get install -y libsndfile1-dev
+          python -m pip install --progress-bar off --upgrade pip
+          pip install --progress-bar off Django django-guardian
```
hunterhector, line 159: so here we expose a couple more dependencies missing, let's address them in the next couple PRs.

hunterhector (approved):

Issue #741, opened 2022-04-14
Move out dependency of PyTorch to extra_requires in forte


feipenghe:
**Describe the bug**
We want to separate out the PyTorch dependency to extra_requires in forte. 

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

PR #742, opened 2022-04-14
Move PyTorch to extra require

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/741 

### Description of changes
* raise ImportError with error message pointing to README.md tutorial when pytorch is not installed properly
* add new extra require for modules requiring pytorch

* User doesn't need to install pytorch by default for using main forte functions

### Test Conducted
* Implemented `torch_non_dependency_test.py` that imports classes that doesn't have dependency on `torch` but might trigger importing `torch` accidentally through `__init__.py` or importing `torch` at the head of class files.


feipenghe:
created a clean PR https://github.com/asyml/forte/pull/750 to solve the issue

PR #750, opened 2022-04-21
Move out PyTorch to extra require

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/741.

### Description of changes
* raise ImportError with error message pointing to README.md tutorial when pytorch is not installed properly
* add new extra require for modules requiring pytorch
* fixed paths for classes's paths that have been changed such as `SRLPredictor` and `CoNLLNERPredictor`
* add `forte[data]` module in `setup.py`, workflow testing, and `README.md`.

* User doesn't need to install pytorch by default for using main forte functions

### Test Conducted
* Implemented `tests/forte/forte_backbone_test.py` that imports classes that doesn't have dependency on `torch` but might trigger importing `torch` accidentally through `__init__.py` or importing `torch` at the head of class files.


.github/workflows/main.yml
```diff
@@ -135,6 +135,7 @@ jobs:
           - { module: "wikipedia", test_file: "tests/forte/datasets/wikipedia"}
           - { module: "nlp",test_file: "tests/forte/processors/subword_tokenizer_test.py tests/forte/processors/pretrained_encoder_processors_test.py"}
           - { module: "extractor",test_file: "tests/forte/train_preprocessor_test.py forte/data/extractors"}
+          - { module: "data", test_file: "tests/forte/data/data_pack_dataset_test.py tests/forte/data/converter/converter_test.py"}
```
hunterhector, line 138: these actually belong to the `extractor` project, can be merged together

README.md
```diff
@@ -72,12 +72,12 @@ Some components or modules in forte may require some [extra requirements](https:
...
 * `pip install forte[wikipedia]`: Install packages required for reading [wikipedia datasets](https://github.com/asyml/forte/tree/master/forte/datasets/wikipedia).
 * `pip install forte[nlp]`: Install packages required for additional NLP supports, such as [subword_tokenizer](https://github.com/asyml/forte/tree/master/forte/processors/nlp/subword_tokenizer.py) and [texar encoder](https://github.com/asyml/forte/tree/master/forte/processors/third_party/pretrained_encoder_processors.py)
 * `pip install forte[extractor]`: Install packages required for extractor-based training system, [extractor](https://github.com/asyml/forte/blob/master/forte/data/extractors), [train_preprocessor](https://github.com/asyml/forte/tree/master/forte/train_preprocessor.py) and [tagging trainer](https://github.com/asyml/forte/tree/master/examples/tagging/tagging_trainer.py)
-
+* `pip install forte[data]`: Install packages required for [data pack dataset](https://github.com/asyml/forte/blob/master/forte/data/data_pack_dataset.py), [types](https://github.com/asyml/forte/blob/master/forte/data/types.py), and [converter](https://github.com/asyml/forte/blob/master/forte/data/converter).
```
hunterhector, line 80: remove this when the previous comment is fixed

forte/data/converter/converter.py
```diff
@@ -277,7 +277,14 @@ def convert(
         # Convert to target type
         if not self.to_numpy and not self.to_torch:
             return data_list, masks_list
-
+        try:
```
hunterhector, line 280: we should put these try blocks at the beginning of the function. If this fails, there is no need to execute the rest.

forte/models/da_rl/magic_model.py
```diff
@@ -21,8 +21,16 @@
...
-import torch
-from torch import nn
+
+try:
+    import torch
+    from torch import nn
+except ImportError as e:
+    raise ImportError(
+        " `pytorch` is not installed correctly."
+        " Please refer to [extra requirement for models](pip install forte[models])"
```
hunterhector, line 31: For all these exception messages, it would be best to:
1. Mention that PyTorch needed to be installed, 
2. Also mention the extra requirements.

So sometimes the user can simply follow the error message to install pytorch, and don't need to go to read the documentation

forte/models/ner/model_factory.py
```diff
@@ -151,14 +160,14 @@ def decode(self, input_word, input_char, mask=None, hx=None):
...
         try:
             import texar.torch as texar  # pylint: disable=import-outside-toplevel
-        except ImportError as e:
+        except ImportError as e3:
             raise ImportError(
                 " `texar-pytorch` is not installed correctly."
                 " Consider install texar via `pip install texar-pytorch`."
                 " Or refer to [extra requirement for Texar model support]"
                 "(pip install forte[models])"
                 " for more information."
```
hunterhector, line 169: This message is better than the previous one, that's why we should actually create a function to centrally manage the messages, instead of copy/pasting them.

setup.py
```diff
@@ -70,6 +70,7 @@
         # transformers 4.10.0 will break the translation model we used here
         "nlp": ["texar-pytorch>=0.1.4"],
         "extractor": ["texar-pytorch>=0.1.4"],
+        "data": ["texar-pytorch>=0.1.4", "torch>=1.1.0"],
```
hunterhector, line 73: as explained, these are part of extractor

forte/data/converter/converter.py
hunterhector: you probably want to do something like

```
if self.to_torch
   try import
```

at the beginning of the function

setup.py
```diff
@@ -69,7 +69,7 @@
         "wikipedia": ["rdflib==4.2.2"],
         # transformers 4.10.0 will break the translation model we used here
         "nlp": ["texar-pytorch>=0.1.4"],
-        "extractor": ["texar-pytorch>=0.1.4"],
+        "extractor": ["texar-pytorch>=0.1.4", "torch>=1.1.0"],
```
hunterhector, line 72: So:
1. texar-pytorch requires `pytorch`, so we don't have to explicitly say it here
2. to install torch, it is better to ask users to install them by following pytorch instructions, they can choose whether to install the CPU, GPU version, and other things. basically, pytorch installation can be tricky and a single `pip install torch` may not work

tests/forte/forte_backbone_test.py
```diff
@@ -56,3 +56,10 @@ def test_import_models(self):
```
hunterhector, line 56: Some of these import tests are not necessary. For example, if we want to use the `SRL` part we should install texar-pytorch. If we don't use that part, classes like `SRLSpan` won't be used.

tests/forte/forte_backbone_test.py
```diff
@@ -56,3 +56,10 @@ def test_import_models(self):
 
     def test_import_trainer(self):
         from forte.trainer.base import BaseTrainer
+
+    def test_import_forte_modules(self):
```
hunterhector, line 60: I think you also want to import them in their short form, for example, should be `from forte import pipeline`

tests/forte/forte_backbone_test.py
feipenghe: Even though some classes under `SRL` are not using `texar-pytorch` directly but the whole module dependent on packages required by `forte[models]`. Therefore, we don't need to test any of them as they are not used individually. Is that what you mean?

tests/forte/forte_backbone_test.py
feipenghe: corrected.

tests/forte/forte_backbone_test.py
hunterhector: yes

feipenghe: ok I have removed them.

hunterhector (approved):

Issue #747, opened 2022-04-19
Try-Except for Importing Packages Needs Modularization


feipenghe:
**Describe the bug**
Current try-except for importing packages is written independently. An example is as the following:
```
try:
    from texar.torch.data import (
        Batch,
    )  # pylint: disable=import-outside-toplevel
except ImportError as e:
    raise ImportError(
        " `texar-pytorch` is not installed correctly."
        " Consider install texar via `pip install texar-pytorch`."
        " Or refer to [extra requirement for extractor system](pip install forte[extractor])"
        " for more information. "
    ) from e
```
Too many of the imports will result in repetitive code, and we can actually modularize the code by designing a function that takes `importing_modules` and `installing_module` as parameters. 

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

PR #751, opened 2022-04-21
Modularize try-except  ImportError

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/747 

### Description of changes
* add a function that try-catch import error for a list of modules
* change all current hard-coded try-catch import errors to modularized ones.

* Future developers can use this method to report import error for third party packages in scripts.

### Test Conducted
* test import modules
* test raise errors when the module is not installed


forte/utils/utils.py
```diff
@@ -303,3 +304,81 @@ def build_alignment(self, text: str, text_to_align: str):
...
+        parent_path, module_name, alias = module_info
+        if parent_path is None:
+            import_script = f"import {module_name}"
+        else:
+            import_script = f"from {parent_path} import {module_name} "
+        if alias is not None:
+            import_script += f" as {alias}"
+        if not top_level:
+            import_script += "# pylint: disable=import-outside-toplevel"
+        exec(import_script)
```
hunterhector, line 371: I don't feel very confident about doing `exec`. Can we simply use importlib https://docs.python.org/3/library/importlib.html#importlib.import_module?

Are you trying to create the code here? I would rather not do this. And even if we need to do this, I don't think we need to include the `pylint` escpae here, I suppose `pylint` won't check the string you concatenated.

hunterhector (commented):
The solution I suggested is actually much simpler, we just need to modularize the "error message", but don't need to do this.  You are working on a solution that tries to mimic the whole import behavior.

If this works then it might be fine, but I don't find it safe/reliable to use `exec` and manual code concat, can we simply use `importlib`?

If this solution has to use `exec` and then the issue is quite complicated, I would suggest still trying to modularize the error message only.

forte/utils/utils.py
```diff
@@ -303,3 +304,45 @@ def build_alignment(self, text: str, text_to_align: str):
...
+    Args:
+        extra_module: module name should be installed by pip.
+        forte_module: forte module User should install by
+            ``pip install forte[`forte_module`]`` to install all
+            extra packages for using the forte module.
+        component_name: the forte component that needs the module.
+
+    """
+    forte_module_link_d = {
+        "data_aug": "https://github.com/asyml/forte/blame/master/README.md#L70",
```
hunterhector, line 328: These links might be updated from time to time, like line number changes. Also, you should not use the `blame` version, even if this points to the correct line. 

A couple of solutions would be:
1. Consider adding link/anchor to that line
2. Or simply use a tagged version or a particular commit

You can point to the general area, without pointing to the specific line if that's not possible.

forte/processors/data_augment/algorithms/dictionary.py
```diff
@@ -95,8 +95,9 @@ def __init__(self):
             )
         except ImportError as err:
             raise ImportError(
-                "nltk is not installed correctly."
-                "Please refer to documentation at [link](pip install forte[data_aug])"
+                create_import_error_msg(
+                    "nltk", "data_aug", "Data Augmentation(UDA)"
```
hunterhector, line 99: This is a dictionary-based data augmentation, now UDA

forte/utils/utils.py
```diff
@@ -303,3 +304,45 @@ def build_alignment(self, text: str, text_to_align: str):
...
+        "test": "https://github.com/asyml/forte/blame/master/README.md#L76",
+        "wikipedia": "https://github.com/asyml/forte/blame/master/README.md#L77",
+        "nlp": "https://github.com/asyml/forte/blame/master/README.md#L78",
+        "extractor": "https://github.com/asyml/forte/blame/master/README.md#L79",
+    }
+
+    error_msg = (
+        f" `{extra_module}` is not installed correctly."
+        + f" Consider install {extra_module}"
+        + f"via `pip install {extra_module}`."
```
hunterhector, line 343: Sometimes the module may not be installed via `pip`, just tell the user to install this python module.

forte/utils/utils.py
feipenghe: To have anchors for these links, there should be section for each forte module like the screenshot below. Do you think it's acceptable? If so I will create a separate PR for it.
image

hunterhector: i don't think we have to point to the exact line here, just point to this section for now until we know a better solution.

hunterhector (approved):

Issue #759, opened 2022-04-25
Move out dependency of rest modules in extra_requires in forte


feipenghe:
**Describe the bug**

We have moved out dependencies of 
PyTorch https://github.com/asyml/forte/issues/741 
Texar https://github.com/asyml/forte/issues/719
and we also need to make sure the rest extra modules are not depended by main forte project.

* transformers
* nltk
* tesnorflow
*  fastapi
* uvicorn
* soundfile
* stave
* rblib

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

PR #760, opened 2022-04-25
Move out dependency of rest modules in extra_requires in forte

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/759

### Description of changes
* add more import tests that don't depend on the extra module but might be imported unexpectedly
* add try-except import error message the extra modules

After this PR, forte main module no longer depend on extra modules and can report error message correctly when extra modules is needed and not installed.

### Test Conducted
* add more import tests that don't depend on the extra module but might be imported unexpectedly

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,15 +112,14 @@ jobs:
           fi
   test_backbone:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [ 3.6 ]
+    env:
+      python-version: 3.7
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ env.python-version }}
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ env.python-version }}
       - name: Cache pip
         uses: actions/cache@v2
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -131,14 +131,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
-      - name: Install Pytest
-        run: |
-          pip install --progress-bar off pytest==5.1.3
       - name: Test backbone Forte import test
         run: |
           # Try to install Forte backbone only and test basic imports.
           pip install --use-feature=in-tree-build --progress-bar off .
-          pytest tests/forte/forte_backbone_test.py
+          python tests/forte/forte_backbone_test.py
 
   test_modules:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,8 +115,6 @@ jobs:
     strategy:
       matrix:
         python-version: [ 3.6 ]
-        torch-version: [ 1.5.0 ]
-        tensorflow-version: [ 1.15.0 ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -131,6 +131,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
+      - name: Install Pytest
+        run: |
+          pip install --progress-bar off pytest==5.1.3
       - name: Test backbone Forte import test
         run: |
           # Try to install Forte backbone only and test basic imports.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,11 +72,9 @@ jobs:
           rm -rf stave
       - name: Test backbone Forte import test
         run: |
-          pip install --progress-bar off torch==${{ matrix.torch-version }}
           # Try to install Forte backbone only and test basic imports.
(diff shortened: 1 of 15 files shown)


tests/forte/forte_backbone_test.py
```diff
@@ -55,3 +90,59 @@ def test_import_forte_modules(self):
         from forte.process_job import ProcessJob, ProcessJobStatus
         from forte.process_manager import ProcessManager
         from forte.train_pipeline import TrainPipeline
+
+    def test_import_base_data_aug(self):
```
hunterhector, line 94: will data aug require some additional dependencies?

feipenghe: Only few modules under `forte/processors/data_augment` require the extra dependencies. For example,
`forte/processors/data_augment/algorithms/dictionary.py` requires `nltk`
`forte/processors/data_augment/algorithms/machine_translator.py` requires `transformers`

feipenghe: I was also thinking maybe split modules under `forte/processors/data_augment` or other folders to two subfolders one requires extra dependencies and the other doesn't. I am not sure if it's good idea tho.

mylibrar (commented):
I cannot pass the `forte_backbone_test.py` on my local machine:
```
E   ModuleNotFoundError: No module named 'requests'
```
It still can pass the CI because `requests` is required by [coverage codecov](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L56). I guess we might want a cleaner environment for [`Test backbone Forte import test`](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L73) in `main.yml` workflow.  

For example, I guess we should remove [`pip instal torch`](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L75) before running the `forte_backbone_test` in workflow since `pytorch` is already moved out.

tests/forte/forte_backbone_test.py
hunterhector: But it seems that most of the data-aug algorithms in the `forte/processors/data_augment` folder are not in here: https://github.com/asyml/forte/blob/master/forte/processors/data_augment/algorithms/__init__.py. So I guess they won't be automatically imported anyway right?

hunterhector:
> I cannot pass the `forte_backbone_test.py` on my local machine:
> 
> ```
> E   ModuleNotFoundError: No module named 'requests'
> ```
> 
> It still can pass the CI because `requests` is required by [coverage codecov](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L56). I guess we might want a cleaner environment for [`Test backbone Forte import test`](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L73) in `main.yml` workflow.
> 
> For example, I guess we should remove [`pip instal torch`](https://github.com/feipenghe/forte/blob/27fa1bfb30c4edc2208d11cf0ccf5ae82ef0daea/.github/workflows/main.yml#L75) before running the `forte_backbone_test` in workflow since `pytorch` is already moved out.

I think the suggestion makes sense. In fact, we should probably put the backbone test into its own job, where we only do:

```
          pip install --use-feature=in-tree-build --progress-bar off .
          pytest tests/forte/forte_backbone_test.py
```

We don't want anything else to affect this, including our workflow packages like `codecov`, `mypy`, `black` etc. Furthermore, we should remove the hacks we used in the workflows. For example, we shouldn't need a line that uninstall pytorch https://github.com/asyml/forte/blob/master/.github/workflows/main.yml#L79

tests/forte/forte_backbone_test.py
feipenghe: That's true. If we only need test from this perspective then `def test_import_base_data_aug(self):` is not needed.

forte/processors/misc/__init__.py
```diff
@@ -18,4 +18,3 @@
 from forte.processors.misc.vocabulary_processor import *
 from forte.processors.misc.delete_overlap_entries import *
 from forte.processors.misc.simple_processors import *
-from forte.processors.misc.remote_processor import *
```
mylibrar, line 21: This might break some code in `stave` (e.g., the [remote handler](https://github.com/mylibrar/stave/blob/master/simple-backend/stave_backend/handlers/nlp.py#L20) used by eliza project). I guess we can put the try-import block of `requests` under `RemoteProcessor.__init__ ` and revert update here and in `remote_processor_test.py`.

.github/workflows/main.yml
```diff
@@ -116,7 +110,32 @@ jobs:
...
             coverage run -m pytest tests/forte/notebooks
           fi
-
+  test_backbone:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.6 ]
+        torch-version: [ 1.5.0 ]
+        tensorflow-version: [ 1.15.0 ]
```
mylibrar, line 119: Maybe we don't need to put `pytorch` and `tensorflow` here in the matrix since they are not required in this job.

mylibrar (approved):

PR #763, opened 2022-04-26
Increment minor version

mylibrar:
This PR bumps `forte`'s version to `0.2.0`

### DIFF
--- a/forte/version.py
+++ b/forte/version.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 _MAJOR = "0"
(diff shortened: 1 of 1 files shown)


hunterhector (commented):
I am checking how setup.py reads this. I find that we should do a small check around here:

https://github.com/asyml/forte/blob/master/setup.py#L16

We need to make sure the version variable should be checked, and throw run time exception if it is not found or not well formated.

hunterhector (approved):

Issue #694, opened 2022-03-23
Implement get_data_with_range in data_store.py


wanglec:
**Is your feature request related to a problem? Please describe.**
This function is used to get data from a data store within a range. It takes in `type_name` and `range_annotation`. The `type_name` is mandatory, where all the entries matching this type will be returned. The `range_annotation` controls the search area of the sub-types. It should support annotation-like types.

**Depends on**
- List issues that this one depends on.

**Describe the solution you'd like**
We first locate the begin and end index of `range_annotation` and bisect the corresponding sortedlist to filter the range. Then, we generate entries within the sortedlist.

**Describe alternatives you've considered**

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

Issue #744, opened 2022-04-18
Fix `get` function in data_store.py


wanglec:
**Is your feature request related to a problem? Please describe.**
The current implementation of the `get` function has three problems. 

- [ ] First, it does not generate annotation-like entries with correct begin/end order, will address in https://github.com/asyml/forte/pull/769
- [ ] Second, it does not have a defined order for non-annotation-like types. 
- [ ] Third, it does not cache the find subtype function.

**Depends on**
- List issues that this one depends on.

**Describe the solution you'd like**
We may use the `co_iterator_annotation_like` function to solve the first problem. 
To solve the second problem, we may sort all subtypes and subsequently generate entries belonging to each subtype.

**Describe alternatives you've considered**

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

PR #769, opened 2022-05-03
Let get function take range annotation

wanglec:
This PR fixes https://github.com/asyml/forte/issues/694 and https://github.com/asyml/forte/issues/744.

### Description of changes
Change the `get` function so that it accepts range annotation, use co-iterator for annotation like, and fix order for non-annotation like subtypes. A helper function that finds the length of entry lists, `get_length(self, type_name: str)`, is also included.

### Test Conducted
Tests are added in data_store_test.py.


forte/data/data_store.py
```diff
@@ -892,27 +911,93 @@ def co_iterator_annotation_like(
...
+                            for m in members:
+                                e = self.__tid_ref_dict[m]
+                                if not within_range(e, range_annotation):
+                                    within = False
+                                    break
+                            if within:
+                                yield entry
         else:
             if type_name not in self.__elements:
                 raise KeyError(f"type {type_name} does not exist")
```
hunterhector, line 1003: 1. we should use check this at the very beginning without going through all the other steps
   - we probably should use `ValueError` since this is an incorrect argument provided from the function signature.
2. what happens if this item is not `Group`, `List`, or `Annotation`?  Maybe we should say we cannot handle.

forte/data/data_store.py
```diff
@@ -759,6 +758,26 @@ def get_entry_index(self, tid: int) -> int:
...
+        Returns:
+            The count of not None entries.
+        """
+        if self._is_annotation(type_name):
+            return len(self.__elements[type_name])
+        else:
+            try:
+                delete_count = self.__deletion_count[type_name]
+            except KeyError:
+                delete_count = 0
```
hunterhector, line 778: you can simply do `delete_count = self.__deletion_count.get(type_name, 0)`, and this whole branch is simply `return len(self.__elements[type_name] - self.__deletion_count.get(type_name, 0)`

forte/data/data_store.py
```diff
@@ -892,27 +911,93 @@ def co_iterator_annotation_like(
             yield entry
 
     def get(
-        self, type_name: str, include_sub_type: bool = True
+        self,
+        type_name: str,
+        include_sub_type: bool = True,
+        range_annotation: Optional[List[int]] = None,
```
hunterhector, line 917: why is this a `List` or not simply a `Tuple[int, int]`?

forte/data/data_store.py
wanglec: yes

forte/data/data_store.py
wanglec: yes

forte/data/data_store.py
wanglec: Yes, I think I can use a tuple instead.

forte/data/data_store.py
```diff
@@ -982,30 +998,101 @@ def co_iterator_annotation_like(
...
-            all_types = []
-            # iterate all classes to find subclasses
             for type in self.__elements:
                 if issubclass(get_class(type), entry_class):
-                    all_types.append(type)
+                    all_types.add(type)
+        else:
+            all_types.add(type_name)
+        all_types = list(all_types)
+        all_types.sort()
```
hunterhector, line 1049: @mylibrar take a note that we will replace this with get subtype function later, including the sorting.

forte/data/data_store.py
```diff
@@ -982,30 +998,101 @@ def co_iterator_annotation_like(
...
+        if self._is_annotation(type_name):
+            if range_annotation is None:
+                yield from self.co_iterator_annotation_like(all_types)
+            else:
+                for entry in self.co_iterator_annotation_like(all_types):
+                    if within_range(entry, range_annotation):
+                        yield entry
```
hunterhector, line 1056: This is very inefficient, let's make an issue @mylibrar, we should only iter in the range from the start

forte/data/data_store.py
```diff
@@ -982,30 +998,101 @@ def co_iterator_annotation_like(
...
+                                parent, range_annotation
+                            ) and within_range(child, range_annotation):
+                                yield entry
+        elif issubclass(entry_class, Group):
+            for type in all_types:
+                if range_annotation is None:
+                    yield from self.iter(type)
+                else:
+                    for entry in self.__elements[type]:
+                        member_type = entry[constants.BEGIN_INDEX]
```
hunterhector, line 1084: same as above, should we use new consts?

forte/data/data_store.py
```diff
@@ -982,30 +998,101 @@ def co_iterator_annotation_like(
...
+                        yield entry
+        elif issubclass(entry_class, Link):
             for type in all_types:
-                yield from self.iter(type)
+                if range_annotation is None:
+                    yield from self.iter(type)
+                else:
+                    for entry in self.__elements[type]:
+                        if (
+                            entry[constants.BEGIN_INDEX] in self.__tid_ref_dict
```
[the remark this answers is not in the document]
wanglec: fix constant names

forte/data/data_store.py
wanglec: yes fixed

forte/data/data_store.py
hunterhector: Issue created here: https://github.com/asyml/forte/issues/806

forte/data/data_store.py
hunterhector: Issue created here https://github.com/asyml/forte/issues/807

hunterhector (approved):

PR #770, opened 2022-05-04
add get_subtypes and get_class in DataStore; cache is_subtype.

### LINKED ISSUE #765: Implement a new function get_subtype in DataStore
Get_subtype gets a class type and returns all the subtypes of the class. We will add cache to

**Depends on**

For a given class, the function will traverse all valid types in _type_attribute, and add all class that is_subclass() return true to the returned subtype list.
Get subtype should have an inner cache structure to store previous query and results.

**Additional context**

qinzzz:
This PR fixes #765. 

### Description of changes
Add function `get_subtypes` in DataStore. Implemented a wrapper for `get_class` for DataStore class. 
Since `get_subtypes` return value is prone to change, I add cache to `is_subtype` instead.

### Test Conducted

### DIFF
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -30,6 +30,7 @@ class DataStore(BaseStore):
     # TODO: temporarily disable this for development purposes.
     # pylint: disable=pointless-string-statement
     _type_attributes: dict = {}
+    _get_subtype_cache: dict = {}
     enable_cache = True
 
     def __init__(
@@ -259,6 +260,7 @@ class DataStore(BaseStore):
             "parent_class": set(),
         }
         DataStore._type_attributes[type_name] = new_entry_info
+        DataStore.enable_cache = False
 
         return new_entry_info
 
@@ -967,11 +969,21 @@ class DataStore(BaseStore):
         Returns:
             A list of subtypes.
         """
+        if DataStore.enable_cache:
+            try:
+                return DataStore._get_subtype_cache[type_name]
+            except:
+                pass
+
         cls = DataStore.get_class(type_name)
         subtypes = []
         for subtype in DataStore._type_attributes:
             if DataStore._is_subclass(subtype, cls):
                 subtypes.append(subtype)
+        subtypes.sort()
+        DataStore.enable_cache = True
+        DataStore._get_subtype_cache[type_name] = subtypes
+
         return subtypes
 
     @staticmethod
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -30,6 +30,7 @@ class DataStore(BaseStore):
     # TODO: temporarily disable this for development purposes.
     # pylint: disable=pointless-string-statement
     _type_attributes: dict = {}
+    enable_cache = True
 
     def __init__(
         self, onto_file_path: Optional[str] = None, dynamically_add_type=True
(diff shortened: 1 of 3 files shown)


forte/data/data_store.py
```diff
@@ -258,6 +260,7 @@ def _get_type_info(self, type_name: str) -> Dict[str, Any]:
             "parent_class": set(),
         }
         DataStore._type_attributes[type_name] = new_entry_info
+        DataStore.enable_cache = False
```
hunterhector, line 263: should be private/protected

forte/data/data_store.py
```diff
@@ -846,11 +850,11 @@ def get(
             An iterator of the entries matching the provided arguments.
         """
         if include_sub_type:
-            entry_class = get_class(type_name)
+            entry_class = DataStore.get_class(type_name)
```
hunterhector, line 853: I think we shouldn't get the actuall class here.

forte/data/data_store.py
```diff
@@ -939,6 +943,49 @@ def _parse_onto_file(self):
             return
         raise NotImplementedError
 
+    @staticmethod
+    def get_class(type_name):
```
hunterhector, line 947: probably not needed if the usage is removed.

forte/data/data_store.py
```diff
@@ -939,6 +943,49 @@ def _parse_onto_file(self):
...
+        Raises:
+            ValueError: If class is not found based on :attr:`class_name` in DataStore.
+        """
+        if type_name not in DataStore._type_attributes:
+            raise ValueError(f"Class not found in DataStore: {type_name}")
+        return get_class(type_name)
+
+    @staticmethod
+    def get_subtypes(type_name):
+        r"""Returns a complete list of subclasses based on class name.
```
hunterhector, line 965: Actually not a complete list, only the subclasses that are "known" to us. You should explain here which types are "known".

forte/data/data_store.py
```diff
@@ -939,6 +943,49 @@ def _parse_onto_file(self):
...
+
+        Args:
+            type_name (str): Fully qualified name of the class.
+        Returns:
+            A list of subtypes.
+        """
+        if DataStore.enable_cache:
+            try:
+                return DataStore._get_subtype_cache[type_name]
```
hunterhector, line 974: So I can see the following scenario to be a problem: 

- We call the subtype of a particular class A.
- We then add a new entry with class A_1, which is a subclass of A, but not known before
- We call this function because the cache is enabled/valid, this will return the old (wrong set).

The solution would be:
1. We clear the subtype_cache when adding unknown class A
2. When adding the new class A, we update subtype_cache

Also, make sure you add these to the test cases.

Note that only disabling `enable_cahce` is not enough, because if the data (subtype_cache) is there, and something else triggers the cache, this will still cause a problem.

forte/data/data_store.py
```diff
@@ -939,6 +943,49 @@ def _parse_onto_file(self):
...
+            except KeyError:
+                pass
+
+        cls = DataStore.get_class(type_name)
+        subtypes = []
+        for subtype in DataStore._type_attributes:
+            if DataStore._is_subclass(subtype, cls):
+                subtypes.append(subtype)
+        subtypes.sort()
```
hunterhector, line 983: A quick suggestion here would be using a sorted list

forte/data/data_store.py
```diff
@@ -939,6 +943,49 @@ def _parse_onto_file(self):
...
+        DataStore.enable_cache = True
```
hunterhector, line 984: I don't understand this `enable_cache` switch here, what can it do? It is like this switch is always on after you first call this function.

PR #771, opened 2022-05-04
Datastore entry method

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/767. 

### Description of changes
* implemented method `all_entries(entry_type_name)`
* implemented method `num_entries(entry_type_name)`

### Test Conducted
* test `all_entries(entry_type_name)` returns an iterator of raw entry data in the `List` format of `entry_type_name`
* test `num_entries(entry_type_name)` returns a correct number of entries of `entry_type_name`


mylibrar (commented):
This might not preserve the previous behavior in `DataPack` since it does not retrieve the subclasses. For example, `DataPack.all_annotations` is expected to return an iterator of all entries that satisfy `issubclass(Annotation)` (e.g., `Token`, `Document`, `Sentence`, etc.). Hence `DataStore.all_entries("forte.data.ontology.top.Annotation")` should also be able to get all subtypes of `Annotation` in its outputs.

tests/forte/data/data_store_test.py
```diff
@@ -296,21 +305,58 @@ def test_get_type_info(self):
...
+            entry
+            for entry in list(self.data_store._DataStore__entry_dict.values())
+            if entry[3] == "ft.onto.base_ontology.Sentence"
+        ]
+        doc_list = [
+            entry
+            for entry in list(self.data_store._DataStore__entry_dict.values())
+            if entry[3] == "ft.onto.base_ontology.Document"
+        ]
+        ann_list = list(self.data_store._DataStore__entry_dict.values())
```
mylibrar, line 343: This might not be correct if there are `Link`s or `Group`s in `DataStore`. For example, the format of  test entry with `tid=10123` looks like a `Group` but will also be included in `ann_list`

feipenghe: Is there any way to check `Link`, I found `Link` type entry in `self.__elements` is similar to a regular entry.

mylibrar: According to the docstring in `DataStore.__init__`, a `Link` entry has the following format: `[<parent_tid>, <child_tid>, <tid>, <type_name>, <attr_1>, <attr_2>, ..., <attr_n>, index_id]`. So I guess you can still use `<type_name>` to check if it's a link.

forte/data/data_store.py
```diff
@@ -485,6 +485,48 @@ def _is_annotation(self, type_name: str) -> bool:
...
+            entry_type_name (str): the type name of entries that the User wants to retrieve.
+
+        Yields:
+            Iterator of raw entry data in list format.
+        """
+        for entry_type_key in self.__elements:
+            if entry_type_key == entry_type_name or self._is_subclass(
+                entry_type_key,
+                get_class(entry_type_name),
+            ):
```
hunterhector, line 503: Can we create a `get_all_subclass` function here?

1. It will be used in multiple places
2. We can create a more efficient implementation later (you can leave the implementation now)
3. I think we need to sort the entry keys, otherwise the iteration order is kinda random

forte/data/data_store.py
```diff
@@ -485,6 +485,48 @@ def _is_annotation(self, type_name: str) -> bool:
...
+                count += len(self.__elements[entry_type_key])
+        # if non-annotation-like entries_type_name
+        # we need to minus the corresponding delete count
+        if entry_type_name in self.__deletion_count:
+            count -= self.__deletion_count[entry_type_name]
```
hunterhector, line 527: Should you be checking the `deletion_count` inside the for loop?

tests/forte/data/data_store_test.py
```diff
@@ -307,6 +307,46 @@ def test_get_type_info(self):
...
+        self.data_store.delete_entry(9999)
+        num_sent_entries = self.data_store.num_entries(sent_type)
+
+        self.assertEqual(num_sent_entries, len(sent_list) - 1)
+
+        # remove a group
+        self.data_store.delete_entry(23456)
+        num_group_entries = self.data_store.num_entries(group_type)
+        self.assertEqual(num_group_entries, len(group_list) - 1)
+
```
hunterhector, line 349: add a few more tests:
1. delete more than one entry
2. delete an entry of a child type and count the number of the parent type
3. try delete and then add, and then count the numbers

tests/forte/data/data_store_test.py
```diff
@@ -596,6 +636,26 @@ def test_add_group_raw(self):
             ["test_group", [], tid, "forte.data.ontology.top.Group"],
         )
 
+    # def test_add_link_raw(self):
```
hunterhector, line 639: if we remove these test cases, do we have test cases for `add`?

feipenghe: They are added from other PRs.

tests/forte/data/data_store_test.py
hunterhector: I think the rest of the PR is fine. What's the current status of this comment?

@mylibrar  feel free to resolve the comments if you are fine with them, and you can approve it after your review too.

hunterhector (commented):
Looks good to me, but will leave it up to @mylibrar to resolve all comments and approve it.

mylibrar (approved):

PR #785, opened 2022-05-10
Add cache to _is_annotation

qinzzz:
This PR fixes 743. 

### Description of changes
Add a cache structure in DataStore to cache the result of _is_annotation.

### Test Conducted

### DIFF
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -136,6 +136,12 @@ class DataStore(BaseStore):
         self._dynamically_add_type = dynamically_add_type
 
         """
+        ``_is_annotation_cache`` is a dictionary for cacheing the result of function
+        ``_is_annotation``.
+        """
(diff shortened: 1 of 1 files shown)


forte/data/data_store.py
```diff
@@ -422,8 +428,13 @@ def _is_annotation(self, type_name: str) -> bool:
             type or not.
         """
         # TODO: use is_subclass() in DataStore to replace this
-        entry_class = get_class(type_name)
-        return issubclass(entry_class, (Annotation, AudioAnnotation))
+        try:
+            return self._is_annotation_cache[type_name]
```
hunterhector, line 432: so according to our discussion, we don't need this cache, we can rely on the internal issubclass

hunterhector:
probably close this one since we can use other caching

PR #793, opened 2022-05-20
Incorporate remote clinical pipeline and stave

### LINKED ISSUE #745: Improve the `clinical_pipeline` example
The current [clinical_pipeline](https://github.com/asyml/forte/tree/master/examples/clinical_pipeline) example needs to be fixed according to the latest [updates](https://github.com/asyml/stave/commit/c09ab5d987109dd1c45459da08a2582fcea85877#diff-d9fe757a2162ba6785dd721f7fff50d7cba0d4ae769cd34782406fe6ec59922b) in `stave` of how to handle remote pipeline service. `stave` will no longer assemble an NLP pipeline itself. Instead, `stave` now uses a [pipeline](https://github.com/asyml/stave/blob/master/simple-backend/stave_backend/handlers/nlp.py#L49) with `RemoteProcessor` to call remote service. 

Some of the steps in [README.md](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/README.md) might not work anymore. Hence we need to improve the example to make it solid:
* [Install extra dependencies](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/README.md#install-extra-dependencies): The command line to install forte-wrappers[elastic,nltk] might no longer work and we need to update it based on [forte-wrapper repo](https://github.com/asyml/forte-wrappers).
* [Prepare elastic searcher](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/README.md#prepare-elastic-searcher): We may want to point out a specific version for user to download and install since the latest elasticsearch(8.1.2) is not compatible with this example.
* [Run Indexer](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/README.md#run-indexer): Here we should provide a download link to `mimiciii/1.4/NOTEEVENTS.csv.gz`. For example, https://physionet.org/content/mimiciii-demo/1.4/D_ICD_DIAGNOSES.csv.
* [Stave](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/README.md#stave):
  * We want to provide a remote pipeline service for `LastUtteranceSearcher`. It should be similar to this [tutorial](https://asyml-forte.readthedocs.io/en/latest/notebook_tutorial/tutorial_eliza.html#Create-Your-Own-Chatbot-Service). The instruction will no longer require users to define the environment variables since they should be already configured in the service pipeline, but we still need to include the steps of starting a remote service.
  * [utterance_searcher.py](https://github.com/asyml/forte/blob/master/examples/clinical_pipeline/utterance_searcher.py) needs to be refactored. Currently it's directly modifying the `stave` database, and this is not a safe operation. We should utilize backend APIs to interact with `stave` database. A [wrapper](https://github.com/asyml/stave/blob/master/simple-backend/stave_backend/lib/stave_session.py) of these APIs is provided in `stave` backend lib.
  * The example requires user to setup two `stave` projects and to manually configure the database path. With the new [interfaces](https://github.com/mylibrar/stave/blob/master/simple-backend/stave_backend/lib/) provided by `stave`, we can automate these steps in a python script. Ideally users only need to run a python script, which should pop out a chatbot window for users to enter queries.

Piyush13y:
This PR fixes #745. 

### Description of changes
We provide a remote service pipeline that handles interaction between the chatbot and the elastic search indexer, to get the results visualized through stave on the UI.
We also create 2 stave projects in the script, one for annotation viewing and the other to act as the search engine/chat bot.


examples/clinical_pipeline/clinical_processing_pipeline.py
```diff
@@ -1,48 +1,153 @@
...
+        chat_doc_json["project_id"] = project_id_chat
+        doc_id = session.create_document(chat_doc_json)
+        project_list = session.get_project_list().json()
+
+    return project_id_base
+
+
+def main(
+    input_path: str, output_path: str, max_packs: int = -1, run_ner_pipeline=0
+):
```
hunterhector, line 79: If I ran this I will have `ModuleNotFoundError: No module named 'ftx.medical`, instructions to add the ontology?

examples/clinical_pipeline/clinical_processing_pipeline.py
```diff
@@ -1,48 +1,153 @@
...
-    pl.add(
-        PackIdJsonPackWriter(),
-        {
-            "output_dir": output_path,
-            "indent": 2,
-            "overwrite": True,
-            "drop_record": True,
-            "zip_pack": True,
-        },
+    if run_ner_pipeline == 1:
```
hunterhector, line 84: Why are we placing two scripts into this one `main` function? One is this NER pipeline, and the other is the search pipeline. Just make them two scripts so that we don't have to worry about to use this `run_ner_pipeline` parameter

examples/clinical_pipeline/clinical_processing_pipeline.py
```diff
@@ -1,48 +1,153 @@
...
+def get_json(path: str):
+    file_obj = open(path)
+    data = json.load(file_obj)
+    file_obj.close()
+    return data
+
+
+def update_stave_db(
+    default_project_json, chat_project_json, chat_doc_json, config
+):
```
hunterhector, line 44: Can we explain what the function is doing? And what are the parameters?

hunterhector (commented):
- An explanation of what is happening here would be very helpful, basically how the data flows.
- Will the `chat_project_old` file be removed?

Piyush13y:
@hunterhector I am assuming we are planning to abandon this PR, now that we have a PR in ForteHealth to add this example there? Or do we want to push it here too?

hunterhector:
> @hunterhector I am assuming we are planning to abandon this PR, now that we have a PR in ForteHealth to add this example there? Or do we want to push it here too?

If everything is moved there we should remove one of them (probably this one). Otherwise, we will have to maintain multiple versions

Piyush13y:
> > @hunterhector I am assuming we are planning to abandon this PR, now that we have a PR in ForteHealth to add this example there? Or do we want to push it here too?
> 
> If everything is moved there we should remove one of them (probably this one). Otherwise, we will have to maintain multiple versions

We will then have to remove the whole example here then? Because it won't work the old way now that Stave uses RemoteProcessors. As in, removing the clinical_processing_pipeline folder itself from forte/examples; or we will have to strip some functionalities and just keep a non stave interaction example, but that again would be an effective copy of our MIMIC3 example in ForteHealth? https://github.com/asyml/ForteHealth/blob/master/examples/mimic_iii/medical_pipeline.py

hunterhector:
> > > @hunterhector I am assuming we are planning to abandon this PR, now that we have a PR in ForteHealth to add this example there? Or do we want to push it here too?
> > 
> > 
> > If everything is moved there we should remove one of them (probably this one). Otherwise, we will have to maintain multiple versions
> 
> We will then have to remove the whole example here then? Because it won't work the old way now that Stave uses RemoteProcessors. As in, removing the clinical_processing_pipeline folder itself from forte/examples; or we will have to strip some functionalities and just keep a non stave interaction example, but that again would be an effective copy of our MIMIC3 example in ForteHealth? https://github.com/asyml/ForteHealth/blob/master/examples/mimic_iii/medical_pipeline.py

I think we can just remove the whole thing since we have a complete example in the other repo

hunterhector:
close since most of the code is moved out and this pr is inactive

PR #796, opened 2022-05-20
Initialize new entry attribute list with default values

### LINKED ISSUE #786: Initialize new entry data from the typing of attributes
In [`DataStore._new_annotation`](https://github.com/asyml/forte/blob/master/forte/data/data_store.py#L322), [`DataStore._new_link`](https://github.com/asyml/forte/blob/master/forte/data/data_store.py#L347) and [`DataStore._new_group`](https://github.com/asyml/forte/blob/master/forte/data/data_store.py#L368), we initialize all the unset attributes to `None`, which might not reflect the correct typing of attributes. Ideally each attribute should be initialized based on its typing annotation. For example, attribute whose annotation is `Optional[...]` can still be initialized as `None`, but attribute of `List[...]` or `FList[...]` should be initialized as empty list. Similarly, attribute of `Dict[...]` or `FDict[...]` should be initialized as empty dictionary.

We can extract the typing alias using `typing.get_origin(Entry.__dataclass_fields__[attribute_name].type)`, which can be used to determine the typing class of each annotated fields in the corresponding entry.

Any better solutions or designs are welcome. 

**Additional context**
This should be merged into `master` branch.

mylibrar:
This PR fixes #786. 

### Description of changes
* In `DataStore._new_annotaion()`, `DataStore._new_link()` and `DataStore._new_group()`,  each attribute will be initialized based on its typing annotation.
  * FDict, Dict -> {}
  * FList, List -> []
  * other types (premitives, Union, NoneType, etc.) -> None

### Test Conducted
`test_add_annotation_raw()` is updated to validate this new initialization behavior.

### DIFF
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -12,11 +12,11 @@
 # limitations under the License.
 
 from typing import Dict, List, Iterator, Tuple, Optional, Any
-from typing_inspect import get_origin
 import uuid
 from bisect import bisect_left
 from heapq import heappush, heappop
 from sortedcontainers import SortedList
+from typing_inspect import get_origin
 
 from forte.utils import get_class
 from forte.data.base_store import BaseStore
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -338,9 +338,9 @@ class DataStore(BaseStore):
         attr_list: List = [None] * len(attr_dict)
         for attr_name, attr_id in attr_dict.items():
             attr_class = get_origin(attr_fields[attr_name].type)
-            if attr_class in (FList, list):
+            if attr_class in (FList, list, List):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = []
-            elif attr_class in (FDict, dict):
+            elif attr_class in (FDict, dict, Dict):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = {}
         return attr_list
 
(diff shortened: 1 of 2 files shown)


forte/data/data_store.py
```diff
@@ -318,18 +320,32 @@ def _get_type_parent(self, type_name: str) -> str:
...
         Returns:
-            attr_dict (dict): The attributes-to-index dict of an entry.
+            attr_dict (list): A list of attributes with default values.
         """
-        return len(self._get_type_attribute_dict(type_name))
+        attr_dict: Dict = self._get_type_attribute_dict(type_name)
+        attr_fields: Dict = self._get_entry_attributes_by_class(type_name)
+        attr_list: List = [None] * len(attr_dict)
+        for attr_name, attr_id in attr_dict.items():
+            attr_class = get_origin(attr_fields[attr_name].type)
```
hunterhector, line 340: This is now inspect on the fly, but it is probably better to keep a record of this. Maybe mark this as a todo?

hunterhector: or otherwise a simple approach is to cache this in the class

hunterhector (commented):
Only one comment, can be resolved either by:
1. create a future todo (issue)
2. create a cache if that's easy to do (hopefully store that cache with some other variables that we have, and create a function to access the cache)

feipenghe (approved):

Issue #773, opened 2022-05-05
Add new generics to DataStore


mylibrar:
**Is your feature request related to a problem? Please describe.**
Implement `add_generics_raw(self, type_name: str)` in `forte/forte/data/data_store.py`.
This function adds an `Generics` entry to the sortedlist at `type_name` of the array which records all sortedlists. It returns the tid for the entry.

**Describe the solution you'd like**
A helper function `_new_generics()` can be added to generate a generics type entry data with default fields. A reference to the entry should be stored in both self.elements and self.entry_dict.

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.

Issue #772, opened 2022-05-05
Add new audio annotation to DataStore


mylibrar:
**Is your feature request related to a problem? Please describe.**
Implement `add_audio_annotation_raw(self, type_name: str, begin: int, end: int)` in forte/forte/data/data_store.py.
This function adds an `AudioAnnotation` entry with `begin` and `end` index to the sortedlist at `type_name` of the array which records all sortedlists. It returns the tid for the entry.

**Describe the solution you'd like**
We should create the `entry data` with the format `[begin, end, tid, type_id, ,attr_1, ..., attr_n]`. Attr_1 to attr_n are attributes for this type and are different for each type. They are initialized as None value. A helper function `_new_audio_annotation()` can be added to generate an audio annotation type entry data with default fields. A reference to the entry should be stored in both self.elements and self.entry_dict.

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.

PR #800, opened 2022-05-23
Datastore Generics & AudioAnnotation

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/773 and https://github.com/asyml/forte/issues/772

### Description of changes

### Test Conducted


forte/data/data_store.py
```diff
@@ -428,6 +459,23 @@ def _new_group(
...
+
+        Args:
+            type_name: The fully qualified type name of the new entry.
+            tid: ``tid`` of the generics entry.
+
+        Returns:
+            A list representing a new generics type entry data.
+        """
+        tid: int = self._new_tid() if tid is None else tid
+        entry = [tid, type_name]
```
mylibrar, line 476: ``tid`` and ``type_name`` should always be the third and the fourth fields (refer to https://github.com/asyml/forte/blob/961c5df923ebfe2e8d907093266d3f5ecfa4aa57/forte/data/data_store.py#L85). In `Generics` maybe we can put some placeholder like `None` in the first and second fields if there is no better ideas.

hunterhector: thanks, we can use the constants to get these positions

tests/forte/data/data_store_test.py
```diff
@@ -584,6 +584,66 @@ def test_add_annotation_raw(self):
...
+        # test add Document entry
+        tid_recording: int = self.data_store.add_audio_annotation_raw(
+            "ft.onto.base_ontology.Recording", 1, 5
+        )
+        # test add Sentence entry
+        tid_audio_utterance: int = self.data_store.add_audio_annotation_raw(
+            "ft.onto.base_ontology.AudioUtterance", 5, 8
+        )
+        tid_utterance: int = self.data_store.add_audio_annotation_raw(
+            "ft.onto.base_ontology.Utterance", 5, 8
```
mylibrar, line 597: `ft.onto.base_ontology.Utterance` is not a subclass of `AudioAnnotation`. Not sure if it's okay to add it using `add_audio_annotation_raw`.

tests/forte/data/data_store_test.py
```diff
@@ -584,6 +584,66 @@ def test_add_annotation_raw(self):
...
+            self.data_store._DataStore__elements[
+                "ft.onto.base_ontology.Utterance"
+            ]
+        )
+        self.assertEqual(num_doc, 1)
+        self.assertEqual(num_sent, 1)
+        self.assertEqual(num_utterance, 1)
+        tid = 77
+        self.data_store.add_annotation_raw(
+            "ft.onto.base_ontology.Recording", 0, 1, tid
```
mylibrar, line 619: `ft.onto.base_ontology.Recording` is a subclass of `AudioAnnotation` and maybe we should use `add_audio_annotation_raw` here.

forte/data/data_store.py
```diff
@@ -516,12 +564,17 @@ def _add_entry_raw(
...
                 self.__elements[type_name].add(entry)
-        elif entry_type in [Link, Group]:
+        elif entry_type in [Link, Group, Generics]:
             try:
                 self.__elements[type_name].append(entry)
             except KeyError:
                 self.__elements[type_name] = []
                 self.__elements[type_name].append(entry)
+        else:
+            raise NotImplementedError(
```
mylibrar, line 574: We can add this error to `Raises:` section in docstring

tests/forte/data/data_store_test.py
feipenghe: Fixed it.

tests/forte/data/data_store_test.py
feipenghe: Fixed it.

forte/data/data_store.py
feipenghe: Added it.

forte/data/data_store.py
```diff
@@ -372,6 +378,31 @@ def _new_annotation(
 
         return entry
 
+    def _new_audio_annotation(
+        self, type_name: str, begin: int, end: int, tid: Optional[int] = None
+    ) -> List:
+        r"""This function generates a new audio annotation with default fields.
+        All default fields are filled with None.
```
mylibrar, line 385: Actually not all fields are initialized with `None`.Refer to https://github.com/asyml/forte/blob/e2ce0a4684c884068f88d4706184304cddedb1f7/forte/data/data_store.py#L325 for details.

forte/data/data_store.py
```diff
@@ -427,6 +458,23 @@ def _new_group(
 
         return entry
 
+    def _new_generics(self, type_name: str, tid: Optional[int] = None):
+        r"""This function generates a new generics with default fields. All
+        default fields are filled with None.
```
mylibrar, line 463: The same here. Not all fields are initialized with `None`.

forte/data/data_store.py
```diff
@@ -593,6 +647,35 @@ def add_annotation_raw(
 
         return self._add_entry_raw(Annotation, type_name, entry)
 
+    def add_audio_annotation_raw(
+        self, type_name: str, begin: int, end: int, tid: Optional[int] = None
```
mylibrar, line 651: For annotation-like entries, we can put `allow_duplicate` here. You may refer to https://github.com/asyml/forte/blob/e2ce0a4684c884068f88d4706184304cddedb1f7/forte/data/data_store.py#L556 to see how to incorporate `allow_duplicate` into this function.

forte/data/data_store.py
feipenghe: I just need to remove the docstring `All default fields are filled with None.` right?

mylibrar: yes

forte/data/data_store.py
```diff
@@ -581,7 +633,75 @@ def add_annotation_raw(
...
+
+        return self._add_entry_raw(Annotation, type_name, entry)
+
+    def add_audio_annotation_raw(
+        self,
+        type_name: str,
+        begin: int,
+        end: int,
+        tid: Optional[int] = None,
+        allow_duplicate=True,
```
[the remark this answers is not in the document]
feipenghe: Added it.

forte/data/data_store.py
```diff
@@ -559,11 +605,12 @@ def _add_entry_raw(
 
         Raises:
             KeyError: raised when the entry type name is not in `self.__elements`.
+            NotImplementedError: raised when add the entry is not implemented.
```
hunterhector, line 608: Well, this sentence is grammatically weird, maybe:
```suggestion
            NotImplementedError: raised when the entry type being added is not supported, currently supporting `Annotation`, `AudioAnnotation`, `Link`, `Group`, `Generics`
```

Btw, how about `Query`: https://github.com/asyml/forte/blob/master/forte/data/ontology/top.py#L38

I think we need to support all the `top` types.

forte/data/data_store.py
```diff
@@ -638,7 +690,78 @@ def add_annotation_raw(
...
+        self,
+        type_name: str,
+        begin: int,
+        end: int,
+        tid: Optional[int] = None,
+        allow_duplicate=True,
+    ) -> int:
+
+        r"""This function adds an audio annotation entry with ``begin`` and
+        ``end``
```
hunterhector, line 711: could consider reorganizing here to make the text has natural line breaks. Note that people would also read the source code directly so documentation is helpful at that time.

forte/data/data_store.py
```diff
@@ -638,7 +690,78 @@ def add_annotation_raw(
...
+        if not allow_duplicate:
+            tid_search_result = self._get_existing_entry_tid(entry)
+            # if found existing entry
+            if tid_search_result != -1:
+                return tid_search_result
+        return self._add_entry_raw(AudioAnnotation, type_name, entry)
+
+    def _get_existing_entry_tid(self, entry: List[Any]):
+        r"""
+        This function searches for tid for existing entry tid. It return tid
```
hunterhector, line 746: Should document that we search for `Annotation`-like data, and provide that which types can be searched here (and based on what to search)

tests/forte/data/data_store_test.py
```diff
@@ -642,6 +642,66 @@ def test_add_annotation_raw(self):
...
+            "ft.onto.base_ontology.Recording", 1, 5
+        )
+        # test add Sentence entry
+        tid_audio_utterance: int = self.data_store.add_audio_annotation_raw(
+            "ft.onto.base_ontology.AudioUtterance", 5, 8
+        )
+        tid_utterance: int = self.data_store.add_annotation_raw(
+            "ft.onto.base_ontology.Utterance", 5, 8
+        )
+        num_doc = len(
```
hunterhector, line 657: these varaible names do not match the entry type. not that this is wrong, but a bit confusing to maintain, probably:

```suggestion
        num_recording = len(
```

here instead.

hunterhector (commented):
Mainly documentation issues, but `Query` from the `top.py` is not supported in the `add` section

forte/data/data_store.py
hunterhector: I see, you are probably right.

forte/data/data_store.py
feipenghe: I think we might need to support non-Annotation like data in the future. So I am raising errors with messages to inform users.

hunterhector: actually, we probably won't do that, the purpose of this function is only searching duplicate in position (e.g begin and end). but this is not quite important

hunterhector:
Let's just pass the CI and merge this.

mylibrar (approved):

Issue #802, opened 2022-05-24
Create a Bounding Box ontology and an IOU computation function


feipenghe:
**Is your feature request related to a problem? Please describe.**
We would like a `BoundingBox` in our ontology and computation of Intersection Over Union.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #813, opened 2022-05-27
Bounding box

feipenghe:
This PR fixes https://github.com/asyml/forte/issues/802. 

### Description of changes
* create ontologies ImageAnnotation -> Region -> Box -> BoundingBox
* implement box function check overlapping and compute iou
* create an example of checking overlapping and compute iou

### Test Conducted

### DIFF
--- a/forte/data/data_pack.py
+++ b/forte/data/data_pack.py
@@ -169,8 +169,8 @@ class DataPack(BasePack[Entry, Link, Group]):
 
         self._data_store: DataStore = DataStore()
         self._entry_converter: EntryConverter = EntryConverter()
-        self.image_annotations: SortedList[ImageAnnotation] = SortedList()
-        self.grids: SortedList[Grids] = SortedList()
+        self.image_annotations: List[ImageAnnotation] = []
+        self.grids: List[Grids] = []
         self.payloads: List[np.ndarray] = []
 
         self.__replace_back_operations: ReplaceOperationsType = []
--- a/forte/data/data_pack.py
+++ b/forte/data/data_pack.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from multiprocessing.sharedctypes import Value
 from pathlib import Path
 from typing import (
     Dict,
@@ -412,6 +413,11 @@ class DataPack(BasePack[Entry, Link, Group]):
         return self._audio[begin:end]
 
     def get_image_array(self, image_payload_idx: int):
+        if image_payload_idx >= len(self.payloads):
+            raise ValueError(
+                f"The input image payload index{(image_payload_idx)}"
+                f" out of range. It should be less than {len(self.payloads)}"
+            )
         return self.payloads[image_payload_idx]
 
     def set_text(
--- a/forte/data/data_pack.py
+++ b/forte/data/data_pack.py
@@ -42,12 +42,13 @@ from forte.data.ontology.core import Entry
 from forte.data.ontology.core import EntryType
 from forte.data.ontology.top import (
     Annotation,
-    ImageAnnotation,
     Link,
     Group,
     SinglePackEntries,
     Generics,
     AudioAnnotation,
+    ImageAnnotation,
+    Grids
 )
 from forte.data.span import Span
 from forte.data.types import ReplaceOperationsType, DataRequest
@@ -167,8 +168,9 @@ class DataPack(BasePack[Entry, Link, Group]):
         self.generics: SortedList[Generics] = SortedList()
         self.audio_annotations: SortedList[AudioAnnotation] = SortedList()
         self.image_annotations: SortedList[ImageAnnotation] = SortedList()
+        self.grids: SortedList[Grids] = SortedList()
+        self.payloads = []
 
-        self.image_payloads = []
         self.__replace_back_operations: ReplaceOperationsType = []
         self.__processed_original_spans: List[Tuple[Span, Span]] = []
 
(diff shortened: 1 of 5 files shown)


forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+
+class Grids(Entry):
+    """
+    Regular grids with a grid configuration.
+
+    Args:
+        pack: The container that this grids will be added to.
+        height: the number of grid cell per column.
+        width: the number of grid cell per row.
+        image_payload_idx: the index of image in the datapack payloads.
```
hunterhector, line 744: Need to mention this is optional and what `None` implies here.

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+
+
+class Region(ImageAnnotation):
+    def __init__(self, pack: PackType, image_payload_idx: int):
+        super().__init__(pack, image_payload_idx)
+        if image_payload_idx is None:
+            self._image_payload_idx = 0
+        else:
+            self._image_payload_idx = image_payload_idx
+        self.img_arr = self.pack.get_image_array(self._image_payload_idx)
```
hunterhector, line 863: I think we don't need to store this variable:

1. if it is just a pointer, it is trivial to use anyway, you can create a property function instead of a variable
2. if it is actual memory copy, it won't be necessary

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+
+    @property
+    def box_min_y(self):
+        return max(self._cy - round(0.5 * self._height), 0)
+
+    @property
+    def box_max_y(self):
+        return min(self._cy + round(0.5 * self._height), self.max_y)
+
+    def is_overlapped(self, other):
```
hunterhector, line 917: need to document what overlap means, especially on the corner/edge cases

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+            grid_cell_h_idx (int): the height index of the associated grid cell.
+            grid_cell_w_idx (int): the width index of the associated grid cell.
+
+        """
+
+        self.grids = Grids(pack, grid_height, grid_width, image_payload_idx)
+        cy, cx = self.grids.get_grid_cell_center(
+            grid_cell_h_idx, grid_cell_w_idx
+        )
+        super().__init__(pack, image_payload_idx, cy, cx, height, width)
```
hunterhector, line 961: better make init first line

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+
+
 SinglePackEntries = (Link, Group, Annotation, Generics, AudioAnnotation)
```
hunterhector, line 964: images should be placed in the singlepackentries

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+        else:
+            self._image_payload_idx = image_payload_idx
+
+        self.img_arr = self.pack.get_image_array(self._image_payload_idx)
+        self.c_h, self.c_w = (
+            self.img_arr.shape[0] // self._height,
+            self.img_arr.shape[1] // self._width,
+        )  # compute the height and width of grid cells
+
+    def get_grid_cell(self, h_idx: int, w_idx: int):
```
hunterhector, line 773: Need to define the convention on the meaning, the direction of `x`, `y`, is it counted from the top left corner, or from the bottom left corner.

forte/data/ontology/top.py
```diff
@@ -686,5 +690,276 @@ def get(
...
+
+
+class Box(Region):
+    """ """
+
+    def __init__(
+        self,
+        pack: PackType,
+        image_payload_idx: int,
+        cy: int,
```
hunterhector, line 873: similarly, need to define the convention on the meaning, the direction of x, y, is it counted from the top left corner, or from the bottom left corner.

forte/data/data_pack.py
```diff
@@ -165,6 +167,9 @@ def __init__(self, pack_name: Optional[str] = None):
         self.groups: SortedList[Group] = SortedList()
         self.generics: SortedList[Generics] = SortedList()
         self.audio_annotations: SortedList[AudioAnnotation] = SortedList()
+        self.image_annotations: SortedList[ImageAnnotation] = SortedList()
+        self.grids: SortedList[Grids] = SortedList()
```
hunterhector, line 171: We probably don't have a reason to make the sorted lists.

forte/data/data_pack.py
```diff
@@ -165,6 +167,9 @@ def __init__(self, pack_name: Optional[str] = None):
...
+        self.payloads: List[np.ndarray] = []
```
hunterhector, line 172: I thought payloads are `Payload` classes, why `ndarray`?

forte/data/ontology/top.py
```diff
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from dataclasses import dataclass
+from email.mime import image
 from functools import total_ordering
+from multiprocessing.sharedctypes import Value
```
hunterhector, line 17: Some unnecessary imports comes back again, please check the auto imports if you are using them

forte/data/ontology/top.py
feipenghe: It relies on `self.grids.get_grid_cell_center()` to get parameter `cy`, `cx`.

hunterhector: You can inline that I think

hunterhector: You can inline the statement on line 958 into 961
```
        super().__init__(pack, image_payload_idx, *self.grids.get_grid_cell_center(grid_cell_h_idx, grid_cell_w_idx), height, width)
```

forte/data/ontology/top.py
feipenghe: defined.

forte/data/ontology/top.py
feipenghe: defined.

forte/data/ontology/top.py
feipenghe: placed it.

mylibrar (approved):

Issue #808, opened 2022-05-25
New test cases after introducing new features via data store


hunterhector:
**Is your feature request related to a problem? Please describe.**
Datastore features are tested independently, and the current set of data pack test do not cover all new features. We would try to cover these new features through the data pack interface

**Describe the solution you'd like**
The test cases are:

- [ ] Serialization/deserialization check

**Describe alternatives you've considered**

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

PR #814, opened 2022-05-27
Datastore Integration

### LINKED ISSUE #573: Implement entry_to_tuple and tuple_to_entry functions
Even though we plan to change the data structure, we do want to support old features and data structures so it is compatible with users who use the old data structure. We want two functions that convert new data tuples to old data packs and vice versa.

**Depends on**
The new data structure interfaces in Issue #571.

Entry_to_tuple should take in an entry object and get all its attributes through class methods. Then, it stores all the attributes in a new tuple according to a specified index dictionary. 
For example, here is the index dictionary of the sentence entry.
sent_struct: {
        “text”:3,
        “speaker”:4, 
        “part_id”:5, 
        “sentiment”:6, 
        “classification”: 7, 
        “classifications”: 8
}
We want the output tuple to be of structure "(tid, begin, end, text, speaker, part_id, sentiment, classification, classifications)".

Tuple_to_entry should take in a tuple, which has the structure of (tid, begin, end, **attributes), and its type such as 'token', 'sentence', etc. It needs to retrieve the specified index dictionary according to its type. Then, it accesses all the attributes through indexes.  Lastly, it creates a new entry object with the obtained attributes and the index dictionary.

**Additional context**

mylibrar:
This PR fixes #573 . 

### Description of changes
* Update the `_pending_entries` mechanism
  * Remove `BasePack.regret_creation` and its corresponding test cases.
  * For `DataPack`, `_pending_entries` will store `entry.tid` instead of `entry` itself.
    * Update typing annotations on several related methods including `add_entry`, `record_entry`, `_add_entry`, `__add_entry_with_check`, etc.
    * TODO: After `DataStore` is integrated with `MultiPack`, we should only maintain a mapping from entry's tid to the corresponding component, i.e., Dict[int, Optional[str]].
* Remove all the `SortedList` containers of entries in `DataPack`
  * Entries will be stored in `DataPack._data_store`
  * All the CRUD operations are directed to `DataStore`
  * Entry class object will be generated on the fly. `EntryConverter` is created to facilitate the conversion between entry data from `DataStore` and the class object.
* Serialization
  * Add pack version checking for deserialization of `DataPack`
    * We skipped several unit tests for outdated serialization format, including `stave_reader_test` and `entry_data_structures_test`
  * Keep `_onto_file_path` and `_dynamically_add_type` in `DataStore.__getstate__` to avoid failure in test cases.
* `FList`/`FDict`
  * They will no longer use `Pointer` to resolve entries
  * Their payload will be list/dict reference from `DataStore`
* Top level entries: `Annotation`, `Group`, `Link`, `Generics`, `AudioAnnotation`
  * All the attributes (including begin, end, parent, child, members, etc.) are now directed to `DataStore`
* Some fixes in `DataStore`
  * `all_entries()` should generate entries in order, so now we should use `co_iterator_annotation_like()` for annotation-like entries.
  * `get()` should not required `type_name` to exist in `__elements` when `include_sub_type` is enabled if it's not `Generics`.

The current integration is probably going to be unstable, hence it's only for [0.3 pre-release (0.3.0dev1)](https://github.com/asyml/forte/milestone/1). We will need to fix a lot of follow-up issues in [0.3 stable version](https://github.com/asyml/forte/milestone/2) and [0.3 interface clearance](https://github.com/asyml/forte/milestone/3).

### Test Conducted
The integration should successfully pass all the current CI tests. Will need more dedicated test cases in future (#808).


hunterhector (commented):
Let's set up the versions (https://github.com/asyml/forte/blob/master/forte/version.py#L23) to make sure Forte will detect that old serialization won't be supported.

hunterhector (approved):

PR #823, opened 2022-06-04
fix a little bug

Leolty:
This PR fixes [issue link]. 

### Description of changes

### Test Conducted

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,8 @@ class NLTKPOSTagger(PackProcessor):
     def _process(self, input_pack: DataPack):
         # get a list of token data entries from `input_pack`
         # using `DataPack.get()`` method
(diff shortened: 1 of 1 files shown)


Leolty:
in the following code:
`for token, tag in zip(token_entries, taggings):`
token_entries is used without definition

hepengfe:
Though it's a minor fix. I have few suggestions from the perspective of the CI pipeline.
1. Create an issue and describe it under the prompts.
2. Create an PR and mention the issue. Also, describe the PR under the prompts. (Rather than an extra comment)
3. Assign a label for it.

Leolty:
Thanks a lot for your suggestions!! It was my first time doing this so I did not know what I should do. I will redo it as your instructions.

Issue #817, opened 2022-05-28
New Grid Ontology Design


hepengfe:
**Describe the bug**
Currently we have `Girds` which is holding CV grid. Initially we want the grid class to hold all grid configuration but it's actually not necessary.
* rename `Grids` -> `Grid`
* create a default grid for bounding box beforehand so that it doesn't create a grid in `BoundingBox.__init__`

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

hepengfe:
@hunterhector I kinda forgot your idea of creating a default grid for `BoundingBox`. 
Do we create a default grid in `BoundingBox.__init__()` and users need to manually set it later?

hunterhector:
The idea I suggested is to create a default grid whenever necessary, but the creation time might not be related to only `BoundingBox`, it should be anything that requires a grid. Some potential solutions are listed below:

1. Create the default grid when creating the image-related payload, and associate the default grid with the payload.
2. Create an `add_or_get_default_grid` function to the image Payload class, but don't create the grid right away. When any other class like `BoundingBox` require the grid, but the user doesn't provide the grid, we can call the `add_or_get_default_grid` function, which will always return a grid (and create one if called the first time).

I don't have a specific reason for which one is better, but the things to consider are:
1. when we create the grids automatically, will the user becomes confused? For example, the default grid shouldn't intervene or overwrite anything a user creates, and vice versa, if the user creates a grid but does not attempt to change the default grid, we shouldn't change the default grid.
2. when to create the default grid, and how can a user change the grid if one wants to?

PR #827, opened 2022-06-07
Grids Ontology Improvement

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/817. 

### Description of changes
* change `Grids` -> `Grid` and move it to `core.py` as a data structure rather than an ontology
    * dealing edge cases of residual grid cells(last grid cells per row/column). We use the image size as upper bound for these grid cells so they might be smaller than regular grid cells. 
* improve grid initialization in `BoundingBox`
* add more properties and functions in `Box` based on three use cases.
    * A standalone box (not associated with grid)
    * Ground truth bounding box. (known absolute position and shape)
    * Predicted bounding box. (associated with a grid cell and known height/width offsets)
* add complete docstring with use cases in `Box`
* `Box` condition attributes
    * `self._is_grid_associated`
    * `self._is_box_center_set()`

### Test Conducted


hepengfe:
The implementation design considerations is:
1. Regular shape CV objects should have a reference point. In our case, Box is an example and BoundingBox is even more special that its reference point needs to be center of grid cell.
2. Due to the limitation of ontology object `__init__`, we can not let `Grid` be the parameter to initialize `BoundingBox` and associate `BoundingBox` and `Grid`. 
3. Based on the considerations above, I initialize Box center (reference point) as the image center and record whether it's a default box center and whether it's associated with a grid. And it will give corresponding Warning/ValueError appropriately. 
4. I provide methods to set box center(reference point) based on user input point location `(cy, cx)` or `Grid`.

Several questions about the design:
I think `Grid` should be a class that is somehow independent from `DataPack` and specific image payload. As I think `Grid` is a just an interface that able to provides grid cell location given an image shape. As images in a dataset is basically the same, we might shouldn't create a `Grid` for every image. But it seems like `Grid` is bound to DataPack and DataPack contains few images. If we store grid inside DataPack, we might have to duplicate grid for each image.

hepengfe:
IMG_F121392AF097-1

Usually, there are many regular ontologies(such as bounding box) and many images but only few grids. The number of grids actually needed only depends on the number of different image sizes and the number of grid configurations. 

My proposal is that
Past design: Each datapack stores few images and store few grids. The number of grids in total is dependent the number of images. With a large datasets, there will be many duplicated grids distributed in many DataPack.

New design: we implement a Grid interface such that only a collection of grids (probably tens of them) is needed to query-related data given image and grid-related parameters/regular ontology such as BoundingBox. Grid itself just like other ontology doesn't store image data. In addition, it is not bound to any other regular ontology. 

If this design is expected, we might not want it to inherit entry as it will create new ones for each DataPack.

hunterhector:
> Usually, there are many regular ontologies(such as bounding box) and many images but only few grids. The number of grids actually needed only depends on the number of different image sizes and the number of grid configurations.
> My proposal is that Past design: Each datapack stores few images and store few grids. The number of grids in total is dependent the number of images. With a large datasets, there will be many duplicated grids distributed in many DataPack.
> New design: we implement a Grid interface such that only a collection of grids (probably tens of them) is needed to query-related data given image 

Good thoughts on limiting "number of grids", feels like a good idea, though I haven't considered everything about it. 

> and grid-related parameters/regular ontology such as BoundingBox. Grid itself just like other ontology doesn't store image data. In addition, it is not bound to any other regular ontology.

I have a few suggestions on this detail though. You can already see the problem of making `Grid` be like other ontology items. First, it cannot be stored as data along with each DataPack (since this will cause duplications). Second, it cannot inherit an Entry. To do so, this basically means we shouldn't consider `Grid` inside the ontology framework.

Instead, if you want to achieve this, we probably will make Grid a basic data structure used by the Forte DataPack system only. (i.e. Grid is more like a List, SortedList, TList or Adjacent Array, instead of an Entry). In DataPack, we can create some default Grids and users won't involve in this process, they are the underlying data structures that speed up certain computations only.

Having `Grid` as a new data structure does not mean that users won't be able to use it, for example, we can imagine supporting `Grid` as an attribute type for a certain ontology (i.e Grid is not used as an Entry type, but can be used as an attribute type, like TList). Consider the following specification:

```json
{
  "entry_name": "User.Defined.Type",
  "entry_attributes":
   {
       "grid": "forte.data.core.grid",      
   }
}
```

We can allow users to invoke some functions like computing the position of a particular bounding box with any grids, even the one they defined.  However, these should be advanced and rare use cases for now.

hunterhector (commented):
To continue this PR, let's make `Grid` a data structure, like [`Span`](https://github.com/asyml/forte/blob/master/forte/data/span.py#L25).

hunterhector (commented):
1. You can store `Grid` in the data store, but don't store them like other entries, because they are not.
2. I think you haven't handled (or at least haven't documented) the behavior of the edge cases well. How do we treat the pixels on the side where  ` y > c_h * self._height`. Let's make sure we have a common understanding of the behavior, test the edge cases and write the documentation for it

forte/data/ontology/core.py
```diff
@@ -865,5 +867,137 @@ def index_key(self) -> int:
...
+        Args:
+            h_idx: the height(row) index of the grid cell in the grid,
+                , the unit is one image array entry.
+            w_idx: the width(column) index of the grid cell in the
+                grid, the unit is one image array entry.
+
+        Returns:
+            A tuple of (y index, x index)
+        """
+        return (
```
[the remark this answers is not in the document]
hepengfe: no as the residual grid cell (index range) will be smaller than other regular ones.

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
...
+            A tuple of (y index, x index)
+        """
+        # compute the center position of the grid cell by
+        # dividing the grid cell height range and wdith range by 2 (round down)
+        # suppose an extreme case that
+        # a grid cell has
+        # a height range of (0, 3) and a width range of (0, 3)
+        # the grid cell center would be (1, 1)
+        # since the grid cell size is usually very large
+        # the minor offset of the grid cell center usually doesn't matter
```
hunterhector, line 1004: Behavior-related documentation needs to be in the docstring, not inline comments.

forte/data/ontology/core.py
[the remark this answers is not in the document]
hunterhector: these behavior needs to be in the docstring

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
...
+        if self.c_h <= 0 or self.c_w <= 0:
+            raise ValueError(
+                "cell height and width must be positive"
+                f"but the cell shape is {(self.c_h, self.c_w)}"
+                "please adjust image shape or grid shape accordingly"
+            )
+
+    def get_grid_cell(self, img_arr: np.ndarray, h_idx: int, w_idx: int):
+        """
+        Get the array data of a grid cell from image of the image payload index.
```
hunterhector, line 928: You said `image payload index` but I didn't see this param in the function. And you didn't have the docstring for `img_arr`

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
...
+        The array is a masked version of the original image, and it has
```
hunterhector, line 929: `masked version` is underspecified, what is a mask anyway?

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
...
+        the same size of the image. The array entries that are not
+        within the grid cell will masked as zeros. The array entries that are
+        within the grid cell will be copied to the zeros numpy array.
```
hunterhector, line 932: `copied to the zeros numpy array` is confusing, where did `the zeros numpy array` come from.

forte/data/ontology/core.py
```diff
@@ -865,5 +867,137 @@ def index_key(self) -> int:
...
+        # An example of computing grid height index range is
+        # index * cell height : (index + 1) * cell height.
+        # It's similar for computing cell width index range
+        array[
+            h_idx * self.c_h : (h_idx + 1) * self.c_h,
+            w_idx * self.c_w : (w_idx + 1) * self.c_w,
+        ] = img_arr[
+            h_idx * self.c_h : (h_idx + 1) * self.c_h,
+            w_idx * self.c_w : (w_idx + 1) * self.c_w,
+        ]
```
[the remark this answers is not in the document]
hunterhector: This needed to be explained in the docstring clearly.

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
...
+        return self._height
+
+    @property
+    def width(self):
+        return self._width
+
+    def __eq__(self, other):
+        if other is None:
+            return False
+        return (self.image_payload_idx, self._height, self._width) == (
```
hunterhector, line 1025: This attribute doesn't exist in the class

hunterhector: And this function is not tested.

hunterhector:
Since we want to support `Grid` as a data structure comparing to `Span`:

```
{
  "entry_name": "User.Defined.Type",
  "entry_attributes":
   {
       "grid": "forte.data.core.Grid",      
   }
}
```

We should also support/test its serializaiton

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
         return self.tid
 
 
+class Grid:
+    """
+    Regular grid with a grid configuration dependent on the image size.
+
+    We intialize and constrain the grid shape by the image shape during
+    initialization since usually there aren't many different image sizes in one dataset.
```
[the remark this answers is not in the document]
hepengfe: The idea I am trying to convey is that the grid can be totally "free-form" that we don't initialize it with any image sizes and pass the image size directly into the method/operation on the fly. However, since the number of different image shapes are limited, we can bound one grid to one image size, and we can also do the image/grid size check during the initialization.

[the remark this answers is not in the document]
hunterhector: Please reconsider the message from the user perspective, and see if they can understand what you want to do here. The users haven't gone through the whole design phase as you so they don't understand why these considerations matter.

Also, consider whether this message should be placed on the docstring (since this class can be used for and other purposes), or it should be placed in some sort of tutorial.

hepengfe:
> Since we want to support `Grid` as a data structure comparing to `Span`:
> 
> ```
> {
>   "entry_name": "User.Defined.Type",
>   "entry_attributes":
>    {
>        "grid": "forte.data.core.Grid",      
>    }
> }
> ```
> 
> We should also support/test its serializaiton

Do you mean we need to custmize `__setstate__` and `__getstate__`? I noticed that `Span` has `__repr__` function that returns a string.

forte/data/ontology/core.py
hepengfe: Removed `image payload index` and added docstring for `img_arr`

forte/data/ontology/core.py
```diff
@@ -865,5 +867,167 @@ def index_key(self) -> int:
         return self.tid
 
 
+class Grid:
+    """
+    Regular grid with a grid configuration dependent on the image size.
```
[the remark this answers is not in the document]
hepengfe: Added docstring.

mylibrar (commented):
I guess we will need a comprehensive example to showcase the power of `Grid` (maybe in another PR).

forte/data/ontology/top.py
```diff
@@ -895,17 +896,68 @@ def __eq__(self, other):
...
+
+    There are several use cases for a box:
+        1. When we use a box standalone, we need cy, cx to be set. The offset
+        between the box center and the grid cell center is not used.
+        2. When we represent a ground truth box, the box center and its shape
+        are given. We can compute the offset between the box center and the
+        grid cell center.
+        3. When we predict a box, we will have the predicted box shape (height,
+        width) and the offset between the box center and the grid cell center,
+        then we can compute the box center.
```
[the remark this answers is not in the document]
hepengfe: Yes. In the first use case, only `cy` and `cx` are required to be set.

forte/data/data_pack.py
```diff
@@ -172,7 +170,9 @@ def __init__(self, pack_name: Optional[str] = None):
         self._data_store: DataStore = DataStore()
         self._entry_converter: EntryConverter = EntryConverter()
         self.image_annotations: List[ImageAnnotation] = []
-        self.grids: List[Grids] = []
+        self.payloads: List[np.ndarray] = []
```
hunterhector, line 173: Here is an unused variable. And more, this make a new assumption that we store a list of payload in the format as `np.ndarray`. Let's not do this now (at least not in this PR).

forte/data/data_pack.py
```diff
@@ -172,7 +170,9 @@ def __init__(self, pack_name: Optional[str] = None):
...
+
+        self.grids: List[Grid] = []
```
hunterhector, line 175: This is also not used in this file.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+    It is a data structure used to retrieve grid-related objects such as grid
+    cells from the image. Grid itself doesn't store any data.
```
hunterhector, line 644: I think the core functionality of a `Grid` is used to index the image area so that we can easily compute the overlap of any two areas. Let's don't forget this. 

Btw, is there any use case you find that we need to retrieve the image inside the grid cell?

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+
+    Based the image size and the grid shape,
+    we compute the height and the width of grid cells.
+    For example, if the image size (image_height,image_width) is (640, 480)
+    and the grid shape (height, width) is (2, 3)
+    the size of grid cells (self.c_h, self.c_w) will be (320, 160).
```
hunterhector, line 650: I guess you might want to discuss how we handle the cases where the numbers are indivisible?

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+    The grid can be totally "free-form" that we don't initialize it with any
+    image size and pass the image size directly into the method/operation on
+    the fly. However, since the number of different image shapes are limited,
+    we can bound one grid to one image size, and we can also do the image/grid
```
hunterhector, line 654: This part is still confusing, try to explain this to someone who never works on this and see if they understand. For example, why does one need to bound the grid to the image size? What are "size check"?

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+    It is a data structure used to retrieve grid-related objects such as grid
+    cells from the image. Grid itself doesn't store any data.
+
+    Based the image size and the grid shape,
+    we compute the height and the width of grid cells.
+    For example, if the image size (image_height,image_width) is (640, 480)
+    and the grid shape (height, width) is (2, 3)
+    the size of grid cells (self.c_h, self.c_w) will be (320, 160).
+    The grid can be totally "free-form" that we don't initialize it with any
+    image size and pass the image size directly into the method/operation on
```
hunterhector, line 652: You said "we don't have to initialize it with any image size", but your `__init__` argument requires the image sizes

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+                "Grid height and width must be smaller than image height and width"
+            )
+
+        self._height = height
+        self._width = width
+
+        self._image_height = image_height
+        self._image_width = image_width
+
+        # if the resulting size of grid is not an integer, we round it up.
```
hunterhector, line 693: This needs to go into the docstring, users need to know.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+
+    def get_grid_cell(self, img_arr: np.ndarray, h_idx: int, w_idx: int):
+        """
+        Get the array data of a grid cell from image of the image payload index.
+        The array is a masked version of the original image, and it has
+        the same size as the original image. The array entries that are not
+        within the grid cell will masked as zeros. The image array entries that
+        are within the grid cell will kept.
+
+        The actual operation is that we intialize a target array with the same
```
hunterhector, line 716: To make this part simpler, users only need to know what they get, and what's the time complexity.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+        return array
+
+    def get_grid_cell_center(self, h_idx: int, w_idx: int) -> Tuple[int, int]:
+        """
+        Get the center position of the grid cell at the specific height index
+        and width index in the ``Grid``.
+        The computation of the center position of the grid cell is
+        dividing the grid cell height range and wdith range by 2 (round down)
+
+        Suppose an extreme case that a grid cell has a height range of (0, 3)
```
hunterhector, line 775: need to explain that these are pixels.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,203 @@ def index_key(self) -> int:
...
+        and a width range of (0, 3) the grid cell center would be (1, 1)
+        since the grid cell size is usually very large
+        the minor offset of the grid cell center usually doesn't matter
```
hunterhector, line 778: Missing `period`. And you don't need to mention "doesn't matter", it is not our decision whether this matters or not.

forte/data/ontology/top.py
```diff
@@ -895,17 +896,82 @@ def __eq__(self, other):
...
+    """
+
+    def __init__(self, pack: PackType, image_payload_idx: int = 0):
+        super().__init__(pack, image_payload_idx)
+        if image_payload_idx is None:
+            self._image_payload_idx = 0
+        else:
+            self._image_payload_idx = image_payload_idx
+
+    def compute_iou(self, other) -> float:
```
hunterhector, line 916: docstring?

forte/data/ontology/top.py
```diff
@@ -895,17 +896,82 @@ def __eq__(self, other):
...
+
+
+class Box(Region):
+    # pylint: disable=too-many-public-methods
+    """
+    A box class with a reference point which is the box center and a box
+    configuration.
+
+    Box of a certain size (height, width), as a regular shape, can be located
+    only when it has a reference point (box center here).
```
hunterhector, line 929: The introduction language is a bit weird, consider rewriting it.

forte/data/ontology/top.py
```diff
@@ -895,17 +896,82 @@ def __eq__(self, other):
...
+
+    There are several use cases for a box:
+        1. When we use a box standalone, we need the box center to be set.The offset
+        between the box center and the grid cell center is not used.
```
hunterhector, line 933: suddenly introduces a `grid cell center`, what is this?

forte/data/ontology/top.py
```diff
@@ -895,17 +896,82 @@ def __eq__(self, other):
...
+
+    For example, in the object detection task, dataset label contains a ground
+    truth box (box shape and box center).
+    The inference pipeline is that given a grid cell, we make a prediction of a
+    bounding box (box shape and box offset from the grid cell center).
+        1. If we want to locate the predicted box, we compute the box center
+        based on the second use case.
+        2. If we want to compute the loss, we need values for the center, shape
+        and offset for both boxes. And we need to compute the offset between
+        the box center and the grid cell center based on the third use case.
```
hunterhector, line 950: Can you explain the examples using the code snippets (how a box should be created?)

forte/data/ontology/top.py
```diff
@@ -914,223 +980,412 @@ def __init__(
         pack: PackType,
         height: int,
         width: int,
+        cy: Optional[int] = None,
+        cx: Optional[int] = None,
         image_payload_idx: int = 0,
     ):
-        if height <= 0 or width <= 0:
+        # assume Box is associated with Grid
```
hunterhector, line 987: associated with which?

forte/data/ontology/top.py
```diff
@@ -914,223 +980,412 @@ def __init__(
...
+            raise ValueError("cx_offset is not set")
+        return self._cx_offset
 
-class Box(Region):
-    """
-    A box class with a center position and a box configuration.
+    @property
+    def cy(self) -> int:
+        """
+        Compute and return row index of the box center in the image array.
```
hunterhector, line 1253: units?

forte/data/ontology/top.py
```diff
@@ -914,223 +980,412 @@ def __init__(
...
+                self._center_condition_check()
+                raise ValueError("cx is not set.")
+        return self._cx
 
     @property
-    def center(self):
-        return (self._cy, self._cx)
+    def box_center(self) -> Tuple[int, int]:
+        """
+        Get the box center by using the property function cy() and cx().
```
hunterhector, line 1301: don't write implementation details in function docstring, users often won't care (and might get confused).

forte/data/ontology/top.py
```diff
@@ -914,223 +980,412 @@ def __init__(
...
+            The box center in a ``Tuple`` format.
+        """
+        return (self.cy, self.cx)
 
     @property
-    def corners(self):
+    def corners(self) -> Tuple[Tuple[int, int], ...]:
         """
-        Get corners of box.
+        Compute and return the corners of the box.
```
hunterhector, line 1312: What is a corner? Which corners? Shape of the tuple?

forte/data/ontology/top.py
```diff
@@ -914,223 +980,412 @@ def __init__(
...
+            ValueError: if the box center is not set.
+
+        Returns:
+            The maximum y coordinate of the box.
+        """
+        return self.cy + round(0.5 * self._height)
 
     @property
-    def area(self):
+    def area(self) -> int:
```
hunterhector, line 1379: Something like `area` should also exist in the upper classes

forte/data/ontology/top.py
```diff
@@ -1179,10 +1450,76 @@ def compute_iou(self, other):
...
+        A function checks whether the offset of the box center from the grid
+        is set.
+
+        Returns:
+            True if the offset is set, False otherwise.
+        """
+        return self._cy_offset is not None and self._cx_offset is not None
+
 
 class BoundingBox(Box):
```
hunterhector, line 1520: what differences would a boundingbox have comparing to a box?

hunterhector (commented):
The main thing about this PR is the documentation. It is a bit hard for me to understand the purpose of classes and functions. We need to create a simple tutorial to explain how these things work (and also include some examples in the docstring itself). 

@mylibrar I also haven't checked the implementation details, please check that part.

forte/data/ontology/core.py
hepengfe: I think bounding box prediction is usually based on only image on inside a grid cell (anchor box) in YOLO algorithm.

hunterhector: 1. Is it a convention or just a choice of an algorithm?
2. you will need to explain these things to the users. before that, can you make sure you explain this to the team so people understand what you are doing?

forte/data/ontology/top.py
hepengfe: https://github.com/asyml/forte/issues/863

forte/data/ontology/top.py
hepengfe: Normally I don't think a box need be associated with a grid but it can be associated with grid anyway. A bounding box are much more often associated with a grid due to the CV algorithm.

forte/data/ontology/core.py
hepengfe: discussed

forte/data/ontology/top.py
hepengfe: Added code

forte/data/ontology/top.py
hepengfe: Added.

forte/data/ontology/core.py
hepengfe: Rewrote the docstring.

hepengfe:
Closed as created a new PR https://github.com/asyml/forte/pull/876
Also, `Box` is kept as it is in master branch(no changes made).

Issue #810, opened 2022-05-25
Improve Payload and apply to more data modality


hunterhector:
**Is your feature request related to a problem? Please describe.**
We created Payload ontology when supporting vision data, yet this concept can be used for Audio and text too, which will allow we supporting multiple payloads per data pack.

**Consider this an optional task for the 0.3 milestone**

**Describe the solution you'd like**
1. Double check Payload implementation to make sure it is flexible for more modalities
2. Create one more layer of ontology supporting different modalities of payload (i.e. ImagePayload, TextPaylod), and make sure these ontology definitions define enough metadata (such as original media type, etc.)
3. Create utilities to load from disk and cache internally
4. Index payloads to make sure they can be accessed (named payload would be good too)

**Describe alternatives you've considered**

**Additional context**

PR #828, opened 2022-06-08
Payload Implementation

hepengfe:
This PR partially implements https://github.com/asyml/forte/issues/810  

### Description of changes
* implement basic adding of `Payload` in DataPack and DataStore 
* implement `Payload` generic class with its subclasses for different modalities such as text, audio and image.
*  Payload can holds loading method and load path(assigned in reader), and they can be used together to achieve lazy loading in processor (rather than reader).
* modified example of audio reader to show how lazy loading works in the new design.

* modularize loading/caching data of different modalities.
* keep data independent from data ontology classes from the perspective of loading/caching such that we make sure data ontologies are lightweight

### Test Conducted


forte/data/ontology/top.py
```diff
@@ -1168,12 +1170,74 @@ def __init__(
...
+        self,
+        pack: PackType,
+        modality: str,
+        payload_idx: int,
+    ):
+        self.payload_idx = payload_idx
+        self.modality = modality
+        super().__init__(pack)
+        self.cache = []
+        self.meta = {}
```
hunterhector, line 1184: Here the `meta` info is a bit too generic. Actually, we can use the ontology system to store `meta`, so that we can have structures on the meta data system too. This way users can define new data meta information as they wish. For example, one could develop more professional meta ontology type (just an example I quickly make up):

```json
{
  "entry_name": "ImageFormat",
  "parent_entry": "Meta",
  "description": "defines meta data for image format",
  "attributes": {
     "mime": "str",
      "type_code": "str",
      "type_identifier": "str",
      "standard": "str"
  }
}
```

```python
meta = ImageFormat()
meta.mime = "image/jpeg"
meta.type_code = "jpeg"
```

hepengfe:
For audio and image data, we don't want to import third-party package directly in the payload class so I keep a loading interface (an abstract method `loading_method`) such that users can/must customize the loading method.

I rewrote the AudioReader into SoundfileAudioPayload example here https://github.com/asyml/forte/pull/828/files#diff-0a7626d45acb22318a269b5ece6049d603ad6326f8108cda7a0ed2dbc3978440. In this example, user import third-party package in the `loading_method` and customize their own payload class.

forte/data/ontology/top.py
```diff
@@ -1168,12 +1170,74 @@ def __init__(
...
+
+    def load(self, path):
+        data, data_meta = self.loading_method(path)
+        self.cache.append(data)
+        self.meta = {**self.meta, **data_meta}
```
hunterhector, line 1189: This wildcard creation basically means we don't have any control here, let's reduce such patterns

forte/data/ontology/top.py
```diff
@@ -1168,12 +1170,74 @@ def __init__(
...
+    def __init__(
+        self,
+        pack: PackType,
+        modality: str,
+        payload_idx: int,
+    ):
+        self.payload_idx = payload_idx
+        self.modality = modality
+        super().__init__(pack)
+        self.cache = []
```
hunterhector, line 1183: Are we planning multiple pieces of content here? Maybe we don't need a list of items? For example, one payload is associated with one piece of image.

hunterhector:
> For audio and image data, we don't want to import third-party package directly in the payload class 
True.

> so I keep a loading interface (an abstract method `loading_method`) such that users can/must customize the loading method.
I am considering whether this is a good idea since that means users will have to write customized code in the ontology class files (which are generated). This will create a few problems:
1. Future generated code would overwrite this.
2. The implementation cannot be passed along with a JSON spec file only.
3. If forces someone to use one single function to do the parsing, which might be a small problem.

Actually, we don't have to make the ontology this complex, the Payload ontology only needs to store the data, but it doesn't have to know how to parse the data. Like in the text processing part, we put most of the work to the reader.

Imaging the `HTMLReader`, you can consider HTML texts are special texts with customized loading methods, but we won't put the loading logic in the entry.

I would suggest the following:
1. define payload class to store how data are stored
2. define meta data ontology to store what are the data formats (like Jpeg)
3. write the reader, which will load the payload depending on the meta data

Note that we are not experts of image formats right now, so for step 2 and 3, we can only create examples, and should not make them core library code, until we fully master the format of certain data types (I think it is possible for easy data types like image stored as NumPy matrices).

forte/data/ontology/top.py
```diff
@@ -1168,12 +1170,74 @@ def __init__(
...
+class Payload(Entry):
+    def __init__(
+        self,
+        pack: PackType,
+        modality: str,
+        payload_idx: int,
+    ):
+        self.payload_idx = payload_idx
+        self.modality = modality
+        super().__init__(pack)
```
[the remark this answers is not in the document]
hepengfe: It's because `super().__init__(pack)` will use those class variables defined before it. Similar thing happened to [Annnotation](https://github.com/asyml/forte/blob/ac34f102ebb3319c1e23f2f81e20a7e9e6c5e60d/forte/data/ontology/top.py#L81)

forte/data/ontology/top.py
hepengfe: Do you mean create a meta ontology class inherits from `Entry`? Is it independent from this [Meta](https://github.com/asyml/forte/blob/ac34f102ebb3319c1e23f2f81e20a7e9e6c5e60d/forte/data/data_pack.py#L67)

forte/data/ontology/top.py
hepengfe: OK. I think it makes sense. In DataPack, we have class variables `self.payloads` that stores a collection of payloads of possibly different modalities.

forte/data/ontology/top.py
hunterhector: yeah, the `Meta` class you referred is a very old implementation and not frequently used, at one point it can be deprecated.

forte/data/ontology/top.py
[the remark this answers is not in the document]
hunterhector: hmm, so these are inevitable.

btw, we should make most variables private.

forte/data/ontology/top.py
hepengfe: created a new `Meta` class

tests/forte/data/readers/audio_reader_test.py
```diff
@@ -42,16 +44,23 @@ def initialize(self, resources: Resources, configs: Config):
         self._model = Wav2Vec2ForCTC.from_pretrained(pretrained_model)
 
     def _process(self, input_pack: DataPack):
+        # it follows the logic of loaidng while using
```
hepengfe, line 47: Lazy loading here.

hepengfe:
> > For audio and image data, we don't want to import third-party package directly in the payload class
> 
> True.
> 
> > so I keep a loading interface (an abstract method `loading_method`) such that users can/must customize the loading method.
> 
> I am considering whether this is a good idea since that means users will have to write customized code in the ontology class files (which are generated). This will create a few problems:
> 
> 1. Future generated code would overwrite this.
> 2. The implementation cannot be passed along with a JSON spec file only.
> 3. If forces someone to use one single function to do the parsing, which might be a small problem.
> 
> Actually, we don't have to make the ontology this complex, the Payload ontology only needs to store the data, but it doesn't have to know how to parse the data. Like in the text processing part, we put most of the work to the reader.
> 
> Imaging the `HTMLReader`, you can consider HTML texts are special texts with customized loading methods, but we won't put the loading logic in the entry.
> 
> I would suggest the following:
> 
> 1. define payload entry to store what data will be stored/cached
> 2. define a metadata ontology to specify the data formats (like JPEG)
> 3. write some readers, which will load the payload depending on the metadata
> 
> Note that we are not experts in image formats right now, so for steps 2 and 3, we can create examples first, and not make them core library code. Once we fully master or get comfortable with certain data types (I think it is possible for easy data types like images stored as NumPy matrices), we can move more to the library core.

1. A more general Meta entry. For one modality most loading can be handled by one or few packages, for example, for image data, `matplotlib.pyplot.imread` can handle many image formats and even url directly. What varies is that we might want to cache different target `tensor` in Payload (it needs further conversion), loading scheme(possibly iterate a directory), offloading method format (maybe load png and save it as jpg). In other words, the meta data is not only for loading but for all payload related data processing (reader, processor and offloading).

2. 
[My current way to pass the meta data](https://github.com/asyml/forte/blob/19afb61cd48d166254b430cae1e3c4ba871ba0d9/forte/data/readers/audio_reader.py#L33) might not be the best, and it passes to a reader only. I think it should be passed to pipeline which is shared by all pipeline components. Not sure this is expected.

forte/data/base_pack.py
```diff
@@ -481,9 +483,19 @@ def get_single(self, entry_type: Union[str, Type[EntryType]]) -> EntryType:
         Returns:
             A single data entry.
         """
-        for a in self.get(entry_type):
-            return a
-
+        idx = -1
+        for idx, a in enumerate(self.get(entry_type)):
+            if idx == payload_index:
+                return a
```
hunterhector, line 489: Don't understand this part, `get` is [returning a list of `Entry`](https://github.com/asyml/forte/pull/828/files#diff-08f2c9d027629fa160cbf4bbd3d8df70b58889cdd88632b3e369126c1c84d447R456) right? Why `payload_index` here?

hunterhector: Also, enumerating and checking for index looks quite weird to me.

forte/data/base_pack.py
```diff
@@ -481,9 +483,19 @@ def get_single(self, entry_type: Union[str, Type[EntryType]]) -> EntryType:
...
+        if idx < payload_index:
+            if idx == -1:
+                raise EntryNotFoundError(
+                    f"There is no {entry_type} in the provided pack."
+                )
+            raise EntryNotFoundError(
+                f"The payload index {payload_index} is larger than maximum"
+                f" {entry_type} index {idx} in the provided pack."
```
hunterhector, line 497: This message is a bit hard to understand to me. Why are we comparing `payload_index` with `entry_type` index?

forte/data/data_pack.py
```diff
@@ -244,14 +252,14 @@ def _validate(self, entry: EntryType) -> bool:
         return isinstance(entry, SinglePackEntries)
 
     @property
-    def text(self) -> str:
+    def text(self, text_payload_index=0) -> str:
         r"""Return the text of the data pack"""
-        return self._text
+        return self.get_single(TextPayload, text_payload_index).cache
```
hunterhector, line 257: I think you are misusing `get_single` here. `get_single` is used to get one object of a particular type, for example, `pack.get(Article)`.  This `get_single` is used when there is one entry of this type, or you want to take the first one for demo purposes.

In a `DataPack`, there should be multiple `TextPayload`, so I don't think you should do this.

Btw, we mentioned that we should move different types of `Payload` out and only use one function to access these `Payload`.

forte/data/data_pack.py
```diff
@@ -453,7 +461,7 @@ def groups(self):
     def groups(self, val):
         self._groups = val
 
-    def get_span_text(self, begin: int, end: int) -> str:
+    def get_span_text(self, begin: int, end: int, text_payload_index=0) -> str:
```
hunterhector, line 464: remember to update all docstring for these.

forte/data/data_pack.py
```diff
@@ -478,69 +488,101 @@ def get_span_audio(self, begin: int, end: int) -> np.ndarray:
...
 
-    def set_audio(self, audio: np.ndarray, sample_rate: int):
+        tp = TextPayload(self, text_payload_index)
+
+        tp.set_cache(text)
+        tp.meta = TextReadingMeta(self)
+
+        tp.meta.replace_back_operations = replace_back_operations
+        tp.meta.processed_original_spans = processed_original_spans
+        tp.meta.orig_text_len = orig_text_len
```
hunterhector, line 556: I feel like the `meta` should not be used to store everything we have. @mylibrar could you consider what to do with this?

forte/data/data_pack.py
```diff
@@ -478,69 +488,101 @@ def get_span_audio(self, begin: int, end: int) -> np.ndarray:
...
+
+    def set_audio(
+        self,
+        audio: np.ndarray,
+        sample_rate: int,
+        audio_payload_index: int = 0,
```
hunterhector, line 562: Please explain in docstring how the index work. Does `audio`, `text`, `image` all have different indexing schemes (i.e. all start from 0). If they share the same indices, then we shouldn't make it has a default value.

forte/data/data_pack.py
```diff
@@ -971,7 +1014,7 @@ def get_annotation_list(
                     " [Annotation, AudioAnnotation]."
                 )
 
-        def get_context_data(c_type, context):
+        def get_context_data(c_type, context, payload_index):
```
hunterhector, line 1017: please add data type and docstring whenever possible

forte/data/data_store.py
```diff
@@ -1079,6 +1124,48 @@ def add_image_annotation_raw(
                 return tid_search_result
         return self._add_entry_raw(AudioAnnotation, type_name, entry)
 
+    def add_payload_raw(
+        self,
+        type_name: str,
+        payload_idx: int,
+        modality: str,
```
hunterhector, line 1131: I would recommend making `modality` as enum instead of general string types. `Enum` is [supported](https://stackoverflow.com/questions/36932/how-can-i-represent-an-enum-in-python) up to python 2.4 with the library installed.

forte/data/data_pack.py
mylibrar: If the replace operations are specific to one `TextPayload`, then I guess maybe we should bound them to that payload?

forte/data/base_pack.py
hepengfe: I think this is `get_single()` which is intended to get a single entry. In the past, it could only get the first one, here I make it to get kth one. Maybe I should change parameter instead. The index checking is used when there is no entry at all or the requested index is not available.

hunterhector: I feel like you are mixing up two different indices, what is the meaning of getting the kth item when k=payload index?

hunterhector: We shouldn't support getting `k`th item, we could support get the first item for a certain payload, but that's not what is currently implemented here.

forte/data/data_pack.py
hepengfe: Do you mean `image1 = pack.get_payload_at(1)`? I think this operation could be problematic as user needs to know the index of modality that he wants. Or do you think `image1 = pack.get_payload_at("image", 1)` is a good interface?

forte/data/base_pack.py
hunterhector: Or maybe you just name the parameter wrong? Why is `payload_index` even appear here?

forte/data/data_pack.py
hepengfe: They don't share the same indices. I have added the explanation.

forte/data/base_pack.py
```diff
@@ -468,22 +468,35 @@ def get(
         """
         raise NotImplementedError
 
-    def get_single(self, entry_type: Union[str, Type[EntryType]]) -> EntryType:
+    def get_single(
```
mylibrar, line 471: We can revert the changes here since `get_single` is no longer used in the current implementation.

forte/data/data_pack.py
```diff
@@ -244,19 +248,22 @@ def _validate(self, entry: EntryType) -> bool:
         return isinstance(entry, SinglePackEntries)
 
     @property
-    def text(self) -> str:
-        r"""Return the text of the data pack"""
-        return self._text
+    def text(self, text_payload_index: int = 0) -> str:
```
mylibrar, line 251: We can remove the argument `text_payload_index` here since there is no way to pass argument to a getter method decorated with `@property`.

forte/data/data_pack.py
```diff
@@ -453,19 +460,88 @@ def groups(self):
     def groups(self, val):
         self._groups = val
 
-    def get_span_text(self, begin: int, end: int) -> str:
+    def get_payload_at(
+        self, modality: str, payload_index: int
```
mylibrar, line 464: Maybe we can reuse the `Modality` enum defined in `top.py` instead of `string` for `modality`?

hunterhector: We probably would also move `Enum` out from the `top.py` file? On one hand, we can keep `top.py` mainly for ontology only, on the other, it would make importing these library-level Enums easier.

forte/data/ontology/top.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from dataclasses import dataclass
+from enum import IntEnum
```
hunterhector, line 15: Make sure you add a dependency to https://pypi.org/project/enum/, to we can keep backward compatibility.

forte/data/data_pack.py
hepengfe: OK. Should I return the text from all text payloads or just the first text payload? I am thinking maybe all text payloads?

forte/data/data_pack.py
hepengfe: created another file `forte/data/modality.py` to keep `Modality`

forte/data/data_pack.py
mylibrar: For backward compatibility, I think it's better to just return the text from the first text payload.

forte/data/base_pack.py
hepengfe: I think there are still many test cases dependent on this function. Even if it's no longer needed, maybe we can raise an issue for it and fix it in another PR.

hunterhector: no, not deleting `get_single`, but revert it to the previous state. This function is still needed.

forte/data/base_pack.py
```diff
@@ -481,9 +483,19 @@ def get_single(self, entry_type: Union[str, Type[EntryType]]) -> EntryType:
         Returns:
```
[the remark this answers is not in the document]
hepengfe: remove `payload_index` parameter

forte/data/ontology/top.py
hepengfe: it's a built-in python package

forte/data/ontology/top.py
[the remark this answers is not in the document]
hepengfe: I have made those variables private.

forte/data/ontology/top.py
hunterhector: Sorry I mean this link: https://pypi.org/project/enum34/, I think this `enum` is available after Py34

hunterhector: need to resolve this

tests/forte/data/audio_annotation_test.py
```diff
@@ -68,7 +81,8 @@ class TextUtteranceProcessor(PackProcessor):
     """
 
     def _process(self, input_pack: DataPack):
-        input_pack.set_text("test text")
```
hunterhector, line 71: This shouldn't need to be changed, the original interface needs to work.

tests/forte/data/audio_annotation_test.py
```diff
@@ -131,51 +144,75 @@ def setUp(self):
         self._pipeline.add(TextUtteranceProcessor())
         self._pipeline.initialize()
 
-
     def test_audio_annotation(self):
-
         # Test `DataPack.get_span_audio()` with None audio payload
-        with self.assertRaises(ProcessExecutionException):
```
hunterhector, line 138: I don't think this behavior needs to be changed in this PR.

tests/forte/data/audio_annotation_test.py
```diff
@@ -131,51 +144,75 @@ def setUp(self):
...
+        with self.assertRaises(ValueError):
             pack: DataPack = DataPack()
-            pack.set_text("test text")
```
hunterhector, line 140: Should keep all old interfaces work

ft/onto/base_ontology.py
```diff
@@ -555,6 +594,73 @@ class AudioUtterance(AudioAnnotation):
...
+    Attributes:
+        sample_rate (Optional[int]):
+    """
+
+    sample_rate: Optional[int]
+
+    def __init__(
+        self,
+        pack: DataPack,
+        modality: IntEnum,
```
hunterhector, line 615: We need to somehow remove the "modality" attribute here, otherwise, this interface is too redundant for the user.

ft/onto/base_ontology.py
```diff
@@ -308,7 +312,12 @@ class PredicateLink(Link):
     ParentType = PredicateMention
     ChildType = PredicateArgument
 
-    def __init__(self, pack: DataPack, parent: Optional[Entry] = None, child: Optional[Entry] = None):
+    def __init__(
```
hunterhector, line 315: why there are so many changes to the `base_ontology` file, it should be generated and all the rest should be untouched.

forte/processors/ir/bert/bert_based_query_creator.py
```diff
@@ -45,10 +45,10 @@ def __init__(self) -> None:
     def initialize(self, resources: Resources, configs: Config):
         self.resource = resources
         self.config = configs
-
-        self.device = torch.device(
-            "cuda" if torch.cuda.is_available() else "cpu"
-        )
+        self.device = "cpu"
```
hunterhector, line 48: why are we changing irrelevant files?

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
+    Raises:
+        ValueError: raised when the modality is not supported.
+    """
+
+    def __init__(
+        self,
+        pack: PackType,
+        modality: IntEnum,
+        payload_idx: int = 0,
+        uri: Optional[str] = None,
```
hunterhector, line 1195: I haven't seen url being used in the test cases

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
+    ):
+        supported_modality = ("text", "audio", "image")
```
hunterhector, line 1197: we have enum why still this

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
+                f"The given modality {modality} is not supported. "
+                f"Currently we only support {supported_modality}"
+            )
+        self._payload_idx: int = payload_idx
+        self._modality: IntEnum = modality
+        self._uri: Optional[str] = uri
+
+        super().__init__(pack)
+        # self._cache: Optional[Union[str, np.ndarray]] = None
+        self._cache: Union[str, np.ndarray] = ""
```
hunterhector, line 1209: default to empty string here is weird, probably `None`

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
+        self.meta: Optional[Generics] = None
+
+    def get_type(self) -> type:
+        """
+        Get the type of the payload class.
```
hunterhector, line 1214: This docstring provides no information. What is a "type" anyway

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
+            the modality of the payload class in str format.
+        """
+        return self._modality.name
+
+    @property
+    def cache(self) -> Union[str, np.ndarray]:
+        if self._cache is None:
+            raise ValueError(
+                "Payload doesn't have a cache."
+                "Please set the reader config `lazy_read` to False"
```
hunterhector, line 1235: wrong, Payload is not associated with a "reader"

forte/data/ontology/top.py
```diff
@@ -1168,12 +1172,125 @@ def __init__(
...
 
+class Payload(Entry):
+    """
+    A payload class that holds data cache of one modality and its data source uri.
+
+    Args:
+        pack: The container that this `Payload` will
+            be added to.
+        modality: modality of the payload such as text, audio and image.
+        payload_idx: the index of the payload.
```
hunterhector, line 1183: Should provide informative docstring. For example, `the index of the payload` does not add any value to `payload_idx`

forte/data/data_store.py
```diff
@@ -1120,6 +1159,48 @@ def add_grid_raw(
...
+        """
+        # We should create the `entry data` with the format
+        # [payload_idx, modality, tid, type_id, None, ...].
+        # A helper function _new_payload() can be used to generate a
+        # payload type entry data with default fields.
+        # A reference to the entry should be store in both self.__elements and
+        # self.__tid_ref_dict.
+        entry = self._new_payload(type_name, payload_idx, modality, tid)
+
+        if not allow_duplicate:
```
hunterhector, line 1197: There is no duplicate to handle for payload.

forte/data/data_pack.py
```diff
@@ -26,15 +26,16 @@
...
 )
 from functools import partial
+
 from typing_inspect import get_origin
 from packaging.version import Version
 import numpy as np
 from sortedcontainers import SortedList
 
 from forte.common.exception import (
-    ProcessExecutionException,
```
hunterhector, line 37: I don't think it is a good idea to change any existing behaviors, such as Exceptions thrown. If you want to change the exception, it should be a different PR.

forte/data/data_pack.py
```diff
@@ -1568,6 +1686,17 @@ def entry_setter(cls: Entry, value: Any, attr_name: str, field_type):
         # Record that this entry hasn't been added to the index yet.
         self._pending_entries[entry.tid] = entry.tid, c
 
+        if isinstance(entry, Payload):
+            if entry.get_modality() == "text":
```
hunterhector, line 1690: If you have enum you shouldn't use strings

forte/data/data_pack.py
```diff
@@ -245,18 +234,26 @@ def _validate(self, entry: EntryType) -> bool:
 
     @property
     def text(self) -> str:
-        r"""Return the text of the data pack"""
-        return self._text
+        """
+        Get text from a text payload at an index.
```
hunterhector, line 238: This docstring exposes the implementation detail (payload), and what is `index` here?

forte/data/data_pack.py
```diff
@@ -245,18 +234,26 @@ def _validate(self, entry: EntryType) -> bool:
...
-        return getattr(self._meta, "sample_rate")
+        Raises:
+            ValueError: raised when the index is out of bound of the text
+                payload list.
+
+        Returns:
+            text data in the text payload.
+        """
+        try:
+            tp = self.get_payload_at("text", 0)
```
hunterhector, line 251: you can simply return here

forte/data/data_pack.py
```diff
@@ -453,19 +450,90 @@ def groups(self):
...
+                in this DataPack's Payload entries of the requested modality.
+
+        Raises:
+            ValueError: raised when the requested modality is not supported.
+
+        Returns:
+            Payload entry containing text data or numpy array for image and
+            audio data.
+        """
+        supported_modality = ("text", "audio", "image")
```
hunterhector, line 471: enum

forte/data/data_pack.py
```diff
@@ -453,19 +450,90 @@ def groups(self):
...
+        Args:
+            modality: data modality among "text", "audio", "image"
+            payload_index: the zero-based index of the Payload
+                in this DataPack's Payload entries of the requested modality.
+
+        Raises:
+            ValueError: raised when the requested modality is not supported.
+
+        Returns:
+            str data for text data or numpy array for image and audio data.
```
hunterhector, line 512: Use a bullet list here

forte/data/data_pack.py
```diff
@@ -743,7 +843,7 @@ def __add_entry_with_check(self, entry: Union[EntryType, int]) -> EntryType:
                     f"is not a valid begin."
                 )
 
-            if end > len(self.text):
+            if end > len(self.get_payload_at("text", 0).cache):
```
hunterhector, line 846: enum

forte/data/ontology/top.py
```diff
@@ -870,7 +872,9 @@ def __init__(
         self._width = width
         self._image_payload_idx = image_payload_idx
         super().__init__(pack)
-        self.img_arr = self.pack.get_image_array(self._image_payload_idx)
+        self.img_arr = self.pack.get_payload_data_at(
+            "image", self._image_payload_idx
```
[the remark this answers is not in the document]
hepengfe: I was thinking from user side they don't need to initialize an Enum object, and we only use Enum object internally.

forte/data/ontology/top.py
hepengfe: For backward compatibility, there are many cases `TextPayload.cache` is expected to be an empty string.

tests/forte/data/readers/audio_reader_test.py
```diff
@@ -14,9 +14,13 @@
...
 """
+from email.mime import audio
+import importlib
 import os
+from sunau import AUDIO_FILE_ENCODING_ADPCM_G721
 import unittest
 from typing import Dict
+from forte.data import Modality
 from torch import argmax
 from transformers import Wav2Vec2Processor, Wav2Vec2ForCTC
```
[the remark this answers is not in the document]
hepengfe: Modality is needed here.

forte/data/readers/audio_reader.py
```diff
@@ -86,4 +94,4 @@ def default_configs(cls):
 
         Returns: The default configuration of audio reader.
         """
-        return {"file_ext": ".flac", "read_kwargs": None}
+        return {"file_ext": ".flac", "lazy_read": False, "read_kwargs": None}
```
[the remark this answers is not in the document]
hepengfe: https://github.com/asyml/forte/issues/841 it's for next PR.

forte/data/base_pack.py
```diff
@@ -149,6 +149,7 @@ def __setstate__(self, state):
                 f"version greater or equal to {PACK_ID_COMPATIBLE_VERSION}"
             )
         super().__setstate__(state)
+
```
mylibrar, line 152: We should revert all changes in irrelevant files. (line152 & line656)

forte/data/data_pack.py
```diff
@@ -932,16 +1050,24 @@ def get_annotation_list(
                     " [Annotation, AudioAnnotation]."
                 )
 
-        def get_context_data(c_type, context):
-            r"""Get context-specific data of a given context type and
-                context.
+        def get_context_data(
+            c_type: Union[Type[Annotation], Type[AudioAnnotation]],
+            context: Union[Annotation, AudioAnnotation],
+            payload_index: int,
```
mylibrar, line 1056: It should default to `0` according to the docstring below

forte/data/readers/audio_reader.py
```diff
@@ -53,21 +54,27 @@ def _collect(self, audio_directory) -> Iterator[Any]:  # type: ignore
...
+            self.configs.file_ext,
+        )
 
     def _cache_key_function(self, audio_file: str) -> str:
         return os.path.basename(audio_file)
 
     def _parse_pack(self, file_path: str) -> Iterator[DataPack]:
         pack: DataPack = DataPack()
-
+        payload_idx = 0
```
mylibrar, line 68: I think we don't need to change the `AudioReader` in this PR since the design of lazy loading is not complete. You can try to address this issue in the PR to #841

forte/processors/ir/bert/bert_based_query_creator.py
```diff
@@ -45,7 +45,6 @@ def __init__(self) -> None:
     def initialize(self, resources: Resources, configs: Config):
         self.resource = resources
         self.config = configs
-
```
mylibrar, line 48: Let's remove changes in irrelevant files.

tests/forte/data/audio_annotation_test.py
```diff
@@ -27,18 +28,30 @@
...
     """
     A processor to add a Recording ontology to the whole audio data.
     """
+
     def _process(self, input_pack: DataPack):
         Recording(
-            pack=input_pack, begin=0, end=len(input_pack.audio)
+            pack=input_pack,
+            begin=0,
+            end=len(input_pack.get_payload_data_at(Modality.Audio, 0)),
```
mylibrar, line 54: This change might also be unnecessary since `input_pack.audio` is doing the same thing here.

tests/forte/data/audio_annotation_test.py
```diff
@@ -143,39 +153,65 @@ def test_audio_annotation(self):
...
+
             # Check Recording
             recordings = list(pack.get(Recording))
             self.assertEqual(len(recordings), 1)
-            self.assertTrue(array_equal(recordings[0].audio, pack.audio))
-
+            self.assertTrue(
+                array_equal(
+                    recordings[0].audio,
+                    pack.get_payload_data_at(Modality.Audio, 0),
```
mylibrar, line 212: The same here. If the audio_payload_index is `0`, we can just use `pack.audio`

tests/forte/data/audio_annotation_test.py
```diff
@@ -192,19 +228,24 @@ def test_audio_annotation(self):
...
                 for audio_utter in audio_utters:
                     configs: Dict = self._test_configs[audio_utter.speaker]
-                    self.assertTrue(array_equal(
-                        audio_utter.audio,
-                        pack.audio[configs["begin"]:configs["end"]]
-                    ))
+                    self.assertTrue(
+                        array_equal(
+                            audio_utter.audio,
+                            pack.get_payload_data_at(Modality.Audio, 0)[
```
mylibrar, line 234: `pack.audio` is enough. In PR to #841, you can try to add unit tests which involve multiple audio payloads.

tests/forte/data/readers/audio_reader_test.py
```diff
@@ -42,23 +45,27 @@ def initialize(self, resources: Resources, configs: Config):
...
-            input_pack.audio, return_tensors="pt", padding="longest"
+            audio_data, return_tensors="pt", padding="longest"
         ).input_values  # Batch size 1
 
         # take argmax and decode
         transcription = self._tokenizer.batch_decode(
             argmax(self._model(input_values).logits, dim=-1)
         )
 
+        tp = TextPayload(input_pack, 0)
```
mylibrar, line 68: This line might be redundant since `input_pack.set_text` will automatically create a `TextPayload`.

tests/forte/data/audio_annotation_test.py
hepengfe: I think Hector mentioned that we want to remove `audio` and `text` property in datapack. In this PR I have already removed `audio` property. Since there are too many code dependent on `text` property, we want to remove it in a separate PR.

tests/forte/data/audio_annotation_test.py
hepengfe: `pack.audio` is removed in this PR.

tests/forte/data/audio_annotation_test.py
hepengfe: `pack.audio` is removed in this PR.

mylibrar (approved):

hunterhector:
I remember clearly I requested to install `enum` package (https://pypi.org/project/enum34/) for compatibility issues, but this is not addressed and ignored. What happened?

forte/data/ontology/top.py
```diff
@@ -1207,12 +1229,149 @@ def __init__(
...
+    ):
+        from ft.onto.base_ontology import (  # pylint: disable=import-outside-toplevel
+            TextPayload,
+            AudioPayload,
+            ImagePayload,
+        )
+
+        # since we cannot pass different modality from generated ontology, and
+        # we don't want to import base ontology in the header of the file
+        # we import it here.
```
hunterhector, line 1264: This is not a solution. We cannot import items that are generated back to the source code. This doesn't make sense.

Issue #829, opened 2022-06-08
Doctest for README


hepengfe:
**Describe the bug**
Currently we don't have a test for readme, this prevent us from making sure readme code is bug-free after modifications.
Therefore, we want to add `doctest` to test README file.

**To Reproduce**
1. Prepare data `...`
2. Run command '....'
3. See error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Python and Package verions: [e.g. Python version, Pytorch version]

**Additional context**

hepengfe:
Also, I think the previous fix is partially wrong. https://github.com/asyml/forte/pull/823 
Although it fixed variable `token_entries` not found, since it iterates `token_entries` which is a generator twice. It actually iterate nothing in the second run which is `for token, tag in zip(input_pack.get(Token), taggings)` and it results in no tag is assigned.

PR #830, opened 2022-06-08
Doc test readme

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/829. 

### Description of changes
* change `TerminalReader` to `StringReader`
* fixed the issue https://github.com/asyml/forte/issues/829#issuecomment-1150472115
* add package `mkcodes` that convert markdown code into python script
* generate python script based on latest README, test it and remove it after testing 

### Test Conducted

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,6 @@ pip install forte.spacy
 Let's start by writing a simple processor that analyze POS tags to tokens using the good old NLTK library.
 ```python
 import nltk
-
 from forte.processors.base import PackProcessor
 from forte.data.data_pack import DataPack
 from ft.onto.base_ontology import Token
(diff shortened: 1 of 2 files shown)


mylibrar (dismissed):

hunterhector:
I find [markdown autodoc](https://stackoverflow.com/questions/10280902/how-to-embed-code-from-repo-in-github-markdown/70584456#70584456 ) to be a better alternative. Instead of extracting code from a markdown (which is a bit hacky), we just need to add a tested piece of code into the README.

README.md
```diff
@@ -87,34 +87,8 @@ pip install forte.spacy
...
-        token_texts = [token.text for token in token_entries]
-
-        # use nltk pos tagging module to tag token texts
-        taggings = nltk.pos_tag(token_texts)
-
-        # assign nltk taggings to token attributes
-        for token, tag in zip(token_entries, taggings):
-            token.pos = tag[1]
-```
+<!-- MARKDOWN-AUTO-DOCS:START (CODE:src=tests/forte/readme_test.py=28-52) -->
```
hunterhector, line 90: btw, if this succeeds eventually, it'd be better if we don't use the line numbers here (because changing the test case itself will render the readme.). I would suggest putting both code snippets in their individual files. And the test cases should be on their own files.

hepengfe: is it like two versions of readme code? One for embedding and the other for testing.

hunterhector: I also realize we cannot avoid using line numbers. But it would be still better to separate code snippets from the test code (we can actually do unit tests in the test file).

hepengfe: - docs
   - readme_code
     - read_me_snippet_1.py
     - ...

- tests
   - readme_code
      - test_read_me_snippet_1.py
      - ...


Or keep all files under `docs/readme_code`

hunterhector: The structure looks good.

Btw, this PR passes the CI but we cannot check the doc build. Also, `readthedoc` rendering doesn't work since it starts before the Github action.

mylibrar (approved):

PR #844, opened 2022-06-29
Integrating CV Ontologies with the DataStore format

### LINKED ISSUE #840: Integrate DataStore with CV ontologies
* We should integrate all the top-level `Entry`s to `DataStore`. Currently the following ontologies have not beed integrated yet:
  * ImageAnnotation
  * Region
  * Box
  * BoundingBox
* We also need to formalize the steps to add a new top level entry. A more details [guide](https://github.com/asyml/forte/blob/330a66711c65d25865cca2ccd1ecd22e0a99128c/forte/data/ontology/top.py#L58) should be created for developers to follow when they want to register a new entry in `top.py`.

**Depends on**
- #837 

* To migrate a top-level Entry, we will need to
  * Design a schema for it in `DataStore`
    * The schema should only consists of primitive types or built-in list/dict. We want it to be serializable.
    * For example, a `Group` type should store its `<member_type>` and `<[members_tid_list]>` as [the first two attributes](https://github.com/asyml/forte/blob/330a66711c65d25865cca2ccd1ecd22e0a99128c/forte/data/data_store.py#L741) in `DataStore`.
  * Direct all the update operations to `DataStore`
    * When we call `Group.add_member(entry)`, [under the hood](https://github.com/asyml/forte/blob/330a66711c65d25865cca2ccd1ecd22e0a99128c/forte/data/ontology/top.py#L405) it should append `entry.tid` to the member list stored in `DataStore`.

It's better to include some dedicated test cases to validate any corner cases.

**Additional context**

Pushkar-Bhuse:
This PR fixes #840 . 

### Description of changes
This PR mainly includes fine-tuning CV ontologies to be used correctly with the `DataStore` setup. Once this was completed, we included test cases (both from the users point of view and also from the `DataStore`'s perspective) to validate the working of these ontologies.

After this PR, DataStore can also include CV ontologies.

### Test Conducted
Apart from data store specific functions, we tested `addition`, `deletion` and `updating` annotations, `compute_iou` and `check_overlap` method. These tests were conducted on the following ontologies
-  `ImageAnnotation`
- `Region`
-  `Box`
- `BoundingBox`


tests/forte/image_annotation_test.py
```diff
@@ -30,21 +31,208 @@ class ImageAnnotationTest(unittest.TestCase):
...
+        # Entries for another image in the same Data Pack
+        self.bb4 = BoundingBox(
+            pack=self.datapack_1,
+            height=3,
+            width=4,
+            grid_height=6,
+            grid_width=6,
+            grid_cell_h_idx=2,
+            grid_cell_w_idx=2,
+            image_payload_idx=1,
```
mylibrar, line 86: You can add a test to see this `image_payload_idx=1` is actually stored in `self.datapack_1._data_store.get_entry(tid=self.bb4.tid)`. I guess it might still be `0`

tests/forte/image_annotation_test.py
```diff
@@ -30,21 +31,208 @@ class ImageAnnotationTest(unittest.TestCase):
...
+        # Change a parameter of the entry object
+        self.bb1._height = 5
+
+        # Fetch attribute value from data store
+        bb1_height = self.datapack_1._data_store.get_attribute(
+            self.bb1.tid, "_height"
+        )
+        # Check new value
+        self.assertEqual(bb1_height, 5)
+
```
mylibrar, line 193: You can also add test cases to update `BoundingBox.image_payload_idx` and see if that change is reflected in `DataStore`

tests/forte/data/data_store_test.py
```diff
@@ -142,42 +143,74 @@ def setUp(self) -> None:
...
+                    "_cx": 5,
+                    "_height": 6,
+                    "_width": 7,
+                    "_grid_id": 8,
+                },
+                "parent_class": set(),
+            },
         }
 
         self.base_type_attributes = {
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Yes that's true. I think the only reason it's there is to help readers understand the underlying structure of the `DataStore` at the beginning. I can remove it if you want.

forte/data/ontology/top.py
```diff
@@ -1057,13 +1088,18 @@ class Box(Region):
...
 
+    _cy: int
+    _cx: int
+    _height: int
+    _width: int
+
     def __init__(
         self,
         pack: PackType,
         cy: int,
```
mylibrar, line 1099: Will this cause a problem at this [line](https://github.com/asyml/forte/blob/a9998d6acb1ff9b4b01c7e90e8c8bd79fbae1ec3/forte/data/entry_converter.py#L202)? Seems like these arguments are required to instantiate a Box object.

Pushkar-Bhuse: Yes you're right. Sorry I missed that.

forte/data/ontology/top.py
```diff
@@ -1208,15 +1247,17 @@ class BoundingBox(Box):
 
     """
 
+    _grid_id: int
```
hunterhector, line 1250: here things become hacky. Why are we having these as `protected`?

forte/data/ontology/top.py
```diff
@@ -1078,13 +1111,18 @@ class Box(Region):
         width: the width of the box, the unit is one image array entry.
     """
 
+    _cy: int
```
hunterhector, line 1114: These should be the public names just like other entries.

tests/forte/data/data_store_test.py
```diff
@@ -142,42 +143,51 @@ def setUp(self) -> None:
...
-            'forte.data.ontology.top.AudioAnnotation': {'parent_class': {'Entry'}}
-        }
-
-        DataStore._type_attributes["ft.onto.base_ontology.Document"] = {
+            "forte.data.ontology.top.BoundingBox": {
                 "attributes": {
-                    "document_class": 4,
-                    "sentiment": 5,
-                    "classifications": 6,
+                    "_cy": 4,
```
hunterhector, line 148: One reason that we don't want these to be `protected` is also because of the serialization. We don't want the serialization format to have these extra `_`

forte/data/ontology/top.py
```diff
@@ -1208,15 +1247,17 @@ class BoundingBox(Box):
...
     def __init__(
         self,
         pack: PackType,
-        height: int,
-        width: int,
-        grid_height: int,
-        grid_width: int,
-        grid_cell_h_idx: int,
-        grid_cell_w_idx: int,
+        height: int = 1,
```
hunterhector, line 1255: Why only `grid` move to the class level? Note that it is not just about moving things to data store, but need to have a consistent approach too.

forte/data/ontology/top.py
```diff
@@ -1078,13 +1111,18 @@ class Box(Region):
...
         self,
         pack: PackType,
-        cy: int,
-        cx: int,
-        height: int,
-        width: int,
+        cy: int = 0,
+        cx: int = 0,
+        height: int = 1,
+        width: int = 1,
```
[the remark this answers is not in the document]
Pushkar-Bhuse: So the reason we had defaults was that during converting `DataStore` entries to objects (and when the entry is not present in `DataIndex`, we reach [here](https://github.com/Pushkar-Bhuse/forte/blob/76e16664002c50d91d7214eb6a6f553603a93f4f/forte/data/entry_converter.py#L212). Since we only pass the pack, it throws an error. That was the reason we had added default values earlier. But I'll discuss a better fix for this with @hepengfe.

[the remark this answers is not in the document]
hunterhector: I will leave it to you to study the resolution. But please note that the principle should be "API drives the implementation", it can never be "the implementation drives the API".

On this issue, @mylibrar can probably help you more.

forte/data/ontology/top.py
Pushkar-Bhuse: I believe that was the original implementation of `BoundingBox`. Do you feel that we should not create the `Grid` object?

forte/data/ontology/top.py
```diff
@@ -1052,12 +1062,35 @@ def __init__(self, pack: PackType, image_payload_idx: int = 0):
...
 
+    @property
+    def image_payload_idx(self):
+        r"""Getter function of ``image_payload_idx``. The function will first try to
+        retrieve the image_payload_idx index from ``DataStore`` in ``self.pack``. If
+        this attempt fails, it will directly return the value in ``_image_payload_idx``.
+        """
+        try:
+            self._image_payload_idx = self.pack.get_entry_raw(self.tid)[
+                PAYLOAD_INDEX
```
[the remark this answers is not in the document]
hunterhector: Let's just try not to add additional consts here. DataStore has a system to add attributes and datastore can manage the attribute position. Can we simply consider `Payload_index` to be a regular attribute?

forte/data/ontology/top.py
hunterhector: No, I mean you moved `_grid_id` to be a class variable. Do you understand how these class variable affect the attribute list?

I don't know what you mean by creeating the `Grid` object, but that should not be a problem for this PR.

forte/data/ontology/top.py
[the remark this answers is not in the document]
Pushkar-Bhuse: Yes. Every `Entry` has two main attributes which occupy the first and second position in the datastore (for Annotation it is begin and end index). Similar, for `ImageAnnotation` and its subclasses, the two main attributes are payload_idx and None (Implying that it only has one main attribute, the extra None is just for consistency). That is the reason I stored the Payload index value in `constants`

forte/data/ontology/top.py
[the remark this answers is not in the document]
Pushkar-Bhuse: Yes so I discussed this with @mylibrar. The solution we came up with was - When converting a datastore to object, the default values are just a placeholder and they are immediately replaced by values stored in the DataStore entry. Also, this case is very specific to `Box` and `BoundingBox` type entries. Thus, we can have separate `elif` branches before [this](https://github.com/Pushkar-Bhuse/forte/blob/76e16664002c50d91d7214eb6a6f553603a93f4f/forte/data/entry_converter.py#L208) to just pass dummy values to `Box` and `BoundingBox` (instead of setting default parameter values). These dummy values will then be overwritten by the values stored in the DataStore entry.

forte/data/ontology/top.py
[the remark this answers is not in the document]
hunterhector: what is the `None` here?

forte/data/ontology/top.py
Pushkar-Bhuse: As you suggested, I am using the new implementation of CV ontologies (from @hepengfe 's branch) and making my changes on that so that merging is easier. I believe he has handled the Grid implementation there.

forte/data/ontology/top.py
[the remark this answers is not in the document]
hunterhector: The solution is another issue, but the interface is important, we cannot fit the interface to implementation. So if the default values don't make sense, we can't have them.

forte/data/ontology/top.py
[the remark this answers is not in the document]
hunterhector: > So Payload Index is the only main attribute for `ImageAnnotation` type entries (just like how `Annotation` has `start` and `end` indices, `payload_index` is stored in the first two indices of the `DataStore` entry. That's why I felt that could be the only constant we could add for `ImageAnnotation`

Ok, so I think the reason I have the confusion is that: once `Payload` is in our system, then all other payload-dependent entries (specifically, Annotation, ImageAnnotation, AudioAnnotation) should have `Payload` associated with it too. Now we only modify it for `ImageAnnotation` in this PR, it feels a bit inconsistent to me.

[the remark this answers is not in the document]
hunterhector: 1. So we should make all the three have a `Payload` field. I am not sure of what `data class attribute` means yet, but we need to make all of the consistent, but with two considerations:
    i. maybe it will make this PR too large?
    ii.  we don't want to change any existing behavior, i.e., previously we don't have multiple payloads for each modality, thus all entries are retrieved without any payload index. To keep the same behavior, we should assume any data operation that does not use a `payload index` default to 0, i.e., get from the first payload or add to the first payload when payload index is not specified.
2. Can you remind me what are `two main attributes`? I just wonder why it could be `None`.

forte/data/ontology/top.py
hunterhector: ok, probably this PR is a bit intertwined with the `Grid` which makes me hard to judge the correctness right now.

mylibrar (commented):
#827 is closed so the implementation should no longer be based on that PR. You can consider reverting or rebasing some of the commits.

Issue #691, opened 2022-03-22
Convert examples to notebook tutorial


feipenghe:
**Is your feature request related to a problem? Please describe.**
We want examples under `examples` folder to be written as jupyter notebooks as it will be easier to learn for new user and can also be rendered in sphinx documentation.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

* [ ] [audio](https://github.com/asyml/forte/tree/master/examples/audio)
* [ ] [blog_post_examples](https://github.com/asyml/forte/tree/master/examples/blog_post_examples)
* [ ] [chatbot](https://github.com/asyml/forte/tree/master/examples/chatbot)
* [ ] [classification](https://github.com/asyml/forte/tree/master/examples/classification)
* [ ] [clinical_pipeline](https://github.com/asyml/forte/tree/master/examples/clinical_pipeline)
* [ ] [content_rewriter](https://github.com/asyml/forte/tree/master/examples/content_rewriter)
* [ ] [data_augmentation](https://github.com/asyml/forte/tree/master/examples/data_augmentation)
* [ ] [ner](https://github.com/asyml/forte/tree/master/examples/ner)
* [ ] [ontology](https://github.com/asyml/forte/tree/master/examples/ontology)
* [ ] [passage_ranker](https://github.com/asyml/forte/tree/master/examples/passage_ranker)
* [ ] [pipelines](https://github.com/asyml/forte/tree/master/examples/pipelines)
* [ ] [serialization](https://github.com/asyml/forte/tree/master/examples/serialization)
* [ ] [srl](https://github.com/asyml/forte/tree/master/examples/srl)
* [ ] [tagging](https://github.com/asyml/forte/tree/master/examples/tagging)
* [ ] [twitter_sentiment_analysis](https://github.com/asyml/forte/tree/master/examples/twitter_sentiment_analysis)
* [ ] [visualize](https://github.com/asyml/forte/tree/master/examples/visualize)
[wiki_parser](https://github.com/asyml/forte/tree/master/examples/wiki_parser)

PR #846, opened 2022-07-01
Audio tutorial

### LINKED ISSUE #850: Jupyter Notebook Tutorial for Automatic Speech recognition
This issue is part of #691 

Write a jupyter notebook tutorial for automatic speech translation based on Forte examples audio.py

**Additional context**

bhaskar2443053:
This issue is part of #691 
This PR fixes #850 

Description of changes
This tutorial is an introduction to the Automatic speech translation using Forte


hunterhector:
please add an issue to track this pr, thanks

hepengfe:
Overall it's a good tutorial, and I don't have much input. I can approve it after the misspelling is fixed and the issue is linked.

hunterhector:
The file is too large and we should keep the master lean. We should move the assets (images) to this branch (https://github.com/asyml/forte/tree/assets)

bhaskar2443053:
> please add an issue to track this pr, thanks

done

bhaskar2443053:
> Overall it's a good tutorial, and I don't have much input. I can approve it after the misspelling is fixed and the issue is linked.

fixed

bhaskar2443053:
> The file is too large and we should keep the master lean. We should move the assets (images) to this branch (https://github.com/asyml/forte/tree/assets)

image referred from URL, to keep master lean

hunterhector (commented):
A couple suggestions:
1. The notebook file is still 2.7 MB (after image removal), if this is plain text it shouldn't be that large, maybe there is some output that should be removed too? The size should be a couple of k.
2. When asking users to click and listening to audio, probably should specify what's required (for example, it won't work if the notebook is run on a server without audio output).
3. Some explanation needs to be made to the Forte processors, for example, what are the "AudioUtterances", why are we adding the "Link"s. Eventually, this is a `Forte` tutorial more than ASR, we need to teach more about Forte and how it helps with managing the data. This is the main issue of the tutorial.
4. The part that plays `seg[1]` to `seg[3]` is a bit repetitive.

bhaskar2443053:
> A couple suggestions:
> 
> 1. The notebook file is still 2.7 MB (after image removal), if this is plain text it shouldn't be that large, maybe there is some output that should be removed too? The size should be a couple of k.

reduced the size to 184k, (removed the almost all audio play outputs)

> 2. When asking users to click and listening to audio, probably should specify what's required (for example, it won't work if the notebook is run on a server without audio output).

mostly it will use the user's laptop or PC to play the audio, still, I have mentioned the requirement.

> 3. Some explanation needs to be made to the Forte processors, for example, what are the "AudioUtterances", why are we adding the "Link"s. Eventually, this is a `Forte` tutorial more than ASR, we need to teach more about Forte and how it helps with managing the data. This is the main issue of the tutorial.

added context on "AudioUtterances" and "link". and why we are using it.

> 4. The part that plays `seg[1]` to `seg[3]` is a bit repetitive.

Now just playing one segment only.

hunterhector:
Let’s add test cases here: https://github.com/asyml/forte/tree/master/tests/forte/notebooks

bhaskar2443053:
> Let’s add test cases here: https://github.com/asyml/forte/tree/master/tests/forte/notebooks

done.

forte/data/types.py
```diff
@@ -24,9 +24,4 @@
 
 DataRequest = Dict[Type[Entry], Union[Dict, List]]
 
-try:
-    import torch
-
-    MatrixLike = Union[torch.TensorType, np.ndarray, List]
-except ImportError:
-    MatrixLike = Union[np.ndarray, List]  # type: ignore
+MatrixLike = Union[np.ndarray, List]
```
hunterhector, line 27: Hmm, why is this file changed? Hopefully a tutorial won't change anything in the library itself.

requirements.txt
```diff
@@ -13,4 +13,4 @@ dataclasses~=0.7; python_version <'3.7'
 importlib-resources==5.1.4;python_version<'3.7'
 packaging~=21.2
 asyml-utilities
-enum34
+enum34
```
hunterhector, line 16: If this is needed in `requirements` then it should be needed in `setup.py` too.

forte/data/types.py
hunterhector: Does this break this PR? This issue should probably be resolved in another PR with more investigation, not piggybacking this PR.

The current fix is not desired. We want to make MatrixLike to include pytorch tensor types if torch is available. We cannot just comment it out.

bhaskar2443053: I have added the code back. And it doesn't break the PR this time. so we can ignore this file change now.

hunterhector: ok, shall we create an issue describing what you think it's wrong?

hunterhector (approved):

PR #849, opened 2022-07-01
Updated  Range Annotation Iteration

### LINKED ISSUE #806: Create efficient range annotation iteration
The implementation of `get` with `range_annotation` is [currently inefficient](https://github.com/asyml/forte/pull/769#discussion_r880810365). This is because we actually iterate all annotations, and filter later. A better implementation could be that we search the first and last item, and only iterate in this range, this is similar to the old [data pack implementation](https://github.com/asyml/forte/blob/a954abf2c8122860c4551c1bb80a439004d05c09/forte/data/data_pack.py#L1294).   

As described above.

**Additional context**

Pushkar-Bhuse:
This PR fixes #806  

### Description of changes
In this PR, we update the `get` function in `DataStore` to use a more efficient way (using the `get_bisect_range` function from an [old `DataPack` implementation](https://github.com/asyml/forte/blob/a954abf2c8122860c4551c1bb80a439004d05c09/forte/data/data_pack.py#L1294)) 

This PR will provide a faster way to fetch all `Annotation` type entries from a given span.


forte/data/data_store.py
```diff
@@ -1365,9 +1408,10 @@ def within_range(
             if range_annotation is None:
                 yield from self.co_iterator_annotation_like(all_types)
             else:
-                for entry in self.co_iterator_annotation_like(all_types):
-                    if within_range(entry, range_annotation):
-                        yield entry
+                for entry_type in all_types:
+                    yield from get_bisect_range(
+                        entry_type, self.__elements[entry_type]
+                    )
```
hunterhector, line 1414: This actually changes the behavior since `co_iterator` is not used.

hunterhector: In addition, consider adding/reusing test cases that work directly from the data pack where you can actually ingest some real data. The DataStore test cases actually use some sample data created manually, which might be inefficient.

Pushkar-Bhuse: Actually I wrote this loop after looking at the working of `co_iterator` and tried to make it the same. Could you point me to where it might be different?

hunterhector: I think `co-iterator` will yield the annotations based on the position, entry type is the secondary tie-breaker. But the new code will yield based on the types first, and then the position.

hunterhector: In addition, we don't want the loop logic to be in two places (inside `co_iterator` and here). Instead, we should send `range_annotatoin` into `co_iterator` and handle it inside.

Pushkar-Bhuse: Okay I get your point. I'll look into that. Also, for the test cases. I looked at a few existing test cases like [`predictor_test.py`](https://github.com/asyml/forte/blob/master/tests/forte/processors/predictor_test.py) where we ingest real data. The new methods works fine in that as well.

hunterhector: 1. Try to see if some data digestion steps are not in `predictor_test.py`, hopefully in `data_pack_test.py`. Since it will be easier to locate the correct test cases. `predictor_test` is not designed for this.
2. The fact it worked now (passing the tests) is a problem. Since this implementation is likely to change the behavior but not cached, we need a test case where you can see the difference between these two implementations

forte/data/data_store.py
```diff
@@ -1212,11 +1215,58 @@ def co_iterator_annotation_like(
...
+                else search_list[-1][constants.END_INDEX]
+            )
+
+            base_class = (
+                Annotation
+                if self._is_subclass(entry_class, Annotation)
+                else AudioAnnotation
+            )
+
+            temp_begin = self.add_entry_raw(
```
mylibrar, line 1252: `SortedList` can perform binary searching based on [sorting function](https://github.com/asyml/forte/blob/62b081ade30b9ab478d15b70954d5fead2029aab/forte/data/data_store.py#L764), so maybe we don't even need to create these temporary entries for index searching.

forte/data/data_store.py
```diff
@@ -1212,11 +1215,40 @@ def co_iterator_annotation_like(
...
 
         Returns:
             An iterator of entry elements.
         """
 
+        def get_bisect_range(search_list: SortedList):
+            """
+            Perform binary search on the specified list for target entry class.
+            Args:
+                entry_class: Name of the target type of entry. It can be Annotation or
```
hunterhector, line 1230: subtype of `Annotation` or `AudioAnnotation`.

Btw, to make this docstring actually useful, we should refer these classes with the `:class:` sphinx syntax

forte/data/data_store.py
```diff
@@ -1226,18 +1258,29 @@ def co_iterator_annotation_like(
...
         # it avoids empty entry lists or non-existent entry list
         first_entries = []
+        all_entries = {}
 
-        for tn in type_names:
+        if range_annotation:
+            for entry_type in type_names:
+                all_entries[entry_type] = get_bisect_range(
+                    self.__elements[entry_type]
+                )
```
hunterhector, line 1267: I think this is inefficient here: `co_iterator` is designed to be efficient,  it has an `O(KN)` complexity, where K is the number of types, N is the length of the list.  Here, when we do `get_bisect_range` we don't have to actually do the list slicing operation `search_list[begin_index:end_index]`, we simply need to return `begin_index` and `end_index`, and then do the iteration within this range, to avoid this extra slicing.

forte/data/data_store.py
```diff
@@ -1282,23 +1325,23 @@ def co_iterator_annotation_like(
...
             # retrieve the popped entry tuple (minimum item in the heap)
-            # and get the p_idx (the index of the current entry list in self.__elements)
+            # and get the p_idx (the index of the current entry list in all_entries)
             entry_tuple = heappop(h)
             (_, _, p_idx), type_name = entry_tuple
             # get the index of current entry
             # and locate the entry represented by the tuple for yielding
             pointer = pointers[p_idx]
-            entry = self.__elements[type_name][pointer]
+            entry = all_entries[type_name][pointer]
```
hunterhector, line 1338: As suggested above, you can adjust the `pointer` here to the `begin_index`, and the line below to be `pointer + 1 < end_index`. Of course, you need to maintain different `begin_index` and `end_index` for all types.

forte/data/data_store.py
```diff
@@ -1212,11 +1215,40 @@ def co_iterator_annotation_like(
 
         Args:
             type_names: a list of string type names
+            range_annotation: a list with two elements,
```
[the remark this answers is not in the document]
hunterhector: After reading the code, `range_annotation` is not a list of two elements, but the raw Annotation entry data, right? It is fine to pass in the begin and end directly to this function.

tests/forte/data/data_pack_test.py
```diff
@@ -170,6 +172,59 @@ def test_get_entries(self):
...
+        for doc in self.data_pack.get(Document):
+            for anno in self.data_pack.get(Annotation, doc):
+                # Just to make the test case concise,
+                # We will only consider Sentence, EntityMention
+                # and PredicateArgument
+                if any(
+                    isinstance(anno, check)
+                    for check in [Sentence, EntityMention, PredicateArgument]
+                ):
+                    all_anno.append(anno.text)
```
hunterhector, line 186: In order for this test to be readable, I suggest we do `all_anno.append(check.__name__, anno.text, anno.begin, anno.end)`, we can actually see whether the return list complies to the order we expect.

tests/forte/data/data_store_test.py
```diff
@@ -980,7 +990,7 @@ def test_get(self):
                 "forte.data.ontology.top.Annotation", range_annotation=(1, 20)
             )
         )
-        self.assertEqual(len(instances), 2)
+        self.assertEqual(len(instances), 3)
```
hunterhector, line 993: could you remind me why this len changes?

forte/data/data_store.py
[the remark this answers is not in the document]
Pushkar-Bhuse: Yes so `range_annotation` is just a tuple of two ints. I can pass in the two values separately in the `co_iterator_annotation_like` function

forte/data/data_store.py
```diff
@@ -1184,7 +1183,9 @@ def get_length(self, type_name: str) -> int:
             return len(self.__elements[type_name]) - delete_count
 
     def co_iterator_annotation_like(
-        self, type_names: List[str]
+        self,
+        type_names: List[str],
+        range_annotation: Optional[Tuple[int]] = None,
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Yes

tests/forte/data/data_store_test.py
Pushkar-Bhuse: I believe this was because of an extra `Annotation` entry that was added [here](https://github.com/Pushkar-Bhuse/forte/blob/93d198e4a25fceb8f2ecd7c1ccb5fad09c23db74/tests/forte/data/data_store_test.py#L233) as a part of the test cases

forte/data/data_store.py
Pushkar-Bhuse: Got it!

forte/data/data_store.py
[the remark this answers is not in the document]
hunterhector: So a tuple of two `int` feels the same as two `int` parameters to me. But my original comment is "whether the input is the raw Annotation data", which is a list of more than two `int`s, but also include other attributes an Annotation has.

The name `range_annotation` implies that it is an `Annotation` object, which is not only a tuple containing the begin and end.

forte/data/data_store.py
```diff
@@ -1227,17 +1298,45 @@ def co_iterator_annotation_like(
...
-        for tn in type_names:
+        # For every entry type, store the index of the first and last entry
+        # of that type that needs to be fetched. When range_end and range_begin
+        # are None, we fetch all entries of each type (mentioned in type_names).
+        # But when range_end and range_end is specified, we find the first and
+        # last entry of a given type that needs to fetched and only iterate
+        # between them
+        all_entries_range = {}
+
+        # This list stores the types of entries that have atleast one entry to
```
hunterhector, line 1309:
```suggestion
        # This list stores the types of entries that have at least one entry to
```

tests/forte/data/data_store_test.py
hunterhector: what I don't understand is there is no change of the test case in this PR, but there is an assertion change, which makes me feel like the implementation is the reason

hunterhector (commented):
The new docstring looks pretty good to me and the logic seems to be correct. But I would recommend @hepengfe to take a look since he is the author of the original logic.

forte/data/data_store.py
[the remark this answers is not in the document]
Pushkar-Bhuse: Oh yes I understand. So `range_annotation` is just a `tuple` of two `int`. If you want we can make the name something more understandable.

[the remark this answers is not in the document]
hunterhector: sure, feel free to do so, you can convert this back to a tuple if you want. I am probably confused previously since some function arguments type annotations are also not ready yet.

tests/forte/data/data_store_test.py
hepengfe: I think there should be only two instances as only `ref2` and `ref5` are in the range `(1,20)`

forte/data/data_store.py
```diff
@@ -1203,21 +1205,91 @@ def co_iterator_annotation_like(
...
+                which we want to get required entries
+            range_end: end index of the range in
+                which we want to get required entries
 
         Returns:
             An iterator of entry elements.
         """
 
-        n = len(type_names)
+        def get_bisect_range(
```
hepengfe, line 1267: maybe we should make it a class method so that other class methods can use it as well

forte/data/data_store.py
```diff
@@ -1203,21 +1205,91 @@ def co_iterator_annotation_like(
...
+                entry_class: Name of the target type of entry. It can be a subtype
+                    of :class:`~forte.data.ontology.top.Annotation` or
+                    :class:`~forte.data.ontology.top.AudioAnnotation`.
+                search_list: A `SortedList` object on which the binary search
+                    will be carried out.
+            """
+
+            # Check if there are any entries within the given range
+            if (
+                search_list[0][constants.BEGIN_INDEX] > range_end
```
hepengfe, line 1282: Maybe you should add `range_begin` and `range_end` as parameters of this function

forte/data/data_store.py
```diff
@@ -1203,21 +1205,91 @@ def co_iterator_annotation_like(
...
+            range_end: end index of the range in
+                which we want to get required entries
 
         Returns:
             An iterator of entry elements.
         """
 
-        n = len(type_names)
+        def get_bisect_range(
+            search_list: SortedList,
```
hepengfe, line 1268: I don't think the logic of this function is correct which might cause the issue that the test case failed. We want to make sure that an entry has a range satisfy the both conditions below
1. `range_begin` < entry begin
2. entry end < `range_end`

But your check is performed on the two conditions independently which might include entry that satisfy one of the two conditions within the range. A possible fix is that we only check `begin_index`.

forte/data/data_store.py
mylibrar: Based on @hepengfe's [comments](https://github.com/asyml/forte/pull/849#discussion_r919548111), it looks like we cannot just use `begin_index` and `end_index` to specify the range?

hunterhector: My understanding of that comment is that we need to use both of them together? What else is needed to specify the range?

mylibrar: [Previously](https://github.com/mylibrar/forte/blob/0ca9602d3d287beb2521584f5fc50c2f5905cebc/forte/data/data_pack.py#L1087) we used `[range_begin, range_begin] <= [anno.begin, anno.end] <= [range_end, range_end]` to narrow down the search, but it can be problematic since we might include unexpected annotations with `range_begin<=anno.begin<range_end and range_end<anno.end`. [Here](https://github.com/asyml/forte/pull/849/files#diff-f1c68405498aa1cba27f81862f28d3edecfdc2d8501394bb3cac9d83dfa1283dR1217) is the latest implementation.

hunterhector: @Pushkar-Bhuse  please make sure the implementation here is correct according to the comments @mylibrar  and @hepengfe. The unexpected change of length in the test case may be caused by the implementation.

forte/data/data_store.py
```diff
@@ -1247,25 +1364,25 @@ def co_iterator_annotation_like(
...
-        # pointers[i] is the index of entry at (i)th sorted entry lists
-        pointers = [0] * n
+        # pointers[tn] is the index of entry of type tn
+        pointers = {key: 0 for key in all_entries_range}
 
         # compare tuple (begin, end, order of type name in input argument
         # type_names)
         # we initialize a MinHeap with the first entry of all sorted entry lists
-        # in self.__elements
+        # in all_entries
```
mylibrar, line 1373: `all_entries_range`. This also applies to line 1402 and 1406.

forte/data/data_store.py
```diff
@@ -1362,15 +1477,18 @@ def within_range(
...
-            if range_annotation is None:
+            if range_span is None:
                 yield from self.co_iterator_annotation_like(all_types)
             else:
-                for entry in self.co_iterator_annotation_like(all_types):
-                    if within_range(entry, range_annotation):
-                        yield entry
+                for entry in self.co_iterator_annotation_like(
+                    all_types,
+                    range_begin=range_span[constants.BEGIN_INDEX],
```
mylibrar, line 1485: Maybe it's better to keep the interface consistent? We can use either `range_begin, range_end` or `range_span` but maybe not both.

tests/forte/data/data_store_test.py
Pushkar-Bhuse: Yes that is correct. The new implementation handles that.

forte/data/data_store.py
Pushkar-Bhuse: Yes I have addressed all the changes and fixed the test cases.

forte/data/data_store.py
```diff
@@ -1183,8 +1182,58 @@ def get_length(self, type_name: str) -> int:
...
+            search_list[0][constants.BEGIN_INDEX]
+            > range_span[constants.END_INDEX]
+            or search_list[-1][constants.END_INDEX]
+            < range_span[constants.BEGIN_INDEX]
+        ):
+            return None
+
+        result_list = []
+
+        begin_index = search_list.bisect_left([range_span[0], range_span[0]])
```
hepengfe, line 1215: Why the parameter is a list of two `span_begin`?

forte/data/data_store.py
```diff
@@ -1183,8 +1182,58 @@ def get_length(self, type_name: str) -> int:
             delete_count = self.__deletion_count.get(type_name, 0)
             return len(self.__elements[type_name]) - delete_count
 
+    def get_bisect_range(
+        self, search_list: SortedList, range_span: Tuple[int]
+    ) -> Optional[List]:
+        """
+        Perform binary search on the specified list for target entry class.
```
hepengfe, line 1189: I think the docstring is not very clear about the usage of the function. Also, we can make the function private if we only use it internally.

forte/data/data_store.py
Pushkar-Bhuse: That's a part of the implementation. `Annotation` is sorted based on start and end index. We basically want to find a position where we can place a span of zero length at range_span[0].

forte/data/data_store.py
```diff
@@ -1317,7 +1437,7 @@ def get(
         self,
         type_name: str,
         include_sub_type: bool = True,
-        range_annotation: Optional[Tuple[int]] = None,
+        range_span: Optional[Tuple[int]] = None,
```
hunterhector, line 1440: Tuple[int, int]

forte/data/data_store.py
```diff
@@ -1328,27 +1448,25 @@ def get(
...
             A helper function for deciding whether an annotation entry is
-            inside the `range_annotation`.
+            inside the `range_span`.
             """
             if not self._is_annotation(entry[constants.ENTRY_TYPE_INDEX]):
                 return False
             return (
                 entry[constants.BEGIN_INDEX]
-                >= range_annotation[constants.BEGIN_INDEX]
+                >= range_span[constants.BEGIN_INDEX]
```
hunterhector, line 1467: I don't think you can (or should) use these constants here. These constants only specify the `BEGIN_INDEX` position for data store entries, but `range_span` is not.

forte/data/data_store.py
```diff
@@ -1183,8 +1182,60 @@ def get_length(self, type_name: str) -> int:
...
+            return None
+
+        result_list = []
+
+        begin_index = search_list.bisect_left([range_span[0], range_span[0]])
+
+        for idx in range(begin_index, len(search_list)):
+            if (
+                search_list[idx][constants.BEGIN_INDEX]
+                > range_span[constants.END_INDEX]
```
hunterhector, line 1222: Please do not use these `contants` if you don't know what they are for.

hunterhector:
@hepengfe @mylibrar please review the algorithm logic and resolve your comments. I will leave it to you two to approve

hepengfe (dismissed):

forte/data/data_store.py
Pushkar-Bhuse: Sorry I was trying to keep the previous implementation in master (there `range_annotation` was indexed in the same way). I have updated it.

mylibrar (approved):

PR #866, opened 2022-07-08
Ocr showcase

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/867

### Description of changes
Implement a notebook tutorial for OCR to show forte's capability on managing related data such as images containing text and recognized text, texts' bounding boxes and their relations..

### Test Conducted


forte/data/data_pack.py
```diff
@@ -47,13 +46,13 @@
 from forte.data.ontology.core import EntryType
 from forte.data.ontology.top import (
     Annotation,
+    Grids,
```
hunterhector, line 49: `Grids` is still an entry? I thought we discuss that it should be a data structure. Or is that change not merged?

forte/data/data_pack.py
```diff
@@ -244,7 +243,7 @@ def text(self) -> str:
     @property
     def audio(self) -> Optional[np.ndarray]:
         r"""Return the audio of the data pack"""
-        return self.get_payload_data_at(Modality.Audio, 0)
+        return cast(np.ndarray, self.get_payload_data_at(Modality.Audio, 0))
```
hunterhector, line 246: I think this `cast` is not safe, not everything can be cast this way

forte/data/data_pack.py
```diff
@@ -569,7 +565,7 @@ def set_text(
         # temporary solution for backward compatibility
         # past API use this method to add a single text in the datapack
         if len(self.text_payloads) == 0 and text_payload_index == 0:
-            from ft.onto.base_ontology import (  # pylint: disable=import-outside-toplevel
+            from ft.onto.payload_ontology import (  # pylint: disable=import-outside-toplevel
```
hunterhector, line 568: why are we importing locally here, these should be imported top level

forte/ontology_specs/payload_ontology.json
```diff
@@ -0,0 +1,76 @@
...
+        "description": "A payload that caches text data",
+        "attributes": []
+      },
+      {
+        "entry_name": "ft.onto.payload_ontology.ImagePayload",
+        "parent_entry": "forte.data.ontology.top.Payload",
+        "description": "A payload that caches image data",
+        "attributes":[]
+      },
+        {
```
hunterhector, line 27: let's make the indent aligned.

forte/ontology_specs/payload_ontology.json
```diff
@@ -0,0 +1,76 @@
...
+            "entry_name": "ft.onto.payload_ontology.JpegMeta",
+            "parent_entry": "forte.data.ontology.top.Meta",
+            "attributes":[
+              {
+              "name": "extension",
+              "type": "str"
+            },
+              {"name": "mime",
+              "type": "str"},
```
hunterhector, line 36: just use a json formatter

forte/ontology_specs/payload_ontology.json
```diff
@@ -0,0 +1,76 @@
...
+      },
+      {
+        "entry_name": "ft.onto.payload_ontology.ImagePayload",
+        "parent_entry": "forte.data.ontology.top.Payload",
+        "description": "A payload that caches image data",
+        "attributes":[]
+      },
+        {
+            "entry_name": "ft.onto.payload_ontology.JpegMeta",
+            "parent_entry": "forte.data.ontology.top.Meta",
```
hunterhector, line 29: Let's add the description for these types

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
+from abc import abstractmethod
+from multiprocessing.sharedctypes import Value
+from typing import Callable
+import numpy as np
+from forte.data.ontology.top import Meta
+
+
+class PayloadFactory:
+    """
+    A class that handles the creation of payloads. It can register meta data and check its validity.
```
hunterhector, line 10: Usage here?

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+
+    def __init__(self):
+        self.valid_meta = (
+            {}
+        )  # map from meta data name to function that handles this meta data
```
hunterhector, line 16: Why is this a tuple of dict?

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def register(self, meta: Meta):
+        """
+        A function that registers a meta data type into the factory.
+
+        Args:
+            meta_name: a Generic object that is used to register a Payload meta data type.
```
hunterhector, line 23: docstring mismatch with func

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        if meta.source_type not in ("web", "local"):
+            raise ValueError("Meta data source must be either 'web' or 'local'")
```
hunterhector, line 26: hmm, I don't think you want to control these things here. Basically you are making a lot of assumption of the data source

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    A class that handles the creation of payloads. It can register meta data and check its validity.
+    """
+
+    def __init__(self):
+        self.valid_meta = (
+            {}
+        )  # map from meta data name to function that handles this meta data
+
+    def register(self, meta: Meta):
```
hunterhector, line 18: I don't understand what is `register` is doing, it creates a dict from meta name to an always-true boolean. What's the purpose?

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+            a function that takes uri and read data from the uri, and it returns the data.
+        """
+        if not self._factory.check_meta(meta):
+            raise ValueError(f"Meta data{meta} not supported")
+        if not meta.source_type in ("web", "local"):
+            raise ValueError("Meta data source must be either 'web' or 'local'")
+
+        if meta.source_type == "local":
+            try:
+                import matplotlib.pyplot as plt  # pylint: disable=import-outside-toplevel
```
hunterhector, line 85: First, too many assumptions, such as source type, libraries that people can use.
Second, you can't do this just by hard-coding a few `if-else` branches, how are people going to provide their own loading methods?
Third, we won't rely on all these external library in Forte, it will make the library very hard to use. These parsing should be done out of the core library.

In addition, we introduce `import-outside-toplevel` once simply to reduce some legacy dependency. We don't want to use them anymore

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        if meta.source_type not in ("web", "local"):
+            raise ValueError("Meta data source must be either 'web' or 'local'")
+        self.valid_meta[type(meta)] = True
+
+    def check_meta(self, meta):
+        return type(meta) in self.valid_meta
+
+
+class Payloading:
```
hunterhector, line 33: Why are we creating so many classes? The `PayloadFactory` only stores a dictionary, and then this class only stores an empty abstract method. Can we just do it with one class?

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+            raise ValueError(f"Meta data{meta} not supported")
+        if not meta.source_type in ("web", "local"):
+            raise ValueError("Meta data source must be either 'web' or 'local'")
+
+        if meta.source_type == "local":
+            try:
+                import matplotlib.pyplot as plt  # pylint: disable=import-outside-toplevel
+            except ModuleNotFoundError as e:
+                raise ModuleNotFoundError(
+                    "ImagePayloading reading local file requires `matplotlib`"
```
hunterhector, line 88: No, what happens should be:
1. We first try to find the user function that is registered to load the URL.
2. (Optional) we can fall back to a default image loading method, these loading methods can be installed via `extra_require` in `setup.py`
3. We say that no loading method available and provide user what to do.

Also, you should know that `matplotlib` is a visualization library, it is very wierd to use it as a default image loading tool.

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+                ) from e
+            return plt.imread
+        else:
+            try:
+                from PIL import Image  # pylint: disable=import-outside-toplevel
+                import requests  # pylint: disable=import-outside-toplevel
+            except ModuleNotFoundError as e:
+                raise ModuleNotFoundError(
+                    "ImagePayloading reading web file requires `PIL` and"
+                    "`requests` packages to be installed."
```
hunterhector, line 99: Same as above, don't make these assumptions on behalf of the user.

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+                ) from e
+
+            def read_uri(uri):
+                # customize this function to read data from uri
+                uri_obj = requests.get(uri, stream=True)
+                pil_image = Image.open(uri_obj.raw)
+                return self._pil_to_nparray(pil_image)
```
hunterhector, line 106: You need to provide a way for user to customize it, now this is buried deeply in our code

forte/utils/payload_factory.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Returns:
+            a callable function that takes uri and read data from the uri, and it returns the data.
+        """
+        if meta.source_type == "local":
+            try:
+                import soundfile  # pylint: disable=import-outside-toplevel
+            except ModuleNotFoundError as e:
+                raise ModuleNotFoundError(
+                    "AudioPayloading requires 'soundfile' package to be installed."
```
hunterhector, line 131: Same as above, `soundfile` can be our optional default audio loading method

ft/onto/base_ontology.py
```diff
@@ -435,7 +463,12 @@ class CrossDocEventRelation(MultiPackLink):
     ParentType = EventMention
     ChildType = EventMention
 
-    def __init__(self, pack: MultiPack, parent: Optional[Entry] = None, child: Optional[Entry] = None):
+    def __init__(
```
hunterhector, line 466: Is this code generated? I feel like there are some manual or tool modifications to this file.

tests/forte/utils/payload_factory_test.py
```diff
@@ -0,0 +1,102 @@
...
+    def setUp(self):
+
+        self.f = PayloadFactory()
+
+    def test_image_payloading(self):
+        datapack = DataPack("image")
+        img_meta = JpegMeta(datapack)
+        img_meta.source_type = "local"
+
+        self.f.register(img_meta)
```
hunterhector, line 47: The purpose of a `register` is to associate something together. For example, here we should associate the `meta` data info with the function to load the meta data, and the function to serialize the meta data. But if you only provide one value to `register`, it cannot achieve what we want.

ocr.ipynb
```diff
@@ -0,0 +1,181 @@
+{
+ "cells": [
```
hunterhector, line 2: Can you link the notebook to the tutorial section next time? So that I can also review what it looks like in the tutorial section

ocr.ipynb
```diff
@@ -0,0 +1,181 @@
...
+    "            raise ModuleNotFoundError(\n",
+    "                \"OcrReader requires 'pytesseract' package to be installed.\"\n",
+    "            ) from e\n",
+    "        image_data = Image.open(data_source) # lazy loading\n",
+    "        ip.set_cache(image_data)\n",
+    "        ocr_text = pytesseract.image_to_string(image_data)\n",
+    "        pack.set_text(ocr_text)\n",
+    "        pack.pack_name = data_source\n",
+    "        \n",
+    "        yield pack\n",
```
hunterhector, line 86: The loading part is not what we desired, for example, it requires one to explicitly set the cache. And I don't even see metadata being used here.

```
pack: DataPack = DataPack()
jpegMeta = JpegMeta()
ImagePayload(pack, url="some_url", meta=jpegMeta) # payload_index default to 0
```

Now it is done! This is what we call `lazy_loading`, it doesn't happen anywhere near the reader, but behind the scene, when the `Payload` is used, it should load the data to cache based on url and jpegMeta, there don't need to be a step where user set the cache

ocr.ipynb
```diff
@@ -0,0 +1,181 @@
...
+    "\n",
+    "\n",
+    "# TODO: split ocr part into a processor"
```
hunterhector, line 89: yeah please do

hunterhector (commented):
1. Let's always put the images to the `asset` branch, besides the size problem, there are two more reasons: 1. git is also not good at managing binary data 2. the asset branch won't subject to frequent change like master (so the url will be more stable)
2.  I have attempted to clean up the payload structure in another [branch](https://github.com/hunterhector/forte/tree/serialization). It is not finished and I deliberately make a wrong test case, but some changes there may help you.
3. It feels like there are a lot of payload basic not ready to finish up this OCR showcase.

forte/data/data_pack.py
hepengfe: It's not merged yet and it's in https://github.com/asyml/forte/pull/827

forte/data/data_pack.py
hepengfe: There was a  typing issue `mypy forte`, and I haven't find a better way to solve it.

forte/utils/payload_factory.py
hepengfe: I was referring to the template that you wrote https://slack-files.com/T222W2A87-F03KJLMM293-a88d5aa6a3 and tried to follow it. And think I can make up the rest function after the skeleton is built up.

forte/data/data_pack.py
hunterhector: Actually, I don't think we should assume `audio` to be `ndarray`

forte/utils/payload_factory.py
hunterhector: Ok, so we won't merge any unfinished structures. I am not sure if these belong to this PR.

hepengfe:
https://github.com/asyml/forte/pull/875

PR #859, opened 2022-07-06
Payload Redesign

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/841. 

### Description of changes
* create base payload ontologies with modalities (`TextPayload`, `AudioPayload`, `ImagePayload` in `forte/data/ontology/top.py`)
* add some example ontologies generated by json files such as `JpegPayload` and `SoundFilePayload` in `ft/onto/payload_ontology.py`
* add `forte/utils/payload_factory.py`.
    * `PayloadFactory` a variable holds the mapping from payload to its loading function across modules
    * `register()` function that can be used as a decorator for payload classes to register the function mapping in `PayloadFactory`
* add test case for audio and image payloads

### Possible influences of this PR.
Describe what are the possible side-effects of the code change.

### Test Conducted
Describe what test cases are included for the PR.

An implementation template:
```
Payload:
 - uri
 - meta
 - index in data pack
 
 
TextPayload(Payload) # Json schema
​
​
{
  "entry_name": "ft.onto.ImagePayload"
  "parent_entry": "forte.data.ontology.top.Payload"
}
​
{
  "entry_name": "ft.onto.JpegMeta"
   "attributes":{
    "extension": str,
    "mime": str,
    "type_code": str,
    "version": str,
  }
}
​
{
  "entry_name": "ft.onto.JpegPayload"
  "parent_entry": "ft.onto.ImagePayload"
  "attributes":{
    {
       "attribute_name": "meta",
       "attribute_type": "ft.onto.JpegMeta",
    }
  }
}
​
​
"Jpegload" <--> "Loading function"
​
class PayloadFactory:
   def register(meta_name, func/class):
       # check whether meta_name is a real meta data
       # remember that this func will handle this meta
       
       
class Payloading():
    def route(meta):
      
class CustomPayloading(Payloading):
    import jpeg_lib
    def route_meta(meta):
       # if else conditions of meta:
          func(meta)
          
       return obj # to be store in cache.
       
       
       
pack = DataPack()
​
payload_entry = pack.add(JpegPayload("url"))
​
payload_entry = pack.get_payload_at(1)
​
payload_entry.get_type()
​
# enum
payload_entry.get_modality()
​
image1 = pack.get_payload_at(1)
​
pack.get("Sentence") # iterating all payload inside the pack.
​
pack.get_payload_at(0).get("Sentence") # iterating all payload of of text at 0
​
pack.get_payload_at(0).get("ft.onto.image.Object") # 1) showing nothing 2) raise error
​
[...]

PR #876, opened 2022-07-12
Change Grid from ontology to a data structure in `core.py`

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/817. 

### Description of changes
* change `Grids` -> `Grid` and move it to `core.py` as a data structure rather than an ontology
    * dealing edge cases of residual grid cells(last grid cells per row/column). We use the image size as upper bound for these grid cells so they might be smaller than regular grid cells. 

### Test Conducted


forte/data/ontology/core.py
```diff
@@ -635,5 +636,210 @@ def index_key(self) -> int:
         return self.tid
 
 
+class Grid:
+    """
+    Regular grid with a grid configuration dependent on the image size.
```
hunterhector, line 641: rewrite this docstring to paragraphs, it is too long

hepengfe: Split the docstring into paragraphs.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,214 @@ def index_key(self) -> int:
...
+    It is a data structure used to retrieve grid-related objects such as grid
+    cells from the image. Grid itself doesn't store any data.
```
hunterhector, line 644: A little-over claim on not storing any data, we can say `grid` only record the position/indices etc.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,214 @@ def index_key(self) -> int:
...
+
+    Based the image size and the grid shape,
```
hunterhector, line 646: `Based on`

tests/forte/grids_test.py
```diff
@@ -12,77 +12,101 @@
...
+
+        self.line = np.zeros((4, 6))
+        self.line[1, 1] = 1
+        self.line[2, 2] = 1
+        self.line[3, 3] = 1
+        # self.line[4, 4] = 1
         ip = ImagePayload(self.datapack)
-        ip.set_cache(line)
+        ip.set_cache(self.line)
         self.datapack.image_annotations.append(
```
mylibrar, line 44: Can we just use `self.datapack.add_entry()`?

forte/data/data_pack.py
```diff
@@ -171,7 +170,7 @@ def __init__(self, pack_name: Optional[str] = None):
         self._data_store: DataStore = DataStore()
         self._entry_converter: EntryConverter = EntryConverter()
         self.image_annotations: List[ImageAnnotation] = []
-        self.grids: List[Grids] = []
+        self.grids: List[Grid] = []
```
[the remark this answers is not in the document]
hepengfe: It's used in test cases, and it could be used later.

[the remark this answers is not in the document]
mylibrar: You've replaced `self.datapack.grids[0]` with `self.grid` in the `grids_test.py`, so I guess it's no longer used in that test case. You can add it back in another PR when it's used.

forte/data/ontology/core.py
```diff
@@ -635,5 +637,214 @@ def index_key(self) -> int:
...
+        if self.c_h <= 0 or self.c_w <= 0:
+            raise ValueError(
+                "cell height and width must be positive"
+                f"but the cell shape is {(self.c_h, self.c_w)}"
+                "please adjust image shape or grid shape accordingly"
+            )
+
+    def get_grid_cell(self, img_arr: np.ndarray, h_idx: int, w_idx: int):
+        """
+        Get the array data of a grid cell from image of the image payload index.
```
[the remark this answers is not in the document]
hepengfe: It passes the image data directly into it. Since this is data structure now and not directly related to a data pack.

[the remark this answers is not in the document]
mylibrar: Yeah then maybe we don't need to mention `image payload index` in the docstring?

forte/data/ontology/core.py
```diff
@@ -635,5 +637,207 @@ def index_key(self) -> int:
         return self.tid
 
 
+class Grid:
+    """
+    Regular grid with a grid configuration dependent on the image size.
+    It is a data structure used to retrieve grid-related objects such as grid
```
hunterhector, line 643: One of the main purposes of having a `Grid` is to use it to index shapes, let's highlight this.

On the other hand, what's the use case of retrieving a grid cell?

forte/data/ontology/core.py
```diff
@@ -635,5 +637,207 @@ def index_key(self) -> int:
...
+    grid cell size which is smaller than other grid cell.
+    For example, if the image
+    size is (128, 128) and the grid shape is (13, 13), the first 12 grid cells
+    per column and row will have a size of (10, 10) since 128/13=9.85, so we
+    round up to 10. The last grid cell per column and row will have a size of
+    (8, 8) since 128%10=8.
+
+    Args:
+        height: the number of grid cell per column, the unit is one grid cell.
+        width: the number of grid cell per row, the unit is one grid cell.
```
hunterhector, line 667: I think the `unit` here is redundant since we know this is `number of grid cell`

forte/data/ontology/core.py
```diff
@@ -635,5 +637,207 @@ def index_key(self) -> int:
...
+        return array
+
+    def get_grid_cell_center(self, h_idx: int, w_idx: int) -> Tuple[int, int]:
+        """
+        Get the center pixel position of the grid cell at the specific height
+        index and width index in the ``Grid``.
+        The computation of the center position of the grid cell is
+        dividing the grid cell height range (unit: pixel) and
+        width range (unit: pixel) by 2 (round down)
+        Suppose an extreme case that a grid cell has a height range
```
hunterhector, line 780: just some wording, this does not really sound "extreme" to me

hunterhector (commented):
Some minor comments on this. Btw, where would the functionality: `grid as index` for fast overlap computation be implemented? I think that's the actual reason of using grid.

forte/data/ontology/core.py
hepengfe: https://github.com/hepengfe/forte/blob/24afd2f516ea104ca55a21e422e62a1e334895d3/forte/data/ontology/core.py#L787 I have rewrite the function that it can be used to compute whether there is overlapping area between the grid cell and the image.

forte/data/ontology/core.py
hepengfe: "extreme" -> "edge"
A grid cell is usually much larger in the unit of pixel.

forte/data/ontology/core.py
hunterhector: No, I mean compute the overlap of two irregular shape via grid.

hepengfe: The full algorithm cannot be complete inside `Grid` alone.
The full algorithm is like below

stage1: build grid cell indices for each image   (use Grid)
```
for img_idx, img in enumerate(images):
      indices = grid.get_overlapped_grid_cell_indices(img) # returns grid cell indices the image is on
      for gc_idx in indices:
            d[gc_idx].append(image_idx)
```

stage2: search for candidate
```
indices = grid.build_index(img)
cand_pool = set()
for gc_idx in indices:
       cand_pool += set(d[gc_idx])
```

stage3: compute iou between the query image and images in cand_pool

forte/data/data_pack.py
```diff
@@ -171,7 +170,7 @@ def __init__(self, pack_name: Optional[str] = None):
         self._data_store: DataStore = DataStore()
         self._entry_converter: EntryConverter = EntryConverter()
         self.image_annotations: List[ImageAnnotation] = []
-        self.grids: List[Grids] = []
+        self.grid: Optional[Grid] = None
```
mylibrar, line 173: This actually not used in test cases and seems like it's not used in #859 or any other CV-related PRs. Is there any specific plan on how it will be used later?

forte/data/ontology/core.py
mylibrar: Looks like we need a tutorial or an example on this usecase for checking overlap between irregular shapes. Could you create a new issue on this?

hepengfe: https://github.com/asyml/forte/issues/888

mylibrar: We'll move the discussion to #888 and this problem shall be addressed in another PR.

mylibrar (approved):

Issue #867, opened 2022-07-08
OCR example tutorial


hepengfe:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
We want to have a notebook tutorial for OCR to show forte's capability on managing related data such as images containing text and recognized text, texts' bounding boxes and their relations.

**Describe alternatives you've considered**
No.

**Additional context**
No.

PR #877, opened 2022-07-12
OCR tutorial using Forte Ontologies to store data

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/867. 

### Description of changes
Implement a notebook tutorial for OCR to show forte's capability on managing related data such as images containing text and recognized text, texts' bounding boxes and their relations.

### Test Conducted


docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "        # construct ImageMeta and store it in DataPack\n",
+    "        return image_path\n",
+    "\n",
+    "    def _parse_pack(self, data_source: str) -> Iterator[DataPack]:\n",
+    "        r\"\"\"Takes a raw string and converts into a :class:`~forte.data.data_pack.DataPack`.\n",
+    "\n",
+    "        Args:\n",
+    "            data_source: str that contains text of a document.\n",
+    "\n",
+    "        Returns: :class:`~forte.data.data_pack.DataPack` containing Document.\n",
```
hunterhector, line 81: no `Document`, but `ImagePayload`

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "            b = b.split(\" \")\n",
+    "            ch = b[0]\n",
+    "            y1, x1 = int(b[1]), int(b[2])\n",
+    "            y2, x2 = int(b[3]), int(b[4])\n",
+    "            while ocr_text[begin].isspace():\n",
+    "                begin += 1\n",
+    "            if ocr_text[begin] == ch:\n",
+    "                character = Annotation(input_pack, begin, begin+1)\n",
+    "                cy = (y1+y2)/2\n",
+    "                cx = (x1+x2)/2\n",
```
hunterhector, line 132: I think here it shows why we don't want to use `center` as the parameter for boxes. It is more common to use the corners. Maybe we should modify this in the `Box` implementation. The box should probably be represented via the diagonal corners, and `height` and `width` also should not be the input, they should be some property computed internally

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "    \"\"\"\n",
+    "    An audio processor for speaker segmentation.\n",
+    "    \"\"\"\n",
+    "    def initialize(self, resources: Resources, configs: Config):\n",
+    "        super().initialize(resources, configs)\n",
+    "\n",
+    "    def _process(self, input_pack: DataPack):\n",
+    "        image_data = input_pack.image_payloads[0].cache\n",
+    "        ocr_text = pytesseract.image_to_string(image_data)\n",
+    "        input_pack.set_text(ocr_text)\n",
```
hunterhector, line 118: In the processor, a few inline comment explaining what we are doing would be helpful

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,535 @@
...
+    "        begin = 0\n",
+    "        for b in boxes:\n",
+    "            b = b.split(\" \")\n",
+    "            ch = b[0]\n",
+    "            y1, x1 = int(b[1]), int(b[2])\n",
+    "            y2, x2 = int(b[3]), int(b[4])\n",
+    "            while ocr_text[begin].isspace():\n",
+    "                begin += 1\n",
+    "            if ocr_text[begin] == ch:\n",
+    "                character = Annotation(input_pack, begin, begin+1)\n",
```
hunterhector, line 130: I think it is fine to add a `Character` annotation in `base_ontology`.

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,535 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
```
hunterhector, line 191: The output list is a bit too long. Maybe we should only show a few characters.

Another interesting idea is, maybe we can add a simple NLTKTokenzier to get the tokens, and then:

```python
for token in pack.get(Token):
    charcters = list(token.get(Character))
    top_left = pack.get_links_by_parent(charcters[0])[0].corners[0]
    bottom_right = pack.get_links_by_parent(charcters[-1])[0].corners[-1]

    # This step can be done in a processor, but can also be here.
    box = pack.add_entry(Box(top_left, bottom_right)) 
    # Visuliaze the bounding box (image with a box around it).
    print(token.text)
    box.visualize()         
```

hepengfe: Added as suggested.

docs/notebook_tutorial/ocr.ipynb
hepengfe: Added.

docs/notebook_tutorial/ocr.ipynb
hepengfe: https://github.com/asyml/forte/pull/879

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -0,0 +1,440 @@
...
...adQYmUwGeKxlWd6+fXsqlbrzzjuvuuqqhoaGRCLR19e3ZcuW5557bt++fS0tLffdd184HJ49e/Y3vvGNs88+G9XlXV1df/7zn19//fW+vr6Ojo6dO3eeccYZ4XDY7XbD9fXoo4++9957uVyuoaHhrrvuWrRoEX4RT+Spp55as2YNwq+VlZXnnnuu3W6nEmBJq0RRNO0uhFBLSkquvvrqaDT6y1/+MhQKPfzww/X19Q0NDbjqLVu2PProo8hPQBCDX6OIhsqago+AwBcEpw39YlqDlILW0gR+kWw2OzAw4HQ6r7jiinvuuae6uhqEo6KiYtasWUePHj169Ggymezt7a2qqvrRj36EQNXg4KDNZlu4cKHRaGxra3vrrbccDsf69etvuukmFBYZDIbXXnsNnrP58+ffeOONl156qd/vRyZ4VVVVWVlZOp2ORCJbt25tamqCTYZyKso6p7JEWZaR0JDL5bq7u3t6elwu19y5c++9994lSxajnQnySKqrq41GI/LMDh8+3NraWlpaSsIBPJ+jMjT42N9+++23334bhXtXXXXVl770pWnTpqGAC06aoqIij8fzxz/+EeEhm82GmiwcE0myyO0gAQ6oozHG0ul0cXGxw/FhWxJcnclkstttsFZxSrC8EXZEfV9ra+s777wD7jVv3ryrrrrqggsuCIfDfr9fVdWhoaFLL73U5/M1NzcjlveHP/zhcsZQKsEYg15DJpOJRCII2ezevfuxxx575ZVXJEm68sorb7nllvnz5yNdz2KxOJ1OSnyhfNu81scQ3judTufz+fC+w+EYN25ceXk5/gp1ooLWFAU291VXXVVRUaHT6aD2ic83NjZ+8MEHsVjsvffeu+GGG6qqqnK53OzZs2fNmtXd3R0IBHbt2rVo0aJkMknys0VFvgMHGrds2YIig2nTphUXF4MPvfzyy7hF06dP/9d//dc5c+YwxmKxmE6nq62tnTBhQl1dncfjefvtt3t7e7du3bpv374pU6aQ9EA0GoXkmMFgsFjMcCRQI0JQDQiTIncQ5aJmsxk54waDIZPJ2O12u91+xRVXfO9736uoKOvp6bPb7S6Xo79/EJcAlzNCcqiyVFU1nU4PDQ3FYrHS0tKSkpIlS5ZccsklCNIhqby6utpqtXZ2djY3NweDwdbW1kQi4fF4PB6P1WqBBBqy/hsaGn7wgx9cccUV
```
mylibrar, line 353: Do we really need this checking here? I guess `OcrCharacterProcessor` has already ruled out all the empty text.

mylibrar, line 397: Maybe `OcrTokenProcessor` has already checked this condition.

forte/data/ontology/top.py
```diff
@@ -878,15 +878,36 @@ def image(self):
...
     @property
     def max_y(self):
-        return self.image.shape[0] - 1
+        return self._image_height - 1
+
+    @property
+    def image_shape(self):
+        return (self._image_height, self._image_width)
+
+    def set_image_shape(self, width, height):
```
mylibrar, line 897: Not sure if it's a good idea to allow flexible image size setting. What if it's in conflict with the size of image at `self._image_payload_idx`?

ft/onto/base_ontology.py
```diff
@@ -311,7 +322,12 @@ class PredicateLink(Link):
     ParentType = PredicateMention
     ChildType = PredicateArgument
 
-    def __init__(self, pack: DataPack, parent: Optional[Entry] = None, child: Optional[Entry] = None):
```
mylibrar, line 314: I guess we don't need to `black` this file. The only update here should be the new `Character` class.

forte/data/ontology/top.py
hepengfe: It's only a temporary solution for this PR. There is more implementation needs to be done as this issue suggests. https://github.com/asyml/forte/issues/885

mylibrar (approved):

Issue #878, opened 2022-07-13
New Box API


hepengfe:
**Is your feature request related to a problem? Please describe.**
Current Box API requires the center and height and width but a more common API requires corners on the diagonal to initialize such as top-left and bottom right corners.

**Describe the solution you'd like**
Create one extra Box ontology while keeping the current Box.

**Describe alternatives you've considered**
 Replace the current Box with the new Box.

**Additional context**

PR #879, opened 2022-07-14
Ontology Box init by top left point and bottom right point

hepengfe:
This PR fixes  https://github.com/asyml/forte/issues/878.

### Description of changes

* The default Box init method: height and width and center -> init Box by top left point and bottom right point
* move init by  height &  width & center to a classmethod

It can simplify the Box API and enable two ways of initialization.

### Test Conducted
Tested wrong box shape and wrong x/y coordinates of top left point and bottom right point.

### DIFF
--- a/forte/data/ontology/top.py
+++ b/forte/data/ontology/top.py
@@ -1052,7 +1052,7 @@ class Region(ImageAnnotation):
         else:
             self._image_payload_idx = image_payload_idx
 
-    def compute_iou(self, other) -> int:
+    def compute_iou(self, other) -> float:
         intersection = np.sum(np.logical_and(self.image, other.image))
         union = np.sum(np.logical_or(self.image, other.image))
         return intersection / union
--- a/forte/data/ontology/top.py
+++ b/forte/data/ontology/top.py
@@ -1147,6 +1147,39 @@ class Box(Region):
             image_payload_idx,
         )
 
+    def compute_iou(self, other) -> float:
+        """
+        A function computes iou(intersection over union) between two boxes
+        (unit: pixel).
+        It overwrites the ``compute_iou`` function in it's parent class
+        ``Region``.
+        Args:
+            other: the other ``Box`` object to be computed with.
+        Returns:
+            A float value which is (intersection area/ union area) between two
+            boxes.
+        """
+        if not isinstance(other, Box):
+            raise ValueError(
+                "The other object to compute iou with is"
+                " not a Box object."
+                "You need to check the type of the other object."
+            )
+
+        if not self.is_overlapped(other):
+            return 0
+        box_x_diff = min(
+            abs(other.box_max_x - self.box_min_x),
+            abs(other.box_min_x - self.box_max_x),
+        )
+        box_y_diff = min(
+            abs(other.box_max_y - self.box_min_y),
+            abs(other.box_min_y - self.box_max_y),
+        )
+        intersection = box_x_diff * box_y_diff
+        union = self.area + other.area - intersection
+        return intersection / union
+
     @property
     def center(self):
         return (self._cy, self._cx)
--- a/forte/data/ontology/top.py
+++ b/forte/data/ontology/top.py
@@ -1081,19 +1081,58 @@ class Box(Region):
     def __init__(
         self,
         pack: PackType,
+        tl_point: List[int],
+        br_point: List[int],
+        image_payload_idx: int = 0,
+    ):
+        super().__init__(pack, image_payload_idx)
+        if tl_point[0] < 0 or tl_point[1] < 0:
+            raise ValueError(
+                f"input parameter top left point indices ({tl_point}) must"
+                "be non-negative"
+            )
+        if br_point[0] < 0 or br_point[1] < 0:
+            raise ValueError(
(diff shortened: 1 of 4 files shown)


forte/data/ontology/top.py
```diff
@@ -1105,9 +1190,10 @@ def corners(self):
...
-            (self._cy + h_offset, self._cx + w_offset)
-            for h_offset in [-0.5 * self._height, 0.5 * self._height]
-            for w_offset in [-0.5 * self._width, 0.5 * self._width]
+            (self.y0, self.x0),
+            (self.y0, self.x1),
+            (self.y1, self.x0),
+            (self.y1, self.x1),
         ]
 
     @property
```
mylibrar, line 1199: Is `box_min_x` just an equivalent of `max(self.x0, 0)`? Could you explain a little bit on why we need `box_min_x/y` and `box_max_x/y`? I think maybe we can compute IOU or check overlapping directly through `self.x0/1` and `self.y0/1`?

forte/data/ontology/top.py
```diff
@@ -1189,42 +1250,30 @@ class BoundingBox(Box):
     the image/grid.
 
     Args:
-        pack: The container that this BoundingBox will
-            be added to.
+        pack: the container that this ``Box`` will be added to.
```
mylibrar, line 1253: This should be ``BoundingBox``

forte/data/ontology/top.py
```diff
@@ -1189,42 +1250,30 @@ class BoundingBox(Box):
...
-        grid_cell_w_idx: the width index of the associated grid cell in
-            the grid, the unit is one grid cell.
-
+            it defaults to 0 which meaning it will load the first image payload.
     """
 
-    def __init__(
-        self,
+    @classmethod
+    def init_from_center_n_shape(
```
mylibrar, line 1264: What's the difference between `BoundingBox.init_from_center_n_shape` and `Box.init_from_center_n_shape`? If they are exactly the same, then why do we need `BoundingBox` for this PR?

tests/forte/image_annotation_test.py
```diff
@@ -56,3 +56,41 @@ def test_image_annotation(self):
...
+
+        self.assertRaises(ValueError, wrong_box)
+
+        def wrong_box():
+            # negative height
+            Box(self.datapack, [2, 2], [2, 1], 0)
+
+        self.assertRaises(ValueError, wrong_box)
+
+        b3 = BoundingBox.init_from_center_n_shape(self.datapack, 4, 4, 5, 5)
```
mylibrar, line 96: I guess it's the same as `Box.init_from_center_n_shape`

forte/data/ontology/top.py
hepengfe: I haven't thought of an actually difference between `BoundingBox` and `Box` yet. Previously, `BoundingBox` requires `Grid` to function correctly but it doesn't seem to be the case. I am thinking we can keep it here and we can add more attributes/functions as needed later.

tests/forte/image_annotation_test.py
hepengfe: It shows the child class can init from center and shape as well, and it keeps the child class type.

forte/data/ontology/top.py
```diff
@@ -1067,33 +1067,118 @@ class Box(Region):
...
         image_payload_idx: int = 0,
     ):
-        # assume Box is associated with Grids
-        super().__init__(pack, image_payload_idx)
+        """
+        A class method to initialize a ``Box`` from a box's center position and
+        shape.
+
+        Args:
+            pack: the container that this ``BoundingBox`` will be added to.
```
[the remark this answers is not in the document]
hepengfe: Corrected it.

hepengfe:
Actually, not one of those PR require `BoundingBox` so I think we can still include here to differentiate `BoundingBox` and `Box`. And current  `BoundingBox` is in vanilla status and can be added more functionality as needed.

forte/data/ontology/top.py
```diff
@@ -1105,26 +1190,27 @@ def corners(self):
...
 
     @property
     def box_min_x(self):
-        return max(self._cx - round(0.5 * self._width), 0)
+        return max(self.x0, 0)
 
     @property
     def box_max_x(self):
-        return min(self._cx + round(0.5 * self._width), self.max_x)
+        return min(self.x1, self.pack.image_width)
```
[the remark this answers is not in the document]
hepengfe: it's actually a mistake and I have corrected it.

forte/data/ontology/top.py
```diff
@@ -1105,26 +1190,27 @@ def corners(self):
...
+            (self.y0, self.x0),
+            (self.y0, self.x1),
+            (self.y1, self.x0),
+            (self.y1, self.x1),
         ]
 
     @property
     def box_min_x(self):
-        return max(self._cx - round(0.5 * self._width), 0)
+        return max(self.x0, 0)
```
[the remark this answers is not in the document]
hepengfe: removed the comparison

forte/data/ontology/top.py
hepengfe: Removed and it's ready for review now.

mylibrar (approved):

Issue #847, opened 2022-07-01
Integrate `DataStore` with `Payload` entry


mylibrar:
**Is your feature request related to a problem? Please describe.**
This is a follow-up issue to #840. After we restructure the payload system in #828, we have a new ontology called `Payload` and it's not entirely integrated to `DataStore` yet. Currently `DataPack` stores three payload lists for different modalities. We want to migrate them into `DataStore` without affecting the current interfaces.

**Depends on**
- #840 

**Describe the solution you'd like**
* We should maintain all payload lists in `DataStore`
  * `__getitem__` should be efficient
* Design how `Payload`'s attributes can be stored in `DataStore` as primitive and built-in types.
  * Nee to determine how to handle the `dataclass` and non-`dataclass` fields.
* We should keep the previous interfaces of accessing payloads, e.g., `pack.text_payloads[text_payload_index]`

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.

mylibrar:
https://github.com/asyml/forte/pull/882#discussion_r951667279

Issue #835, opened 2022-06-16
Store attribute type of `Entry`


mylibrar:
**Is your feature request related to a problem? Please describe.**
We want to store the type of all dataclass fields of `Entry` to facilitate the conversion between primitive data in `DataStore` and the actual object [retrieved](https://github.com/asyml/forte/blob/ac34f102ebb3319c1e23f2f81e20a7e9e6c5e60d/forte/data/data_pack.py#L1499)/[set](https://github.com/asyml/forte/blob/ac34f102ebb3319c1e23f2f81e20a7e9e6c5e60d/forte/data/data_pack.py#L1522) by users.

After the integration of `DataStore`, `Entry`'s attributes are stored into `DataStore` object, which only maintains primitive types. When the attribute itself is also an `Entry`, it will be stored as an integer (`tid`) in `DataStore`. When users access this attribute later, it will be converted back to `Entry` again. Currently we determine the type of the attribute value [on the fly](https://github.com/asyml/forte/blob/ac34f102ebb3319c1e23f2f81e20a7e9e6c5e60d/forte/data/data_pack.py#L1514), which is inefficient and unstable.

**Describe the solution you'd like**
We can maintain a dictionary for each dataclass fields that maps the field name to the corresponding type. One potential solution is to use `typing_inspect.get_origin()` and `typing_inspect.get_args()` to parse the typing info stored in `Entry.__dataclass_fields__`.

**Describe alternatives you've considered**
The attribute type info can also be used in `DataStore` to initialize default data structures.

**Additional context**
- This is part of the data efficiency project
- This PR should be made to the `master` branch.
- After the data tuple class is finished, we may switch to a new branch for integration.

mylibrar:
May be related to #804.

mylibrar:
May be related to #804.

PR #882, opened 2022-07-18
Refactoring Data Store Structure

### LINKED ISSUE #874: Change Data Store Structure for Annotation Type Entries
The current structure of entries stored in Data Stores is not scalable. This was made apparent by the requirement of adding the `payload index` to a data store entry. This creates a need to change the way data is stored in a Data Store.

One option would be to change the order in which attributes are stored in the entry list in such a way that universal attributes like `tid`, `type name` are stored first. The remaining attributes can be added as `data class` fields. This structure can be followed for all Data Store entries (not just Annotation Type Entries)

Pushkar-Bhuse:
This PR fixes #874  

### Description of changes
This PR attempts to transform how data is stored in the `DataStore` entry. The main idea behind this new format is that all attributes of an entry other than its `tid` and `type_name` should be considered as `data class` attributes. As a result of this, the DataStore format of any entry can be visualized as : `[tid, type_name, ....(dataclass attributes)...]`. 

For example, the `type_attrubutes` of `Sentence` can be seen as 
```
{
            "attributes": {
                "begin": 2,
                "end": 3,
                "payload_idx": 4,
                "speaker": 5,
                "part_id": 6,
                "sentiment": 7,
                "classification": 8,
                "classifications": 9,
            },
            "parent_class": set(),
        }
```

The way this is implemented is by creating a class variable for `Entry` called `cached_attribute_data`. This is a `dict` that stores the **initial** values of `dataclass` attributes. The implementation makes sure that before initializing a data store entry for a given entry object, the `cached_attribute_data` dict store all data class attributes and their initial values. There are two ways in which a `dataclass` attribute can be added to `cached_attribute_data`

1. Attribute has a `entry_setter` property: In this case, the entry will automatically be added to `cached_attribute_data`.
2. Attribute does not have a `entry_setter` property: In this case, attributes values are store in the entry object. These are fetched before the creation of the data store entry to populate `cached_attribute_data`.

1. Since all attributes are now `dataclass` attributes, we do not need to rely on `constants`. Instead, we use the function `get_datastore_attr_idx` to fetch the position in the datastore where a given attributes is stored.
2. The new format makes `DataStore` more scalable now since any new attribute can be added to the entry as well as its datastore but declaring it as a `dataclass` attribute

### Test Conducted
The main aim of this PR was to keep the outermost interface unchanged and still be able to pass the `data_store_test`, `data_pack_test` and `multi_pack_test`


forte/data/data_pack.py
```diff
@@ -1322,10 +1341,10 @@ def _generate_link_entry_data(
                 )
 
             a_dict["parent"].append(
-                np.where(data[parent_type]["tid"] == link.parent)[0][0]
+                np.where(data[parent_type]["tid"] == link.parent.tid)[0][0]  # type: ignore
```
Pushkar-Bhuse, line 1344: This is a hacky fix. But this stems from another problem that we are attempting solve through #835

forte/data/data_store.py
```diff
@@ -759,27 +772,41 @@ def _add_entry_raw(
...
             ``tid`` of the entry.
         """
-        if entry_type in (Annotation, AudioAnnotation):
+        if any(
+            self._is_subclass(type_name, cls)
+            for cls in (Annotation, AudioAnnotation)
+        ):
+            begin_idx = self._type_attributes[type_name][
+                constants.TYPE_ATTR_KEY
+            ]["begin"]
```
hunterhector, line 781: hard coded string

forte/data/data_store.py
```diff
@@ -759,27 +772,41 @@ def _add_entry_raw(
...
+                Group,
+                Generics,
+                ImageAnnotation,
+                Grids,
+                Payload,
+                MultiPackLink,
+                MultiPackGroup,
+                MultiPackGeneric,
+            ]
+        ):
```
hunterhector, line 809: Does this mean that we have to update this list for every new `top` type we created? Can we check whether `cls` is `entry`?

forte/data/data_store.py
```diff
@@ -788,7 +815,7 @@ def _add_entry_raw(
         else:
             raise NotImplementedError(
                 "_add_entry_raw() is not implemented "
-                f"for entry type {entry_type}."
+                f"for the parent entry of {type_name}."
```
hunterhector, line 818: This change makes the error message a bit weird, why emphasize `parent entry` here?

forte/data/data_store.py
```diff
@@ -835,18 +859,97 @@ def _create_new_entry(
...
         entry: List[Any] = []
 
-        entry.extend(attribute_data)
-
         entry += [tid, type_name]
         entry += self._default_attributes_for_type(type_name)
 
         return entry
 
+    def get_datastore_attr_idx(self, type_name: str, attr: str) -> int:
```
hunterhector, line 867: need docstring for internal functions for bookkeeping purposes.

forte/data/data_store.py
```diff
@@ -835,18 +859,97 @@ def _create_new_entry(
...
+            return self._type_attributes[type_name][constants.TYPE_ATTR_KEY][
+                attr
+            ]
+        except KeyError as e:
+            raise KeyError(
+                f"Attribute {attr} is not a part"
+                f"of type attributes for {type_name}"
+            ) from e
+
+    def _validate_entry(self, new_entry: List) -> List:
```
hunterhector, line 878: need docstring for internal functions for bookkeeping purposes.

forte/data/data_store.py
```diff
@@ -835,18 +859,97 @@ def _create_new_entry(
...
+                )
+
+        return new_entry
+
     def add_entry_raw(
         self,
         type_name: str,
-        attribute_data: List,
-        base_class: Type[Entry],
+        attribute_data: Dict,
```
hunterhector, line 952: Actually the use of a `Dict` here defeats the purpose of `DataStore`. DataStore is designed to efficiently handle data structures, and adding data should be list minimum array operations. But using Dict adds some additional speed lost. Further, this actually changes the interface types.

forte/data/data_store.py
```diff
@@ -1198,27 +1320,38 @@ def _get_bisect_range(
                 will be carried out.
             range_span: a tuple that indicates the start and end index
                 of the range in which we want to get required entries
+            type_name: Type of entry represented by the DataStore
 
         Returns:
             List of entries to fetch
         """
 
+        begin = self.get_datastore_attr_idx(type_name, "begin")
```
hunterhector, line 1329: hard coded string

forte/data/data_store.py
```diff
@@ -1485,19 +1631,18 @@ def within_range(entry: List[Any], range_span: Tuple[int, int]) -> bool:
...
                 else:
                     for entry in self.__elements[type]:
-                        if (
-                            entry[constants.PARENT_TID_INDEX]
-                            in self.__tid_ref_dict
-                        ) and (
-                            entry[constants.CHILD_TID_INDEX]
-                            in self.__tid_ref_dict
+                        parent_idx = self.get_datastore_attr_idx(
+                            entry[constants.ENTRY_TYPE_INDEX], "parent"
```
hunterhector, line 1635: hard coded string

hunterhector (commented):
This is a large PR, maybe too large so I am not too confident in reviewing this. There are currently many changes in interfaces that won't be merged too. Maybe we should actually break down this work into multiple PRs instead of doing it at once.

1. We need to understand the DataPack interface layer and DataStore interface layer, we don't want to make **any** changes to public PRs of DataPack
2. For safety, let's do not add any `type: ignore` in this PR since a lot of them are caused by real errors
3. Please double-check whether serialization/deserialization data packs work after this PR.

forte/data/data_pack.py
```diff
@@ -446,7 +446,7 @@ def get_payload_at(
 
         """
         supported_modality = [enum.name for enum in Modality]
-
+        payload_length: int = 0
```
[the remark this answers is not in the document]
Pushkar-Bhuse: I had added this in order to fix pylint errors which did not allow initializing a variable in the conditional statement.

forte/data/data_pack.py
```diff
@@ -1548,18 +1567,29 @@ def update(self, datapack: "DataPack"):
...
         r"""Save an existing entry object into DataStore"""
-        self._entry_converter.save_entry_object(entry=entry, pack=self)
+        self._entry_converter.save_entry_object(
+            entry=entry, pack=self, attribute_data=attribute_data
+        )
 
         if isinstance(entry, Payload):
-            if entry.modality == Modality.Text:
+            if (
+                getattr(Modality, attribute_data["modality_name"])
```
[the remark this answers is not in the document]
Pushkar-Bhuse: That is because the `modality_name` is stored in the primitive form (`str`) in the datastore. That is why, when going through the conditions, we need to convert it back to its original form.

forte/data/multi_pack.py
```diff
@@ -715,6 +715,25 @@ def get_single_pack_data(
             context_type, request, skip_k
         )
 
+    def get_data_store_attribute_idx(
```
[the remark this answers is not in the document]
Pushkar-Bhuse: I feel we do, based on my comment above

tests/forte/remote_processor_test.py
```diff
@@ -83,3 +83,7 @@ def test_remote_processor(self, input_output_pair):
         utterance = get_last_utterance(res, "ai")
         self.assertEqual(len([_ for _ in res.get(Utterance)]), 2)
         self.assertEqual(utterance.text, o_str)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Sure

forte/data/data_pack.py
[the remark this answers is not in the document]
hunterhector: what do you mean? this variable is not used later, I saw another variable named `payloads_length`

forte/data/data_pack.py
```diff
@@ -499,6 +499,25 @@ def get_payload_data_at(
         """
         return self.get_payload_at(modality, payload_index).cache
 
+    def get_data_store_attribute_idx(
```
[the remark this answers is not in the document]
hunterhector: I don't think so: 
1. this does not looks like a public api suitable at the data pack layer, it is exposing internal implementation details
2. this function simply calls another function with almost the same signature, why don't we just call `self._data_store.get_datastore_attr_idx(type_name, attr_name)`

forte/data/data_pack.py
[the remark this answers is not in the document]
hunterhector: there are two questions here:
1. let's try to use less hard code, such as `modality_name` here.
2. can we access the enum using the [name](https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes)? Using `getattr` looks hacky to me

forte/data/ontology/core.py
```diff
@@ -492,18 +492,13 @@ def data(self, array: Union[np.ndarray, List]):
...
 
 class BaseLink(Entry, ABC):
-    def __init__(
-        self,
-        pack: ContainerType,
-        parent: Optional[Entry] = None,
-        child: Optional[Entry] = None,
-    ):
-        super().__init__(pack)
+    def __init__(self, pack: ContainerType, attribute_data: Dict):
```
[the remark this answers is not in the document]
Pushkar-Bhuse: I get the idea. I will discuss this with @mylibrar and move to another way of doing this, one that does not require this dict `attribute_data`.

forte/data/data_store.py
Pushkar-Bhuse: Okay, I will take this into account

forte/data/data_store.py
Pushkar-Bhuse: We can't just check for `Entry` because just above that there is a check for `Annotation` and `AudioAnnotation` as well.

forte/data/data_pack.py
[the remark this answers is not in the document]
hunterhector: The fact that we need such access such the data store layer from outside is concerning by itself. Instead of exposing this method to the outside, we should provide other methods for stuff like `top.py` to access the data pack data. 

Note our rule of thumb here, by integrating with data store, we do not want to change the interfaces of data pack.

forte/data/data_store.py
hunterhector: why would the check for `Annotation` affect here? We first check for `Annotation`, if not, then we check for the other `Entry`, no?

Pushkar-Bhuse: I'll try that

forte/data/data_pack.py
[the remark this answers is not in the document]
Pushkar-Bhuse: Okay I'll keep that in mind

mylibrar:
> The way this is implemented is by creating a class variable for `Entry` called `cached_attribute_data`. This is a `dict` that stores the **initial** values of `dataclass` attributes. The implementation makes sure that before initializing a data store entry for a given entry object, the `cached_attribute_data` dict store all data class attributes and their initial values. There are two ways in which a `dataclass` attribute can be added to `cached_attribute_data`
> 
> 1. Attribute has a `entry_setter` property: In this case, the entry will automatically be added to `cached_attribute_data`.
> 2. Attribute does not have a `entry_setter` property: In this case, attributes values are store in the entry object. These are fetched before the creation of the data store entry to populate `cached_attribute_data`.
> 

I'm thinking we might want to make the behavior consistent. Right now we are maintaining two distinctive approaches to set dataclass attributes before and after registering property function. But it's out of the scope of this PR and it's not of high priority.

forte/data/entry_converter.py
```diff
@@ -157,8 +184,8 @@ def save_entry_object(
             )
 
         # Store all the dataclass attributes to DataStore
-        for attribute in entry.__dataclass_fields__:
-            value = getattr(entry, attribute, None)
+        for attribute, value in entry.cached_attribute_data.items():
+            # value = getattr(entry, attribute, None)
```
mylibrar, line 188: We can remove this comment if it's no longer useful.

forte/data/ontology/core.py
```diff
@@ -494,6 +501,11 @@ def data(self, array: Union[np.ndarray, List]):
 
 
 class BaseLink(Entry, ABC):
+
+    # this type Any is needed since subclasses of this class will have new types
```
mylibrar, line 505: What's the purpose of adding these two fields here? Does it make them become non-dataclass attribute?

forte/data/ontology/top.py
```diff
@@ -382,22 +304,29 @@ def get_child(self) -> Entry:
...
+    member_type: Type[Entry]
+
+    MemberType = Entry
 
     def __init__(
         self,
         pack: PackType,
         members: Optional[Iterable[Entry]] = None,
     ):  # pylint: disable=useless-super-delegation
+        # attribute_data = {"member_type": self.MemberType, "members": members}
```
mylibrar, line 328: Remove commented code lines if they are not useful.

forte/data/ontology/top.py
```diff
@@ -412,7 +341,10 @@ def add_member(self, member: Entry):
                 f"The members of {type(self)} should be "
                 f"instances of {self.MemberType}, but got {type(member)}"
             )
-        self.pack.get_entry_raw(self.tid)[MEMBER_TID_INDEX].append(member.tid)
+        if self.members is None:
```
mylibrar, line 344: When do we hit this condition? I thought after `__init__` is called, `self.members` should no longer be `None`.

tests/forte/data/data_store_serialization_test.py
```diff
@@ -36,86 +39,111 @@ def setUp(self) -> None:
         DataStore._type_attributes = {
             "ft.onto.base_ontology.Document": {
                 "attributes": {
-                    "sentiment": 4,
-                    "classifications": 5,
+                    "begin": 2,
```
mylibrar, line 42: What if begin and end are not included in "attributes"? Will it trigger an error.

forte/data/base_pack.py
```diff
@@ -550,11 +566,24 @@ def entry_setter(cls: Entry, value: Any, attr_name: str, field_type):
                 tid=cls.tid, attr_name=attr_name, attr_value=attr_value
             )
 
+        # If this is the first time an entry of this type is
+        # created, its attributes do not have a getter and setter
+        # property associated with them. In this case, we manually
+        # add all dataclass attributes to the cached_attribute_data
+        # dict.
+        for name in entry.__dataclass_fields__:
```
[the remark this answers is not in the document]
Pushkar-Bhuse: This is correct. However, the first two points are relevant for initialization and the third point is useful otherwise. I will make sure to specify all these scenarios in required places.

tests/forte/data/data_store_serialization_test.py
Pushkar-Bhuse: Yes I think they will, since `begin` and `end` are attributes of `Annotation`, they are required to initialize a data store entry (since they are stored in a sorted list based on the `begin` and `end`) and also in other operations where we fetch the index of the position where these attributes are stored (using `get_datastore_attr_idx`)

forte/data/ontology/top.py
Pushkar-Bhuse: That's right. I will remove it

forte/data/data_store.py
```diff
@@ -1044,20 +1072,166 @@ def _create_new_entry(
...
+        """
+        try:
+            return DataStore._type_attributes[type_name][
+                constants.ATTR_INFO_KEY
+            ][attr][constants.ATTR_INDEX_KEY]
+        except KeyError as e:
+            raise KeyError(
+                f"Attribute {attr} is not a part "
+                f"of type attributes for {type_name}"
+                f"or an entry of type {type_name} does not exist"
```
hunterhector, line 1100: Would an index out of bound happen here?

forte/data/data_store.py
```diff
@@ -1044,20 +1072,166 @@ def _create_new_entry(
...
+
+    def initialize_and_validate_entry(
+        self, entry: List, attribute_data: List
+    ) -> List:
+        r"""This function performs validation checks on the initial
+        attributes added to a data store entry. This functions also
+        modifies the value of certain attributes to fit data store's
+        purpose of storing primitive types. For example, In the data store
+        entry of type :class:`~forte.data.ontology.top.Group`, attribute
+        `member_type` is converted from an object to `str`. When initializing
```
hunterhector, line 1111: Sometimes we probably support initializing attributes with pure string in the first place. Would this be supported?

tests/forte/data/data_store_test.py
```diff
@@ -128,67 +125,137 @@ def setUp(self) -> None:
         self.reference_type_attributes = {
             "ft.onto.base_ontology.Document": {
                 "attributes": {
-                    "document_class": {"index": 4, "type": (list, (str,))},
-                    "sentiment": {"index": 5, "type": (dict, (str, float))},
+                    "begin": {"index": 2, "type": (None, (int,))},
```
hunterhector, line 128: Recalled that Suqi mentioned the test case is kinda duplicated in these two test cases, we should centralize them. Have we created tickets for that?

forte/data/base_pack.py
```diff
@@ -541,16 +568,35 @@ def entry_getter(cls: Entry, attr_name: str):
...
+                pack = cls.pack
+            except AttributeError as err:
+                if attr_name not in cls.cached_attribute_data:
+                    cls.cached_attribute_data[attr_name] = value
+                    return
+                else:
+                    raise KeyError(
+                        "You are trying to overwrite the value "
+                        f"of {attr_name} for a data store entry "
+                        "before it is created."
```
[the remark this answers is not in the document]
Pushkar-Bhuse: So this happens when a value stored for an attributes in `cached_attributes` (this indicates that this value is added to the entry object but not with its corresponding data store entry) is tried to be updated. I mainly added this part to ensure consistency of attribute values in the entry object and data store. Does this sound okay?

I can add a short test case by creating a dummy entry where the attribute value is updated before the data store entry is created.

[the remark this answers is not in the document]
hunterhector: There is no real need to add tests for test's sake by covering this branch. In normal usage, could this happen?

forte/data/base_pack.py
```diff
@@ -594,6 +640,18 @@ def entry_setter(cls: Entry, value: Any, attr_name: str):
...
+        # created, its attributes do not have a getter and setter
+        # property associated with them. In this case, we manually
+        # add all dataclass attributes to the cached_attribute_data
+        # dict.
+        for name in entry.__dataclass_fields__:
+            if (
+                hasattr(type(entry), name)
+                and not isinstance(getattr(type(entry), name), property)
+            ) or (not hasattr(type(entry), name)):
+                entry.cached_attribute_data[name] = getattr(entry, name, None)
```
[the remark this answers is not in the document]
Pushkar-Bhuse: So one workaround that I have in mind for this is to fetch the `_type_attributes` dict somehow to check if the current entry being created is present in `type_attributes` or no. In order to do that, I will have to add a method in `DataStore` to fetch the `_type_attributes`. Would that be okay?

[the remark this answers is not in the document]
hunterhector: fetching `_type_attributes` should be fine. what's the difference in terms of the cost here?

further, how often would each of these boolean statements being invoked?

forte/data/data_store.py
```diff
@@ -751,7 +785,7 @@ def fetch_entry_type_data(
                 # None. But to maintain the consistency of
                 # type_dict, we only store the type of every
                 # value, even None.
-                attr_class = type(None)
+                attr_class = None
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Got it. Also, there are scenarios when the output of `get_origin` is also `None`. Should these be converted to `NoneType` as well?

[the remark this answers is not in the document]
hunterhector: If the purpose of `get_origin` is to get the type and the output is actually `NoneType` then yes.

forte/data/data_store.py
Pushkar-Bhuse: I don't think so because we are accessing the position index from the `_type_attributes` dict which has all these keys set at all times.

forte/data/data_store.py
Pushkar-Bhuse: No if a pure string is passed into `get_full_module name`, its returns its value as `str`. A solution can be that if a string is passed, we can validate it by first converting it to an object using `get_class` and raise an error if its not a valid entry. Does that sound okay?

hunterhector: We need to look at the path when creating an entry now. At DataPack level, we should support creating an entry object via "class" or "classname". Note that in datastore we can check whether a string is a an entry using the _type_attribute_dict.

And we shouldn't convert back and forth between these two types, this defeats the purpose of being "data efficient", we should support string and class naturally.

forte/data/base_pack.py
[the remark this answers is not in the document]
Pushkar-Bhuse: So, checking if an entry is present in `type_attributes` would be O(1) time. 

- If an entry is present in `type_attributes`, we can assume that a getter and setter method is associated with its `dataclass` fields which in turn means that `cached_attributes` dict is already populated (so nothing needs to be done)
- If an entry is not present in `_type_attributes`, `cached_attributes` will need to be populated. We do this by iterating over all its `dataclass` attributes.

I feel that by introducing this check, we can get rid of the current `if` statement which you said is computationally expensive. The only thing we will have to check is if this way still works when we parse ontology files (which I feel it would).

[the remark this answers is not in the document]
hunterhector: checking `type_attributes` will be preferable to doing class conversion, type checking etc.

forte/data/base_pack.py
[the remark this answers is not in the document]
Pushkar-Bhuse: No I don't think so

tests/forte/data/data_store_test.py
Pushkar-Bhuse: Here you go - https://github.com/asyml/forte/issues/894

forte/data/data_store.py
Pushkar-Bhuse: The reason I mentioned using `get_class` was to validate if the string passed is a path to a valid class. 

Also, with regards to `member_type`, are you saying that during initializing a data store entry, it can be a string rather than an object?

hunterhector: So we don't want to have calls to validate classes during entry creation. This might be slow. Recall that the reason to include a datastore is to reduce class creation overhead. We can't afford new overhead during entry creation.

It is fine to collect the classes and types needed as a one-time process, but we cannot have heavy procedures during `initialize_and_validate_entry`.

You can also notice that in data_store most of the data are stored in primitive forms such as strings, like 
https://github.com/asyml/forte/blob/1fcaebcee6e7ab0ea4e4b8defc8c9192c235d6e5/forte/data/data_store.py#L982

Pushkar-Bhuse: So based on what I understand, there's a need to reduce the overhead introduced by the calls of `self._is_subclass` because it tried to create classes. In that case we can just get rid of this check and just directly store the output of `get_full_module_name` for attributes that are stored in the data store entry as strings.

hunterhector: 1. `_is_subclass` may not create classes, it could use an internal cache. It is supposed to be an efficient implementation. It only checks a string containment in a small dict when the setup is correct.
2. `_is_subclass` is not a tool to be used for general class checking. It is only designed for checking subclass among Forte entries.
3. where do we need to call `get_full_module_name`? We also don't want to call this function if it introduced overhead
4. Instead of `get rid of the check`, the desired approach here is to understand what's being passed in and try to validate this in a quick way. 

I don't think we want to do [this](https://github.com/asyml/forte/blob/1fcaebcee6e7ab0ea4e4b8defc8c9192c235d6e5/forte/data/data_store.py#L1200) ` issubclass(get_class(type_name), cls) for cls in (Link, MultiPackLink)`, this involves a lot of class manipulation. Actually, why not `self._is_subclass(type_name, cls) for cls in (Link, MultiPackLink)`?

We need to have a careful review of the efficiency here, minimizing conversion between class and string etc. @mylibrar could you help monitoring this?

Pushkar-Bhuse: 1, 2: I understand and I will use it accordingly
3: So [here](https://github.com/Pushkar-Bhuse/forte/blob/d8cab15cef6fce6bbcf6ef580200a179e922f503/forte/data/data_store.py#L1201), `attribute_data[0]` stores the value of `self.member_type` which is a `class`. Thus, before storing it in data store, we need to convert it into a string using `get_full_module_name`. This was something we did before as well [here](https://github.com/asyml/forte/blob/e86602f55530a739a8e9d165c8cb51d564d88474/forte/data/entry_converter.py#L90)
4. The reason I feel we can get rid of the check is that this check validates if the `member_type` is a valid `Entry`. But since this value is derived from [here](https://github.com/Pushkar-Bhuse/forte/blob/d8cab15cef6fce6bbcf6ef580200a179e922f503/forte/data/ontology/top.py#L328), it will always be a valid `Entry`

For your last point, you're right, I will update that.

forte/data/data_store.py
```diff
@@ -1044,20 +1084,155 @@ def _create_new_entry(
...
+                entry[payload_id_idx] = 0
+
+        elif any(
+            self._is_subclass(type_name, cls) for cls in (Group, MultiPackGroup)
+        ):
+            members_idx = self.get_datastore_attr_idx(
+                type_name, constants.MEMBER_TID_ATTR_NAME
+            )
+
+            if entry[members_idx] is None:
```
mylibrar, line 1192: Is this already handled by `self._default_attributes_for_type`?

forte/data/data_store.py
```diff
@@ -1857,8 +2081,18 @@ def _parse_onto_file(self):
             entry_name = entry_node.name
             if entry_name in DataStore._type_attributes:
                 continue
+
+            entry_dict = {}
```
mylibrar, line 2085: This is a duplicate implementation of line  2110 - 2112.

tests/forte/remote_processor_test.py
```diff
@@ -82,4 +82,4 @@ def test_remote_processor(self, input_output_pair):
         res: DataPack = test_pl.process("")
         utterance = get_last_utterance(res, "ai")
         self.assertEqual(len([_ for _ in res.get(Utterance)]), 2)
-        self.assertEqual(utterance.text, o_str)
+        self.assertEqual(utterance.text, o_str)
```
mylibrar, line 85: Let's revert any irrelevant changes.

forte/data/ontology/top.py
```diff
@@ -456,59 +397,17 @@ def __init__(
...
-        return self._child
-
-    @child.setter
-    def child(self, val: Tuple):
-        r"""Setter function of ``child``. The update will also be populated
-        into ``DataStore`` in ``self.pack``.
-        """
-        self._child = val
-        self.pack.get_entry_raw(self.tid)[CHILD_TID_INDEX] = val
+        # if parent is not None:
```
mylibrar, line 407: Let's remove them if they are no longer used.

forte/data/ontology/top.py
```diff
@@ -1140,6 +1011,10 @@ class Payload(Entry):
         ValueError: raised when the modality is not supported.
     """
 
+    payload_idx: int
+    modality_name: str
+    uri: Optional[str]
+
```
mylibrar, line 1017: Right now it's okay to not store `cache` as dataclass attributes, but we will need to remove all the `payload` lists in `DataPack` in future. I think it worths another PR. Maybe you can fix this in #844

forte/data/data_store.py
Pushkar-Bhuse: Yes it is, I'll remove it

forte/data/data_store.py
Pushkar-Bhuse: I think the only difference would be that `_init_top_to_core_entries`will add the parent class to `type+attributes immediately. But I can change the code to incorporate it.

forte/data/ontology/top.py
Pushkar-Bhuse: I think this needs to be done in a whole different PR. Mainly because we want to do the same for Audio, Text and Image entries.

mylibrar: Yeah that makes sense.

forte/data/data_store.py
mylibrar: I mean these lines:
```python
entry_dict = {}
entry_dict[constants.PARENT_CLASS_KEY] = set()
entry_dict[constants.PARENT_CLASS_KEY].add(entry_node.parent.name)
```
Why do we need to call them twice?

Pushkar-Bhuse: Oh I got you. I'll fix it

forte/data/ontology/top.py
Pushkar-Bhuse: We can make those changes [here](https://github.com/asyml/forte/issues/847)

forte/data/ontology/top.py
```diff
@@ -268,17 +228,26 @@ class Link(BaseLink):
         child: the child entry of the link.
     """
     # this type Any is needed since subclasses of this class will have new types
-    ParentType: Any = Entry
-    ChildType: Any = Entry
+
+    parent_type: str
+    child_type: str
+    parent: Optional[Entry]
+    child: Optional[Entry]
```
mylibrar, line 235: This is changing the interface. Previously `Link.parent` and `Link.child` are `tid` instead of entry object.

forte/data/data_pack.py
```diff
@@ -1399,11 +1399,14 @@ def _generate_link_entry_data(
                     f"{a_type}"
                 )
 
+            link_parent = cast(Entry, link.parent)
             a_dict["parent"].append(
-                np.where(data[parent_type]["tid"] == link.parent)[0][0]
+                np.where(data[parent_type]["tid"] == link_parent.tid)[0][0]
```
mylibrar, line 1404: Maybe we don't want to change the usage here. `link.parent` should be a `tid` instead of an entry object.

mylibrar (approved):

PR #886, opened 2022-07-22
Storing Attribute Types in `_type_attributes`

### LINKED ISSUE #835: Store attribute type of `Entry`

Pushkar-Bhuse:
This PR fixes #835. 

### Description of changes
This PR introduces a new structure of `_type_attributes` of Data Store. The `_type_attributes` now store the type values of individual attributes of each entry in a Data Store. For example:
```
DataStore._type_attributes = {
"ft.onto.base_ontology.Document": {
                "attributes": {
                    "document_class": {"index": 4, "type": (list, (str,))},
                    "sentiment": {"index": 5, "type": (dict, (str, float))},
                    "classifications": {
                        "index": 6,
                        "type": (FDict, (str, Classification)),
                    },
                },
                "parent_class": set(),
            },
}
```

Through this PR, `getter` methods of `dataclass` attributes could be more comprehensive and return values based on their respective types. This is particularly useful in the case where attributes of type `Entry` are only stored through their `tid`.  Through this PR, it can be easy to distinguish these attributes from the ones stored as `int`.

### Test Conducted
The tests were conducted on `data_store_test`, `data_pack_test`, `data_store_serialization`


forte/data/ontology/ontology_code_generator.py
```diff
@@ -1032,16 +1036,17 @@ def parse_entry(
         property_items, property_names = [], []
         for prop_schema in properties:
             # TODO: add test
-            prop_name = prop_schema["name"]
-            if prop_name in RESERVED_ATTRIBUTE_NAMES:
+            prop = (prop_schema["name"], prop_schema["type"])
```
mylibrar, line 1039: We might want to document this change somewhere.

forte/data/ontology/top.py
```diff
@@ -1261,6 +1267,15 @@ def __setstate__(self, state):
         self.__dict__.update(state)
         self._modality = getattr(Modality, state["_modality"])
 
+        # During de-serialization, convert the list back to numpy array.
+        if "_embedding" in state:
+            state["_embedding"] = np.array(state["_embedding"])
+        else:
+            state["_embedding"] = np.empty(0)
+
+        if "_cache" in state and isinstance(state["_cache"], list):
```
mylibrar, line 1276: This checking might not be stable. Can we assume that `_cache` is a numpy array when it was stored as a list in serialization format?

forte/data/data_store.py
```diff
@@ -286,6 +311,20 @@ def __setstate__(self, state):
...
+        # Update `_type_attributes` to store the types of each
+        # entry attribute as well.
+        for tn in self._type_attributes:
+            entry_type = self._add_entry_types(tn)
+            for attr, type_val in entry_type.items():
+                try:
+                    info_dict = self._type_attributes[tn][
+                        constants.TYPE_ATTR_KEY
+                    ][attr]
+                except KeyError:
```
mylibrar, line 323: Does this mean that `entry_type` has some new attributes that are not present in `self._type_attributes`?

forte/data/data_store.py
```diff
@@ -537,18 +592,24 @@ def _get_type_info(self, type_name: str) -> Dict[str, Any]:
 
         attr_dict = {}
         attr_idx = constants.ENTRY_TYPE_INDEX + 1
+        type_dict = self._add_entry_types(type_name)
+
         for attr_name in attributes:
-            attr_dict[attr_name] = attr_idx
+            attr_dict[attr_name] = {
```
mylibrar, line 598: Let's also update the docstring of `_get_type_info` on this change

forte/data/data_store.py
```diff
@@ -594,16 +655,65 @@ def _default_attributes_for_type(self, type_name: str) -> List:
         attr_dict: Dict = self._get_type_attribute_dict(type_name)
         attr_fields: Dict = self._get_entry_attributes_by_class(type_name)
         attr_list: List = [None] * len(attr_dict)
-        for attr_name, attr_id in attr_dict.items():
+        for attr_name, attr_info in attr_dict.items():
             # TODO: We should keep a record of the attribute class instead of
             # inspecting the class on the fly.
+            attr_id = attr_info[constants.ATTR_INDEX_KEY]
+
             attr_class = get_origin(attr_fields[attr_name].type)
```
mylibrar, line 663: Right now we are still calling `self._get_type_attribute_dict(type_name)` and `self._get_entry_attributes_by_class(type_name)`. Can we get these info directly from `self._type_attributes` or `self._get_type_info()`?

tests/forte/data/data_store_serialization_test.py
```diff
@@ -36,19 +39,29 @@ def setUp(self) -> None:
         DataStore._type_attributes = {
             "ft.onto.base_ontology.Document": {
                 "attributes": {
-                    "sentiment": 4,
-                    "classifications": 5,
+                    "document_class": {"index": 4, "type": (list, (str,))},
```
mylibrar, line 42: We are adding a new attribute. So will there be an error if we don't include `document_class` here?

tests/forte/data/data_store_test.py
```diff
@@ -126,19 +128,34 @@ def setUp(self) -> None:
         self.reference_type_attributes = {
```
mylibrar, line 128: Looks like `data_store_test.py` and `data_store_serialization_test.py` share a lot of resources. Maybe we should just keep single source of truth. Note that they do have difference and some of the difference may be intentional.

forte/common/constants.py
```diff
@@ -38,6 +38,14 @@
 # ``_type_attributes`` of ``DataStore``.
 TYPE_ATTR_KEY = "attributes"
 
+# Name of the key to access the type of an attribute from
+# ``_type_attributes`` of ``DataStore``.
+ATTR_TYPE_KEY = "type"
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Changing `TYPE_ATTR_KEY` to  `ATTR_INFO_KEY`

forte/data/ontology/code_generation_objects.py
```diff
@@ -864,9 +864,9 @@ def collect_parents(self, node_dict: Dict[str, Set[str]]):
             found_node = search(self.root, search_node_name=node_name)
             if found_node is not None:
                 while found_node.parent.name != "root":
-                    node_dict[
-                        found_node.parent.name
-                    ] = found_node.parent.attributes
+                    node_dict[found_node.parent.name] = set(
```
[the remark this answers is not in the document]
Pushkar-Bhuse: That's right

forte/data/ontology/top.py
Pushkar-Bhuse: I think we can. Because if `_cache` is not stored as a string (which is in the case of text), it is stored an an `ndarray`. It can be seen by the definition of `set_cache` [here](https://github.com/asyml/forte/blob/8a61272fe2dd7d3f266f4d66a2f0d3e8f6fc3389/forte/data/ontology/top.py#L1226)

forte/data/data_store.py
Pushkar-Bhuse: Yes

forte/data/data_store.py
Pushkar-Bhuse: so `self._get_type_attribute_dict(type_name)` actually uses `self._get_type_info()`. With regards to `self._get_entry_attributes_by_class(type_name)`, we need this function because it gives us the `Field` object which us used to fetch the type information.

tests/forte/data/data_store_test.py
Pushkar-Bhuse: Should we keep that a part of this PR?

mylibrar: I think it can be another PR and it's not a prority right now.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -811,22 +812,25 @@ def parse_schema(
...
+                property_name = property[0]
                 # Check if the name is allowed.
                 if not property_name.isidentifier():
                     raise InvalidIdentifierException(
                         f"The property name: {property_name} is not a valid "
                         f"python identifier."
                     )
 
-                if property_name in self.allowed_types_tree[en.class_name]:
+                if property_name in set(
```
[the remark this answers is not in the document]
Pushkar-Bhuse: We can't do that because the value of `allowed_types_tree` is updated in the loop

forte/data/ontology/top.py
mylibrar: Cool. Then let's just add the assumption here in the comments since it might not be true in future when we have a more flexible Payload that can stores any type of data.

forte/data/data_store.py
mylibrar: Maybe it's better to add some comments here on why skipping them and when this may happen.

forte/data/ontology/ontology_code_generator.py
```diff
@@ -811,22 +812,25 @@ def parse_schema(
...
+                    val[0] for val in self.allowed_types_tree[en.class_name]
+                ):
                     warnings.warn(
                         f"Attribute type for the entry {en.class_name} "
                         f"and the attribute {property_name} already present in "
                         f"the ontology, will be overridden",
                         DuplicatedAttributesWarning,
                     )
-                self.allowed_types_tree[en.class_name].add(property_name)
+                self.allowed_types_tree[en.class_name].add(tuple(property))
```
hunterhector, line 833: Do we need this `tuple` conversion? The type of `property` should already be a `tuple`

forte/data/ontology/code_generation_objects.py
```diff
@@ -797,7 +797,7 @@ def __init__(self, name: str):
         self.children: List[EntryTreeNode] = []
         self.parent: Optional[EntryTreeNode] = None
         self.name: str = name
-        self.attributes: Set[str] = set()
+        self.attributes: Set[Tuple] = set()
```
hunterhector, line 800: `Tuple` is under specific, what's the structure of this `Tuple`

forte/data/ontology/code_generation_objects.py
```diff
@@ -817,7 +817,7 @@ def add_node(
         self,
         curr_entry_name: str,
         parent_entry_name: str,
-        curr_entry_attr: Set[str],
+        curr_entry_attr: Set[Tuple],
```
hunterhector, line 820: what tuple

forte/data/ontology/code_generation_objects.py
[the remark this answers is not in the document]
hunterhector: so where's the note?

forte/data/base_pack.py
```diff
@@ -455,32 +456,53 @@ def on_entry_creation(
             # Use the auto-inferred control component.
             c = self.__control_component
 
-        def entry_getter(cls: Entry, attr_name: str, field_type):
+        def entry_getter(cls: Entry, attr_name: str):
             """A getter function for dataclass fields of entry object.
             When the field contains ``tid``s, we will convert them to entry
```
hunterhector, line 461: This docstring should be more specific on when they will be convert to entries.

forte/data/data_store.py
```diff
@@ -160,9 +161,16 @@ def __init__(
         ``type_name``, their parent entry, and the order of corresponding attributes.
         The keys are fully qualified names of every type; The value is a
         dictionary with two keys. Key ``attribute`` provides an inner dictionary
-        with all valid attributes for this type and the indices of attributes
-        among these lists. Key ``parent_class`` is a string representing the
-        ancestors of this type.
+        with all valid attributes for this type and the information of attributes
```
hunterhector, line 164: Can we restructure this docstring as bullet points of numbered lists? A bit hard to follow right now.

forte/data/data_store.py
```diff
@@ -160,9 +161,16 @@ def __init__(
...
+        among these lists. This information is represented as a dictionary. The
+        dictionary has two entries, the first is index which determines the position
+        where an attribute is stored in a data store entry. The second is type, which
+        is a tuple of two elements that provides the type information of a given
+        attribute. The first element is the unsubscripted version of
+        the attribute's type and the second element is the type arguments
```
hunterhector, line 170: examples will help this a lot

forte/data/data_store.py
```diff
@@ -174,20 +182,30 @@ def __init__(
...
-            #     "ft.onto.base_ontology.Document": {
-            #       "attributes": {"document_class": 4,
-            #               "sentiment": 5, "classifications": 6},
-            #       "parent_class": set("forte.data.ontology.top.Annotation"), },
-            #     "ft.onto.base_ontology.Sentence": {
-            #       "attributes": {"speaker": 4,
-            #               "part_id": 5, "sentiment": 6,
-            #               "classification": 7, "classifications": 8},
-            #       "parent_class": set(), }
+            #    "ft.onto.base_ontology.Document": {
```
hunterhector, line 185: this is a code-block for python, shouldn't comment it (no need to have the #s).

forte/data/data_store.py
```diff
@@ -594,16 +666,90 @@ def _default_attributes_for_type(self, type_name: str) -> List:
...
+            attributes: This argument is used when parsing ontology
+                files. The entries in the set are a tuples of two
+                elements.
+
+                .. code-block:: python
+
+                    attributes =  {
+                                ('passage_id', 'str'),
+                                ('author', 'str')
+                            }
```
hunterhector, line 715: where's return?

forte/data/data_store.py
```diff
@@ -594,16 +666,90 @@ def _default_attributes_for_type(self, type_name: str) -> List:
...
             if attr_class in (FList, list, List):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = []
             elif attr_class in (FDict, dict, Dict):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = {}
         return attr_list
 
+    def _new_entry_types(
+        self, type_name: str, attributes: Optional[Set[Tuple[str, str]]] = None
+    ) -> Dict[str, Tuple]:
+        r"""This function takes a fully qualified ``type_name`` class name
```
hunterhector, line 684: This function takes two items.

forte/data/data_store.py
```diff
@@ -594,16 +666,90 @@ def _default_attributes_for_type(self, type_name: str) -> List:
...
+        of two elements. The first element is the unsubscripted version of
+        the attribute's type and the second element is the type arguments
+        for the same. The `type_dict` is used to populate the type
+        information for attributes of an entry specified by ``type_name``
+        in `_type_attributes`.
+
+
+        Args:
+            type_name: A fully qualified name of an entry class.
+            attributes: This argument is used when parsing ontology
```
hunterhector, line 706: Can't understand this docstring. When should we provide this?

forte/data/data_store.py
```diff
@@ -594,16 +666,90 @@ def _default_attributes_for_type(self, type_name: str) -> List:
...
+                attr_class = get_origin(attr_info.type)
+                attr_args = get_args(attr_info.type)
+
+                # Prior to Python 3.7, typing.List and typing.Dict
+                # is not converted to primitive forms of list and
+                # dict. We handle them separately here
+                if attr_class == Dict:
+                    attr_class = dict
+                if attr_class == List:
+                    attr_class = list
```
hunterhector, line 747: I am not confident in this function. Are we only handling these types here? What if we are dealing with something new, like `Set`? How can we make sure this function is robust?

forte/data/data_store.py
```diff
@@ -690,6 +836,34 @@ def _is_annotation(self, type_name: str) -> bool:
             for entry_class in (Annotation, AudioAnnotation)
         )
 
+    def get_entry_types(
```
hunterhector, line 839: The function name does not match with its behavior

forte/data/data_store.py
```diff
@@ -594,16 +666,90 @@ def _default_attributes_for_type(self, type_name: str) -> List:
...
+            attr_id = attr_info[constants.ATTR_INDEX_KEY]
+
             attr_class = get_origin(attr_fields[attr_name].type)
             if attr_class in (FList, list, List):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = []
             elif attr_class in (FDict, dict, Dict):
                 attr_list[attr_id - constants.ATTR_BEGIN_INDEX] = {}
         return attr_list
 
+    def _new_entry_types(
```
hunterhector, line 681: what is `new` here in the function name?

forte/data/base_pack.py
```diff
@@ -455,32 +456,53 @@ def on_entry_creation(
...
             """
+
             data_store_ref = (
                 cls.pack._data_store  # pylint: disable=protected-access
             )
             attr_val = data_store_ref.get_attribute(
                 tid=cls.tid, attr_name=attr_name
             )
-            if field_type in (FList, FDict):
+            entry_type = data_store_ref.get_entry_types(
```
hunterhector, line 475: It is not an `entry_type`, it is the type of the attribute. Variable names also affect the quality of the code.

forte/data/base_pack.py
```diff
@@ -455,32 +456,53 @@ def on_entry_creation(
...
+                cls.entry_type(), attr_name
+            )
+
+            if entry_type[0] in (FList, FDict):
```
hunterhector, line 479: not sure if hardcoding `FList` and `FDict` is a good idea. 

We have some constants here. https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_const.py#L86

forte/data/data_store.py
```diff
@@ -258,18 +276,28 @@ def __getstate__(self):
         """
         state = super().__getstate__()
         state["_DataStore__elements"] = {}
+
+        # Make a copy of the updated type_attributes
+        type_attributes = deepcopy(DataStore._type_attributes)
```
hunterhector, line 281: why do we need a deepcopy?

hunterhector:
In the future, for this type of more complex PR's , the better idea is to split out small steps and tackle them each one separately.

forte/data/data_store.py
Pushkar-Bhuse: The example is given below

forte/data/data_store.py
Pushkar-Bhuse: It is because we want to remove only the type information from the type_attributes dict before serialization. However, since a dict is mutable in python, we cannot do this operation on the original `_type_attributes` dict. Thus, we create a copy of this dictionary, remove the type information from that and then use it for serialization. (Here, type information mainly means the dictionary stored by key `type` in the `_type_attributes` for each entry)

forte/data/data_store.py
Pushkar-Bhuse: This is function is invoked when new entries are added to `_type_attributes` and thus their type information will also need to be added.

forte/data/ontology/code_generation_objects.py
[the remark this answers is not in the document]
Pushkar-Bhuse: I'll add it to the docstrings/comments of the individual variables.

forte/data/ontology/top.py
```diff
@@ -1250,6 +1250,12 @@ def __getstate__(self):
         # Entry store is being integrated into DataStore
         state = self.__dict__.copy()
         state["_modality"] = self._modality.name
+
+        if isinstance(state["_cache"], np.ndarray):
+            state["_cache"] = list(self._cache.tolist())
+        if isinstance(state["_embedding"], np.ndarray):
+            state["_embedding"] = list(self._embedding.tolist())
```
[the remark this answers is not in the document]
Pushkar-Bhuse: Because they were numpy arrays, they were causing errors during serialization. That is why we added this check.

tests/forte/data/data_store_serialization_test.py
```diff
@@ -73,39 +86,39 @@ def setUp(self) -> None:
...
+                        {"Negative": 0},
                     ],
                     [
                         10,
                         25,
                         3456,
                         "ft.onto.base_ontology.Document",
                         "Negative",
-                        "Class B",
+                        {},
```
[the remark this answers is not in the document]
Pushkar-Bhuse: I was thinking of keeping the data store entry structure consistent with the type information provided in `_type_attributes`. But I can change it back to the original form.

forte/data/ontology/top.py
```diff
@@ -1261,6 +1267,20 @@ def __setstate__(self, state):
         self.__dict__.update(state)
         self._modality = getattr(Modality, state["_modality"])
 
+        # During de-serialization, convert the list back to numpy array.
+        if "_embedding" in state:
+            state["_embedding"] = np.array(state["_embedding"])
+        else:
+            state["_embedding"] = np.empty(0)
```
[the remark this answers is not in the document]
Pushkar-Bhuse: `_embedding` was being stored an an `np.array` which was causing problems during serialization.

forte/data/data_store.py
Pushkar-Bhuse: So this is an issue caused by the change in the typing module from python 3.6 to python 3.7. Another workaround for this is that we can let the the output `get_origin` be stored in any way depending on the python version and just handle this in the test cases. This is similar to a suggestion made [here](https://github.com/ilevkivskyi/typing_inspect/issues/27)

forte/data/base_pack.py
Pushkar-Bhuse: Can I add constants here?

forte/data/data_store.py
Pushkar-Bhuse: It is because we want to only remove the type information atom the `_type_attributes` before serialization without affecting the class variable. Since `dict` is mutable, we create a copy of `_type_attributes` which will be modified and then used during serialization.

forte/data/base_pack.py
Pushkar-Bhuse: I'm not sure if we can use constants from here in base_pack because it makes a partial import of `PackType` from base_pack itself.

Pushkar-Bhuse: I have instead added a constant in `core.py` where these classes are defined.

forte/data/data_store.py
hunterhector: it is pretty hard to understand this section now

forte/data/data_store.py
hunterhector: But how can we make sure we we list all these cases, how about `Set`?

forte/data/data_store.py
hunterhector: 1. Why is type_attribute information new when an entry is added? Some entries could have the same type
2. Still, I don't think anything is "new", the information is stored and get via `self._get_entry_attributes_by_class(type_name)`, so this function only `get` this information

forte/data/base_pack.py
hunterhector: they should be better in the same place and nearby, probably here: https://github.com/asyml/forte/blob/master/forte/data/ontology/ontology_code_const.py#L86

Pushkar-Bhuse: As I mentioned above, if we put it there we won't be able to use it in `base_pack`

forte/data/data_store.py
Pushkar-Bhuse: So this function is invoked in the case where it is the first time an entry of this type is being created (and thus it is not present in `_type_attributes`. But I do get your point, I think a better name for this function could be `fetch_entry_type_data`. Would this be okay?

forte/data/data_store.py
Pushkar-Bhuse: Oh, I'll try explaining it better. Is there any area specifically that you feel needs to change?

forte/data/data_store.py
Pushkar-Bhuse: I have added a more general solution in the updated version of this PR which does not require modifying test cases

forte/data/data_store.py
hunterhector: let's use the new name, since the "add" action seems to happen after this.

forte/data/base_pack.py
hunterhector: huh, let's see whether this workout.

forte/data/data_store.py
hunterhector: so this variable will go to serialized data. I normally prefer the other way around where we assemble the serialized data instead of removing stuff from a dict. One reason is that we have more contorl over the data being serialized

Pushkar-Bhuse: Yes I actually had started off with your approach, but I noticed that it made the code a bit large unnecessarily because the part to remove (type information) is much lesser than the part to keep. But if you want I can try the other way round as well.

forte/data/data_store.py
Pushkar-Bhuse: Sounds good

forte/data/data_store.py
hunterhector: ok, btw. is it necessary to remove the type information?

Pushkar-Bhuse: Yes, because they are not stored in serializable formats

forte/data/ontology/code_generation_objects.py
```diff
@@ -856,17 +856,19 @@ def collect_parents(self, node_dict: Dict[str, Set[str]]):
 
         Args:
             node_dict: the nodes dictionary of nodes to collect parent nodes
-                for.
+                for. The entry represented by nodes in this dictionary do not store
```
mylibrar, line 859: The entry node actually stores the type info of attributes according to your latest changes.

forte/data/data_store.py
```diff
@@ -157,12 +164,22 @@ def __init__(
...
         """
         The ``_type_attributes`` is a private dictionary that provides
-        ``type_name``, their parent entry, and the order of corresponding attributes.
-        The keys are fully qualified names of every type; The value is a
-        dictionary with two keys. Key ``attribute`` provides an inner dictionary
-        with all valid attributes for this type and the indices of attributes
-        among these lists. Key ``parent_class`` is a string representing the
-        ancestors of this type.
+        ``type_name``, their parent entry, and the metadata of its corresponding
+        attributes. This metadata includes the order and type information of attributes
```
mylibrar, line 168: Feels like it would be better to introduce key and value before explaining the metadata of attributes.

forte/data/ontology/code_generation_objects.py
Pushkar-Bhuse: So `node_dict` is another component just used in `collect_parents`, this does not store type information because we do [this](https://github.com/Pushkar-Bhuse/forte/blob/39b4de96d625ea3231b35432847bc2f467b27aaa/forte/data/ontology/code_generation_objects.py#L870). On the other hand, [here](https://github.com/Pushkar-Bhuse/forte/blob/39b4de96d625ea3231b35432847bc2f467b27aaa/forte/data/data_store.py#L1852) we use the node with its index and type information and I've mentioned it in the comments.

forte/data/data_store.py
Pushkar-Bhuse: So key and value here refer to the components of Python `dict`. Do you mean mentioning that they key-value refer to that of a Python `dict`?

mylibrar: I'm just thinking moving this sentence `The keys are fully qualified ...` before `This metadata includes the order ...`

Pushkar-Bhuse: Got it

mylibrar: The docstring still looks a little bit off.

> The ``_type_attributes`` is a private dictionary that provides ``type_name``, ie. their parent entry (represented by the fully qualified name of the entry) as the key, and the

The keys are not parent_entry. And ``type_name`` is different from ``their parent entry``.

Pushkar-Bhuse: How about - "The ``_type_attributes`` is a private dictionary that contains the parent entry ( (represented by the fully qualified name of the entry) as key and the metadata ...." ?

mylibrar: I still don't think `parent entry` is a key of `_type_attributes`. `parent_class` may be a key of `_type_attributes[type_name]` but it's not a key for `_type_attributes`.

mylibrar (approved):

Issue #873, opened 2022-07-11
`DataPack` image getter and setting function


hepengfe:
**Is your feature request related to a problem? Please describe.**
Currently, we don't have `DataPack.set_image()` that can simplify the DataPack API to set image payload cache data.

**Describe the solution you'd like**
* `DataPack.set_image()` that takes image data and set the cache data of image payload at the specified index.
* `DataPack.image` that getting image data from `ImagePayload`

**Describe alternatives you've considered**

**Additional context**

PR #891, opened 2022-07-25
Image getter setter

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/873. 

### Description of changes
* `DataPack.set_image()` that takes image data and set the cache data of image payload at the specified index.
* `DataPack.add_image()` that takes image data and add new `ImagePayload` containing the image.
* `DataPack.image` that getting image data from `ImagePayload`

* it's more convenient for users to get and set image data in data pack.

### Test Conducted
* test setting index beyond datapack payloads length
* test get image from datapack


forte/data/ontology/top.py
```diff
@@ -891,6 +891,7 @@ def max_y(self):
 
     @property
     def image_shape(self):
+        self.pack.get_payload_data_at(Modality.Image, self._image_payload_idx)
```
mylibrar, line 894: Let's note down the purpose of this line. Are we trying to validate some conditions or to load the image size?

tests/forte/image_annotation_test.py
```diff
@@ -42,6 +43,21 @@ def setUp(self):
         ip.set_cache(self.line)
         ImageAnnotation(self.datapack)
 
+    def test_datapack_image_operation(self):
+        datapack = DataPack("image2")
+        datapack.set_image(self.line, 0)
```
hunterhector, line 48: looking at these test cases, I am wondering how do we add a next payload? Do we need to create a `add_image` function?

hepengfe: set_image can add next payload but having `-1` as the index

forte/data/data_pack.py
```diff
@@ -610,6 +619,32 @@ def set_audio(
...
+        object.
+
+        Args:
+            image: A numpy array storing the image.
+            image_payload_index: the zero-based index of the ImagePayload
+                in this DataPack's ImagePayload entries. Defaults to 0.
+        """
+        # temporary solution for backward compatibility
+        # past API use this method to add a single image in the datapack
+        if len(self.image_payloads) == 0 and image_payload_index == 0:
```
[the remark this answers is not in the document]
hepengfe: do we record some class variables whether `self.image_payloads` is empty?

forte/data/data_pack.py
```diff
@@ -610,6 +619,32 @@ def set_audio(
...
+            from ft.onto.base_ontology import (  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
hepengfe: Actually, we might not want to move it to top level it makes the dependencies more complex and it has import cycle issue as we discussed before. We also don't want to create `Payload` with modalities in the `top.py`.

forte/data/ontology/top.py
hepengfe: https://github.com/asyml/forte/pull/893

update the function in this PR

forte/data/data_pack.py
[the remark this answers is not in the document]
mylibrar: Okay. So is [this](https://github.com/asyml/forte/pull/859/files#diff-b74f73aea0aeb3c17371323001cb91e0ab0001e961f610e90184f5905c636dfcR1304) just a temporary solution?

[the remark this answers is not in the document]
hepengfe: OK. I think I missed the solution. I thought what we are having right now is the final solution. I think you are right and I will edit this part in https://github.com/asyml/forte/pull/859

forte/data/data_pack.py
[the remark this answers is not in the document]
hunterhector: So all entry data should be stored in data store, we won't have` self.image_payloads` anymore.

tests/forte/image_annotation_test.py
hunterhector: This usage looks a bit tricky though, and does this behavior applies to all the payloads? We should make them the same.

hepengfe: Currently, it applies to payloads of all modalities. I can make another function if that's more user-friendly.

hunterhector: yeah, we need to have a natural way to add image, `datapack.set_image(self.line, -1)` is hard to figure out (we also don't have documentation for it).

```
datapack.set_image(image, payload=x) # set the base payload, 
datapack.add_image(image) # add another payload and set the image. This should be the future interface, it starts by assuming there is no image payload in the data pack (i.e. it won't assume there is an index=0 payload).
```

forte/data/data_pack.py
```diff
@@ -610,6 +639,36 @@ def set_audio(
...
+            image_payload_index: the zero-based index of the ImagePayload
+                in this DataPack's ImagePayload entries. Defaults to 0.
+        """
+        # temporary solution for backward compatibility
+        # past API use this method to add a single image in the datapack
+        if (
+            self._data_store.num_entries("ft.onto.base_ontology.ImagePayload")
+            == 0
+            and image_payload_index == 0
+        ):
```
hunterhector, line 661: instead of only providing the comment, we should provide a log saying for backward compatibility we are setting the first image when there is nothing and image_payload_index=0, however, we should encourage users to use `add_image` function.

Let's add these notes to both `image` and `audio`, and avoid using `set_image` and `set_audio` in this way (only use them when the index actually exists) in our own tutorials/examples.

For `set_text`, since it is widely used in the repo, let's not add the logging to it for now.

hepengfe: Added logging info and `add_image` function.

mylibrar: Maybe we should also add these notes to the `set_audio` function. And do we need an `add_audio` function as well?

hepengfe: https://github.com/asyml/forte/issues/895

mylibrar (approved):

Issue #890, opened 2022-07-25
DataPack setting for Modality data should support setting data in a new payload


hepengfe:
**Is your feature request related to a problem? Please describe.**
Current datapack setting for modality data only can add new payload for the first one. We expect setter can add payload to a new index.

For example, suppose we have 2 text payloads with text payload indices 0, 1. We should support `datapack.set_text("text", 2)`.

**Describe the solution you'd like**
* we should support `datapack.set_text("text", -1)` which is equivalent to append text payload
* we should also support `datapack.set_text("text")` which the text payload index is by default -1

**Describe alternatives you've considered**

**Additional context**

PR #892, opened 2022-07-26
Payload data setter

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/890

### Description of changes
* support datapack.set_text("text", -1) which is equivalent to append text payload
* support datapack.set_text("text") which the text payload index is by default 0. When there is no text payloads in the datapack, it will add a new payload.

* user can set/add data of modality in the DataPack easily

### Test Conducted
* it's backward compatible with past tests


forte/data/data_pack.py
```diff
@@ -244,6 +244,15 @@ def audio(self):
         r"""Return the audio of the data pack"""
         return self.get_payload_data_at(Modality.Audio, 0)
 
+    @property
+    def image(self):
+        r"""Return the image of the data pack"""
```
hunterhector, line 249: could update the docstring

forte/data/data_pack.py
```diff
@@ -244,6 +244,15 @@ def audio(self):
...
+        return self.get_payload_data_at(Modality.Image, 0)
+
+    def get_image(self, index: int):
+        r"""Return the image at the given index"""
```
hunterhector, line 253: update docstring with parameters

forte/data/data_pack.py
```diff
@@ -546,13 +555,19 @@ def set_text(
...
+            ValueError: raised when the text payload index is out of range.
 
         Args:
             text: a str text.
             replace_func: function that replace text. Defaults to None.
             text_payload_index: the zero-based index of the TextPayload
-                in this DataPack's TextPayload entries. Defaults to 0.
+                in this DataPack's TextPayload entries.
+                If it's -1, a new TextPayload will be added to the DataPack.
+                If it's 0, it adds a new TextPayload if there is no text payload in the data pack.
```
hunterhector, line 570: The explanation for the `0` part is a bit hard to understand

forte/data/data_pack.py
```diff
@@ -585,31 +608,79 @@ def set_audio(
...
+        Raises:
+            ValueError: raised when the audio payload index is out of range.
 
         Args:
             audio: A numpy array storing the audio waveform.
             sample_rate: An integer specifying the sample rate.
             audio_payload_index: the zero-based index of the AudioPayload
-                in this DataPack's AudioPayload entries. Defaults to 0.
+                in this DataPack's AudioPayload entries. Defaults to -1, and
+                it adds a new audio payload if there is no audio payload in the data pack.
```
hunterhector, line 624: docstring different from `set_text`. The idea you have is interesting, probably want to make it more intuitive to users.

hunterhector (commented):
interesting idea, could consider improving the docstring, or maybe re-think the behavior so it won't make people confused.

forte/data/data_pack.py
hepengfe: It's a bit tricky indeed because I designed so for backward compatibility. Past `set_text` set/add the the first text payload depending on the condition of the `DataPack.text_payloads`.
* if `DataPack.text_payloads` is empty, it adds a text payload
* if `DataPack.text_payloads` is not empty, it reset the data in the first text payload

What I wanted to is only reset the first text payload and throw error when `DataPack.text_payloads`  is empty. Maybe I can create another issue for it and it could be fixed later together with all test cases modifications.

hunterhector: If we throw an error when `text_payloads` is empty, does it mean that some previous usage will break?

Let's probably keep backward compatibility here.

hepengfe: Yes throwing error will break the past test cases. The current implementation ensures backward compatibility.

mylibrar (approved):

Issue #880, opened 2022-07-14
ImagePayload initialized with image size


hepengfe:
**Is your feature request related to a problem? Please describe.**
Current `ImageAnnoatation` get image size by get the datapack's image at corresponding payload and then get the shape of image by a fixed function `image_data.shape`. It has an assumption that `image_data` is a numpy array which is not always the case. ImagePayload could store image object from other third-party library such as pillow image object or cv2 image object. Also, their shape order could be (width, height) or  (height, width). Therefore, we want to store image size in ImagePayload for easier access for `ImageAnnoatation`.

Therefore, it's better to initialize width and height ahead to ensure the consistency of `ImageAnnoatation`'s functions. 

**Describe the solution you'd like**
* add parameters `image_height` and `image_width` in `ImageAnnotation.__init__`
* adjust init functions in `ImageAnnotation`'s child classes accordingly
* remove `ImageAnnotation.set_image_shape`

**Describe alternatives you've considered**
* get the image size by `ImageAnnotation.pack.get_payload().cache_shape`

Third third possible implementation is we let `ImagePayload` to return the meta data of image. 

**Additional context**

hepengfe:
Since it's redundant to initialize image shape in `ImageAnnotation`. The alternative method is preferred.

PR #893, opened 2022-07-27
Image size access from ImageAnnotation

hepengfe:
This PR fixes https://github.com/asyml/forte/issues/880.

### Description of changes
* implement a function in image annotation to get the image size by `ImageAnnotation.pack.get_payload().cache_shape`

* `ImageAnnotation` can access image size that it operates on.
* all payloads can be initialized with data shape
    For text data, it is the length of the text.[length, text_embedding_dim]
    For audio data, it is the length of the audio. [length, audio_embedding_dim]
    For image data, it is the shape of the image. [height, width, channel]

### Test Conducted


forte/data/ontology/top.py
```diff
@@ -883,26 +883,41 @@ def image(self):
...
-            height
-        )
+        if self.pack is None:
+            raise ValueError(
+                "Cannot get image because image annotation is not "
+                "attached to any data pack."
+            )
+        image_shape = self.pack.get_payload_at(
+            Modality.Image, self._image_payload_idx
+        )._cache_shape
```
mylibrar, line 913: Maybe it's better to use `image_shape` here.

forte/data/ontology/top.py
```diff
@@ -1219,15 +1235,36 @@ def payload_index(self) -> int:
...
+                on the modality and its length.
+                For text data, it is the length of the text.[length, text_embedding_dim]
+                For audio data, it is the length of the audio. [length, audio_embedding_dim]
+                For image data, it is the shape of the image. [height, width,
+                channel]
+                For example, for image data, if the cache_shape length is 2, it
+                means the image data is a 2D image [height, width].
         """
         self._cache = data
+        if isinstance(data, np.ndarray):
```
mylibrar, line 1263: Looks like we can also set `cache_shape` when `data` is a string.

forte/data/ontology/top.py
hepengfe: do you mean change the variable name? `cache_shape` -> `image_shape`. `cache_shape` is more generic variable for payloads of all modalities.

mylibrar: No you don't need to change the variable name. I think `Payload` already exposed a property function called `image_shape` right?

forte/data/ontology/top.py
hepengfe: fixed it.

docs/notebook_tutorial/ocr.ipynb
```diff
@@ -80,7 +89,8 @@
     "        # Read in Image data and store in DataPack\n",
     "        # add Image payload into DataPack.payloads\n",
     "        image_data = Image.open(requests.get(data_source, stream=True).raw)\n",
-    "        pack.set_image(image_data, payload_idx)\n",
+    "        width, height = image_data.size\n",
+    "        ip.set_cache(image_data, [height, width])\n",
```
mylibrar, line 93: Do we need to explicitly set the shape here? `set_cache` can automatically infer the image shape when the input data is `ndarray`.

mylibrar (approved):

PR #899, opened 2022-08-23
Remove class objects from DataStore

### LINKED ISSUE #898: Remove class objects from `DataStore`
`DataStore` should only store primitive types as well as built-in list or dict. However, we still find some class objects being stored into `DataStore`. These structures are usually not JSON serializable. We need to remove them and maybe add some conversion layer if necessary.

Class objects to be removed from `DataStore`:
* `FNdArray`: We will need to refactor `FNdArray` for conversion.
* `Span`

**Additional context**
* This is part of the data efficiency project
* This PR should be made to the master branch.

mylibrar:
This PR fixes #898 . 

### Description of changes
* Refactor `FNdArray` and add conversion layer to `entry_getter` and `entry_setter` in `BasePack.on_entry_creation`.
* Remove deprecated `image_annotations` in `DataPack`.
* Remove redundant `__getstate__` and `__setstate__` methods in `Annotation` and `AudioAnnotation`.
* Serialize `Span` in `replace_back_operations` and `processed_original_spans` from Payload to make it JSON serializable.
* Revert `get_parent` and `get_child` in `MultiPackLink`

### Test Conducted
* ndarray_attribute_test.py (test `FNdArray` conversion)
* html_reader_test.py (test `Span` conversion)


forte/data/ontology/core.py
```diff
@@ -446,30 +408,33 @@ class FNdArray:
...
-        )
-        self._data: Optional[np.ndarray] = None
-        if dtype and shape:
-            self._data = np.ndarray(shape, dtype=dtype)
 
     @property
     def dtype(self):
-        return self._dtype
+        dtype = self.__data_ref[0]
+        return np.dtype(dtype) if dtype is not None else dtype
```
hunterhector, line 424: `else dtype` simply returns `None`?

forte/data/ontology/core.py
```diff
@@ -446,30 +408,33 @@ class FNdArray:
...
-        self, dtype: Optional[str] = None, shape: Optional[List[int]] = None
+        self,
+        dtype: Optional[str] = None,
+        shape: Optional[List[int]] = None,
+        data_ref: Optional[List] = None,
     ):
         super().__init__()
-        self._dtype: Optional[np.dtype] = (
-            np.dtype(dtype) if dtype is not None else dtype
+        self.__data_ref: List = (
```
hunterhector, line 417: looks like we convert the 3 values into one list `data_ref`? Any reasons for this?

Also, it seems `dtype` and `shape` contain the same information as `data_ref` in the arguments.

forte/data/ontology/top.py
```diff
@@ -1101,6 +1065,15 @@ def __getstate__(self):
         if isinstance(state["_embedding"], np.ndarray):
             state["_embedding"] = list(self._embedding.tolist())
 
+        state["replace_back_operations"] = [
+            ((span.begin, span.end), text)
+            for span, text in self.replace_back_operations
+        ]
+        state["processed_original_spans"] = [
+            ((span1.begin, span1.end), (span2.begin, span2.end))
+            for span1, span2 in self.processed_original_spans
```
hunterhector, line 1074: Would it be easier if we change the `__setstate__` `__getstate__` for the `Span` class?

forte/data/base_pack.py
```diff
@@ -515,44 +516,47 @@ def entry_getter(cls: Entry, attr_name: str):
...
-                    and all(isinstance(element, int) for element in attr_val)
-                    and hasattr(cls.pack, "get_subentry")
-                ):
-                    # Multi pack entry
-                    return cls.pack.get_subentry(*attr_val)
-            except KeyError:
-                pass
+            elif attr_type[1] == (FNdArray,):
+                # Generate FNdArray object on the fly
+                return FNdArray(data_ref=attr_val)
```
hunterhector, line 521: Maybe this is the reason that the interface of `FNdArray` is a bit weird? Can we use kwargs to support this?

forte/data/ontology/core.py
mylibrar: Yes. The reason is to store only single source of truth for `FNdArray` inside `DataStore`. Each `FNdArray` attribute will be stored as a list of `data_ref=[dtype, shape, data]`. All the getter and setter method of `FNdArray` object should now be directed into the reference of that list in `DataStore`.

`dtype` and `shape` are still kept for backward compatibility.

forte/data/ontology/core.py
mylibrar: Yes. This is consistent with the previous behavior 
```python
self._dtype: Optional[np.dtype] = (
    np.dtype(dtype) if dtype is not None else dtype
)
```

forte/data/base_pack.py
```diff
@@ -630,6 +634,12 @@ def entry_setter(cls: Entry, value: Any, attr_name: str):
                         "which can only accept a mapping whose values are "
                         "`Entry` objects."
                     ) from e
+            elif attr_type[1] == (FNdArray,):
```
Pushkar-Bhuse, line 637: Would there be a scenario where the type of an attribute is just `FNdArray`? In that case we should check `attr_type[0]` as well.

forte/data/data_store.py
```diff
@@ -302,6 +302,7 @@ def __getstate__(self):
         state.pop("_DataStore__tid_ref_dict")
         state.pop("_DataStore__tid_idx_dict")
         state.pop("_DataStore__deletion_count")
+        state.pop("_type_attributes", None)
```
Pushkar-Bhuse, line 305: I think it might be useful to mention the instance when the data store state has the `_type_attributes` key and why we need to remove it.

forte/data/ontology/core.py
hunterhector: a single source of truth internally is fine. But do we need to change the interface from separate attributes `dtype` and `shape` to `data_ref`? 

During usage, a user would still prefer the more meaningful full `shape` rather than an ambiguous `data_ref`. 

By using kwargs, I actually mean when we are calling the constructor, we should can pass all arguments as kwargs to the constructor, not adding one additional kwargs, i.e. doing something like `FNdArray(**data_ref)`, which is simply `FNdArray(**{shape=x, dtype=y})`

hunterhector: basically, we don't need to introduce one additional `kwargs` or `data_ref`. Notice `FNdArray` is already a user-facing interface, which is something we should keep clean and logical

mylibrar: Then maybe it's better to keep the interface as `FNdArray(shape=x, dtype=y)`. Meanwhile we can also add a new internal method called `FNdArray.set_data_ref(data_ref)` to pass the reference.

hunterhector: what is the reference here? I only saw `self.__data_ref = [dtype, shape, None]` in the constructor here.

mylibrar: It's also a list of `[dtype, shape, data]` but it's stored inside `DataStore` and can be retrieved with `DataStore.get_attributes`. `set_data_ref` will overwrite the list created in constructor
```python
def set_data_ref(data_ref):
    self. __data_ref = data_ref
```
so that `self. __data_ref` is now linked to a list stored in `DataStore`.

forte/data/base_pack.py
mylibrar: Yeah. It's now updated to `attr_type == (type(None), (FNdArray,))`

forte/data/data_store.py
mylibrar: Explanations added to the docstring

Issue #881, opened 2022-07-14
A new set of `DataPack` interfaces to add/get/delete entries without class objects


mylibrar:
**Is your feature request related to a problem? Please describe.**
Now that we have introduced `DataStore` into Forte, `DataPack` is no longer required to handle or maintain entry class objects. However, the current interface still require users to handle entry objects (e.g., `delete_entry`, `add_entry`, `get_entry`). We should design a new set of interfaces to add/get/delete entries and they should interact with entry data with only consists of primitive types and built-in list/dict.

**Describe the solution you'd like**
* `DataPack` should accept `onto_file_path` and be able to parse it using `DataStore`. This can eliminate the needs for users to import all the entry classes.
* Design schema for a more user-friendly representation of entry data. Currently we store entry as list of attributes in `DataStore`, which is not very convenient for users to use.
* Implement `DataPack.add_entry_raw`, `DataPack.get_entry_raw`, `DataPack.delete_entry_raw`, etc. These new interfaces only interact with the new entry representation that you designed in last step.

**Describe alternatives you've considered**
We can compare and discuss the trade-offs of different designs.

**Additional context**
This is blocked by #874

mylibrar:
Hi @Pushkar-Bhuse, did you have a second PR for this issue? #900 is already merged and you can now push the rest of the codes. Thanks!

PR #900, opened 2022-08-24
Interface to fetch entries in primitive types from `DataPack`

Pushkar-Bhuse:
This PR is the first step towards fixing #881 

### Description of changes
Current, when fetching entries from a `DataPack` or `MultiPack` using the `get` method, Forte converts data store entries into object form. We wanted a way for users to directly interact with `DataStore` entries. In this PR, we provide a modification to the `get` method of `DataPack` to be able to return an entry in its primitive form directly from `DataStore` without needing to be converted to an object.

Additionally, since `DataStore` entries are not very interpretable (since they are in a `list` format), this PR introduces a way to retain data store entries in their primitive form and also represent them in a more interpretable way by converting it to a `dictionary`. This happens by the `transform_data_store_entry` method in `data_store.py`. An example of this is as follows:
```
# Entry of type 'ft.onto.base_ontology.Sentence'
            data_store_entry = [
                171792711812874531962213686690228233530,
                'ft.onto.base_ontology.Sentence',
                0,
                164,
                0,
                '-',
                0,
                {},
                {},
                {}
            ]

            transformed_entry = pack.transform_data_store_entry(
                data_store_entry
            )

            # transformed_entry = {
            #   'begin': 0,
            #   'end': 164,
            #   'payload_idx': 0,
            #   'speaker': '-',
            #   'part_id': 0,
            #   'sentiment': {},
            #   'classification': {},
            #   'classifications': {},
            #   'tid': 171792711812874531962213686690228233530,
            #   'type': 'ft.onto.base_ontology.Sentence'}
            # }
```
 
By allowing `DataPack` or `MultiRack` to fetch entries in their primitive form, users can interact with `DataStore` more easily.

### Test Conducted
The working of the `get` method with the `get_raw` attribute set to `True` was tested in `data_pack_test.py` and `multi_pack_test.py`


forte/data/base_pack.py
```diff
@@ -791,13 +791,13 @@ def get_ids_by_creator(self, component: str) -> Set[int]:
         return entry_set
 
     def is_created_by(
-        self, entry: Entry, components: Union[str, Iterable[str]]
+        self, entry_tid: int, components: Union[str, Iterable[str]]
```
mylibrar, line 794: For public interfaces, I think it's better to take care of backward compatibility.  For example, maybe we can make it `Union[Entry, int]`.

forte/data/data_pack.py
```diff
@@ -1566,11 +1569,25 @@ def require_annotations(entry_class=Annotation) -> bool:
         # If we don't have any annotations but the items to check requires them,
         # then we simply yield from an empty list.
         if (
-            len(self.annotations) == 0
+            len(
```
mylibrar, line 1572: You can also use `self.num_annotations`/`self.num_audio_annotations`. However, there might be a more efficient way to do this since we are only checking if it's empty.

mylibrar (commented):
There might be a few more places to refactor in the current `get` method:
* `entry_type` can just be a string. This will save the cost of `get_full_module_name`
* `range_annotation` could be a `tid` instead of an entry object.

To maintain backward compatibility, there is a potential workaround:
* We may add a new `get_raw` method with the following signature
  ```python
  def get_raw(  # type: ignore
      self,
      entry_type: str,
      range_annotation_tid: Optional[int] = None,
      components: Optional[Union[str, Iterable[str]]] = None,
      include_sub_type: bool = True,
  ) -> Iterator[List]
  ```
* And then in the `get` method, we can do:
  ```python
  def get(  # type: ignore
        self,
        entry_type: Union[str, Type[EntryType]],
        range_annotation: Optional[Union[Annotation, AudioAnnotation]] = None,
        components: Optional[Union[str, Iterable[str]]] = None,
        include_sub_type: bool = True
    ):
      # Convert entry_type to string if it's Type[EntryType]
      ...
      # Convert range_annotation to tid
      ...
      # Convert result from get_raw to entry objects
      for entry_data in get_raw(...):
        yield self.get_entry(tid=entry_data[TID])
  ```

You can try out other solutions as well.

forte/data/data_pack.py
Pushkar-Bhuse: I think `num_annotations` should be good because we want to know the number of entries in the subclasses too

forte/data/data_pack.py
```diff
@@ -1539,38 +1542,57 @@ def get(  # type: ignore
...
+            )
 
         def require_annotations(entry_class=Annotation) -> bool:
-            if issubclass(entry_type_, entry_class):
+            if self._data_store._is_subclass(entry_type_, entry_class):
                 return True
-            if issubclass(entry_type_, Link):
+
+            curr_class: Type[EntryType] = as_entry_type(entry_type_)
+            if issubclass(curr_class, Link):
```
mylibrar, line 1580: I think we should have some way to support getting `link.childtype, link.parenttype, or group.membertype` without converting `entry_type_` back to class.

forte/data/data_pack.py
```diff
@@ -1597,28 +1619,58 @@ def require_annotations(entry_class=Annotation) -> bool:
...
+                    entry = self._data_store.transform_data_store_entry(
+                        entry_data
                     )
-                ):
-                    continue
+                else:
+                    entry = self.get_entry(tid=entry_data[TID_INDEX])
+
+                    # Filter out incompatible audio span comparison for Links and Groups
+                    if (
```
mylibrar, line 1664: Maybe this filtering also applies to cases when `get_raw==True`?

forte/data/data_pack.py
```diff
@@ -1597,28 +1619,58 @@ def require_annotations(entry_class=Annotation) -> bool:
             yield from []
             return
 
+        # If range_annotation is specified, we record its begin and
+        # end index
+        range_begin: int
+        range_end: int
+
+        if range_annotation is not None:
+            if isinstance(range_annotation, (Annotation, AudioAnnotation)):
```
mylibrar, line 1628: We can also use `_is_subclass` here to replace `isinstance` checking.

forte/data/data_pack.py
Pushkar-Bhuse: Yes it does. But I felt that this change is currently specific to entries who's objects exist. Changing methods like `in_span` and `in_audio_span` would make this PR a little big. That's why I felt like we could modify `in_span` and `in_audio_span` in another PR.

mylibrar: Okay then let's create an issue for that

forte/data/data_pack.py
Pushkar-Bhuse: One way that I think we can do it is by storing default values of attributes , or storing class variables in the `type_attributes` dict of a given entry at the time of creation. Should I try it for this PR?

forte/data/data_pack.py
Pushkar-Bhuse: Done: #901

forte/data/data_pack.py
Pushkar-Bhuse: For that `range_annotation` should have been a `str`. But we already have `range_annotation` as an object in this case so we can use `isinstance`

forte/data/data_pack.py
mylibrar: Maybe it can addressed in another PR

hunterhector:
quick comment on the title, not "fetch entries directly from Data Store", but fetch primitive types from data pack. Data store is still invisible to users.

forte/data/base_pack.py
```diff
@@ -806,8 +806,10 @@ def is_created_by(
         if isinstance(components, str):
             components = [components]
 
+        entry_tid = entry.tid if isinstance(entry, Entry) else entry
```
hunterhector, line 809: maybe it is easier/faster if we don't check `entry` or `int` here, but create different functions so the caller can decide which to call.

forte/data/data_pack.py
hunterhector, line 1580: some using `issubclass` while some using `self._data_store._is_subclass`?

forte/data/data_pack.py
hunterhector, line 1628: We might want to create a type of joint type for `AnnotationLike` so we don't have to hard-code.

forte/data/data_store.py
```diff
@@ -1315,6 +1329,111 @@ def _get_existing_ann_entry_tid(self, entry: List[Any]):
...
+        understandable to users. Data Store entries are stored as lists and
+        are not very easily understandable. This method converts ``DataStore``
+        entries from a list format to a dictionary based format where the key
+        is the names of the attributes of an entry and the value is the values
+        corresponding attributes in the data store entry.
+        For example:
+
+        .. code-block:: python
+
+            # Entry of type 'ft.onto.base_ontology.Sentence'
```
hunterhector, line 1382: The more conventional way to write this docstring is the `doctest` format (https://docs.python.org/3/library/doctest.html). We actually are testing the correctness of these automatically here: https://github.com/asyml/forte/blob/master/.github/workflows/main.yml#L99

forte/data/ontology/top.py
```diff
@@ -515,7 +514,7 @@ class MultiPackGroup(MultiEntry, BaseGroup[Entry]):
     of members.
     """
     member_type: str
-    members: Optional[FList[Entry]]
+    members: List[Tuple[int, int]]
```
hunterhector, line 517: Hmm, I thought we won't change the interface the users see. @mylibrar any idea here?

forte/data/base_pack.py
Pushkar-Bhuse: So with regards to Forte, the argument entry can be converted to just `int` which represents `tid`. We only keep it `Union[Entry, int]` to ensure backward compatibility.

Also, I feel that making two separate functions would introduce some redundancy. Should I still make two separate functions?

forte/data/data_pack.py
Pushkar-Bhuse: Yes that is because `curr_class` is an `EntryType` object. `self._data_store._is_subclass` requires a `str`. The reason this is done is because we don't currently have a way of getting `link.childtype, link.parenttype, or group.membertype` without converting `entry_type_`  back to class as we have discussed [here](https://github.com/asyml/forte/pull/900#discussion_r955458229)

forte/data/ontology/top.py
Pushkar-Bhuse: This was actually a small bug that I identified in the previous PR (#882). This keeps the structure of `MultiPackGroup` consistent with its implementation (since it was previously implemented as a list of tuples).

hunterhector: I think the interface for `Group` should be that the members are exposed as a list of entries? I think as the interface, a tuple of `int` is a bit weird. When you say "keep the structure consistent with implementation", where is the implementation you are referring to?

Pushkar-Bhuse: So for `Group` type entries, we fetch all members of the group using `get_members` and if you see the implementation of [that](https://github.com/Pushkar-Bhuse/forte/blob/23e8ac88092c70fe6696f827a4dec9257bd27683/forte/data/ontology/top.py#L540) you will see that we return a list of entries.

The `members` attribute is a `dataclass` attributes which means that it will be stored in the `DataStore`. Thus it should be in primitive python types. Moreover, this `members` list does the job of [this](https://github.com/asyml/forte/blob/d6b137d315a958475568ebda85fde9c2d98afb54/forte/data/ontology/top.py#L630) members list in the previous implementation which was just used to store the member list in `DataStore` and is thus never exposed to the user directly.

hunterhector: One related interface problem is on ontology generation. I think our ontology code generator, will anything happen there.

The reason that I am worried about the interface is that  MultiPackGroup is a user-facing class anyway, so the public member of it could have some implication/impact on users.

I will let @mylibrar decide how to deal with this here.

forte/data/data_pack.py
mylibrar: https://github.com/asyml/forte/issues/915

forte/data/base_pack.py
mylibrar: https://github.com/asyml/forte/issues/916

mylibrar (approved):

PR #907, opened 2022-11-29
Performance improvement 904

J007X:
This PR fixes [https://github.com/J007X/forte/tree/PerformanceImprovement_904]. 

### Description of changes
Profiling test find bottleneck in the "get"method of Entry in data_pack.py where using len() causing 
slowdown so after discussion this is now fixed using underlying number count.

This change can significantly improve the performance and the impact is minimal .

### Test Conducted
This should pass all existing unit test code and have good performance on the profiling test.


hunterhector (commented):
The PR is simple and I think it is fine.

Just fix the first line of the PR message, use the issue url.

J007X:
> The PR is simple and I think it is fine.
> 
> Just fix the first line of the PR message, use the issue url.

Sorry the issue link previously was using a wrong URL of my branch. Now use the public issue URL.

hunterhector:
> > The PR is simple and I think it is fine.
> > Just fix the first line of the PR message, use the issue url.
> 
> Sorry the issue link previously was using a wrong URL of my branch. Just changed to the public issue URL of 904.

I just updated the regular way.

hunterhector (approved):

Issue #923, opened 2023-03-01
_is_annotation_tid() in data_store exceptions throwing causing (significant) slowing down in typical usage scenarios (such as NLP)


J007X:
**Describe the bug**
In some routine profiling check on performance on typical usage scenarios , (significant) slowdown is detected after some recent code change. After analysis, _is_annotation_tid() in data_store is identified as the method currently consuming too much time and causing slow down. After detailed analysis and further debugging, it looks like with the recent underlying changes making the look up of the dictionaries (populated by methods affected by changes) different than before , and newly added exception handling code also affects performance. 

**To Reproduce**
1. using standard profiling test code such as in #805 (using a standard pipeline for NLTK based POS tagger and NER processing) 
2. Perform profiling test in PyCharm
3. See performance difference, and the Cprofile analysis report and Call Graph it generated

**Expected behavior**
_is_annotation_tid() is identified to be "hot"(consumes significant time) by cprofile. Also further debugging shows excessive exceptions (more than previous version) was thrown from it.

**Environment (please complete the following information):**
 - OS: All
 - Version : current code base (0.3, snapshot of Feb.06 or Feb.14)
 - Python and Package verions: 3.8

**Additional context**
(Currently investigating) recent underlying code change related to tid, entry and related code for populating related dictionary

J007X:
Adjust the title and description slightly to reflect latest  investigation results --  after adding some tracing/debugging code it seems the slow down is caused by exception throwing code newly added (in which the Call Graph identified related method as "green" as they are system method). The dictionary access was different from before however the changes are small and not causing this significant changes in performance.

PR #925, opened 2023-03-13
Fix performance impact of exception handling in _is_annotation_tid()

J007X:
This PR fixes https://github.com/asyml/forte/issues/923. 

### Description of changes
Related test code were added to determine details related exception (including statistics comparison with earlier code). The data shows the newly added exception handling code impact the performance most while the difference in dictionary access is insignificant so only the exception handling code is changed.   

Since no other place than the exception handling part is changed, the impact to other code is minimum, and the performance issue can be resolved (speed restored to previous level).

### Test Conducted
Usual unit tests and profiling tests.

### DIFF
--- a/forte/data/data_store.py
+++ b/forte/data/data_store.py
@@ -806,7 +806,6 @@ class DataStore(BaseStore):
         else:
             attr_fields: Dict = self._get_entry_attributes_by_class(type_name)
             for attr_name, attr_info in attr_fields.items():
-
                 attr_class = get_origin(attr_info.type)
                 # Since we store the class specified by get_origin,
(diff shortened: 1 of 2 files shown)


hunterhector:
Thanks for the fix. It looks like it is not ideal to create and include a large string in the exception. 

I think this PR is fine. Please fix the CI problem and we can merge it.

To avoid unnecessary turnaround due to things like "black", you could add the pre-commit-hook locally as specified here.
https://github.com/asyml/forte/blob/master/CONTRIBUTING.md#coding-style

hunterhector:
ci not passed yet.

hunterhector (approved):

